PR #4, opened 2017-06-12
Moving minification options

weaverryan:
The previous spot is an outdated spot anyways, and it caused an edge-case bug with
resolution of some SASS paths: https://github.com/symfony/symfony-demo/pull/586

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -100,7 +100,7 @@ class ConfigGenerator {
     buildRulesConfig() {
         const cssLoaders = [
             {
-                loader: 'css-loader' + this.getSourceMapOption(),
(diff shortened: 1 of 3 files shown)


lib/config-generator.js
```diff
@@ -100,7 +100,7 @@ class ConfigGenerator {
     buildRulesConfig() {
         const cssLoaders = [
             {
-                loader: 'css-loader' + this.getSourceMapOption(),
+                loader: 'css-loader' + this.getSourceMapOption() + (this.webpackConfig.isProduction() ? '?minimize=1' : ''),
```
stof: shouldn't this use `options` instead of the query string way ?

PR #10, opened 2017-06-13
Add window.jquery to autoProvidejQuery()

alOneh:
When using webpack it's common to also provide window.jQuery because some libraries use it this way.


weaverryan:
👍

weaverryan:
Thank you!

alOneh:
@weaverryan can you tag a new version ?

PR #19, opened 2017-06-15
Allow to add extra plugin

alOneh:
Encore does not support adding an extra plugin to the list of registered plugins.
In Symfony demo, we try to reduce the size of one of our entries and have to manually add the plugin to the config by doing this : 

```js
var config = Encore.getWebpackConfig();
config.plugins.push(new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/));
```

I propose to update the API of Encore through a new method `addPlugin(plugin)`.

```js
Encore
// ...
    .addPlugin(new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/))
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,6 +127,21 @@ module.exports = {
     },
 
     /**
+     * Add a plugin to the sets of plugins already registered by Encore
+     *
+     * For example, if you want to add the "webpack.IgnorePlugin()", then:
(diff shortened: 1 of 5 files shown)


weaverryan:
Docs issue: https://github.com/symfony/symfony-docs/issues/8070

PR #21, opened 2017-06-15
Use NamedModulesPlugin for HMR debugging

weaverryan:
When using HMR, the name of your module shows up in the console. By default, since numbers are used, it looks like:

```
[HMR] Updated modules:
[HMR]  - 21
[HMR]  - 241
```

After:

```
[HMR] Updated modules:
[HMR]  - ./node_modules/css-loader/index.js?sourceMap!./node_modules/vue-loader/lib/style-compiler/index.js?{"vue":true,"id":"data-v-504ecf1c","scoped":true,"hasInlineConfig":false}!./node_modules/vue-loader/lib/selector.js?type=styles&index=0!./app/Resources/assets/vuejs/components/Hello.vue
[HMR]  - ./node_modules/style-loader/addStyles.js
```

... which at least makes a bit more sense :).

This adds a bit more logic:

1) In dev, always add `NamedModulesPlugin`. We actually only need this if you're using HMR... but I don't think there's a disadvantage of including it always, so I've done it to try to keep the dev environment as consistent as possible with itself.

2) In prod, we *only* add `HashedModuleIdsPlugin` when versioning is enabled. We could at it always (for consistency), but it slightly increases the size of the built assets, because the module ids are longer.

3) Enable `WebpackChunkHash` only when versioning is used, because it's the only time we use the `[chunkhash]`

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -245,11 +245,10 @@ class ConfigGenerator {
          */
         let moduleNamePlugin;
         if (this.webpackConfig.isProduction()) {
-            // only enable if versioning is on, as it makes the assets *slightly* larger
-            if (this.webpackConfig.useVersioning) {
-                // shorter, and obfuscated module ids
-                moduleNamePlugin = new webpack.HashedModuleIdsPlugin();
-            }
+            // shorter, and obfuscated module ids (versus NamedModulesPlugin)
+            // makes the final assets *slightly* larger, but prevents contents
+            // from sometimes changing when nothing really changed
+            moduleNamePlugin = new webpack.HashedModuleIdsPlugin();
         } else {
             // human-readable module names, helps debug in HMR
             // enable always when ! production for consistency
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -243,20 +243,15 @@ class ConfigGenerator {
          *  * https://github.com/webpack/webpack.js.org/issues/652#issuecomment-273324529
          *  * https://webpack.js.org/guides/caching/#deterministic-hashes
          */
-        let moduleNamePlugin;
         if (this.webpackConfig.isProduction()) {
             // shorter, and obfuscated module ids (versus NamedModulesPlugin)
             // makes the final assets *slightly* larger, but prevents contents
             // from sometimes changing when nothing really changed
-            moduleNamePlugin = new webpack.HashedModuleIdsPlugin();
+            plugins.push(new webpack.HashedModuleIdsPlugin());
         } else {
             // human-readable module names, helps debug in HMR
-            // enable always when ! production for consistency
-            moduleNamePlugin = new webpack.NamedModulesPlugin();
-        }
-
-        if (moduleNamePlugin) {
-            plugins.push(moduleNamePlugin);
+            // enable always when not in production for consistency
+            plugins.push(new webpack.NamedModulesPlugin());
         }
 
         if (this.webpackConfig.useVersioning) {
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -243,20 +243,26 @@ class ConfigGenerator {
          *  * https://github.com/webpack/webpack.js.org/issues/652#issuecomment-273324529
          *  * https://webpack.js.org/guides/caching/#deterministic-hashes
          */
-        // when not using versioning, none of this is important
-        if (this.webpackConfig.useVersioning) {
-            let moduleNamePlugin;
-            if (this.webpackConfig.isProduction()) {
+        let moduleNamePlugin;
+        if (this.webpackConfig.isProduction()) {
+            // only enable if versioning is on, as it makes the assets *slightly* larger
+            if (this.webpackConfig.useVersioning) {
                 // shorter, and obfuscated module ids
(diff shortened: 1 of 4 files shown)


lib/config-generator.js
```diff
@@ -320,20 +320,26 @@ class ConfigGenerator {
          *  * https://github.com/webpack/webpack.js.org/issues/652#issuecomment-273324529
          *  * https://webpack.js.org/guides/caching/#deterministic-hashes
          */
-        // when not using versioning, none of this is important
-        if (this.webpackConfig.useVersioning) {
-            let moduleNamePlugin;
-            if (this.webpackConfig.isProduction()) {
+        let moduleNamePlugin;
+        if (this.webpackConfig.isProduction()) {
+            // only enable if versioning is on, as it makes the assets *slightly* larger
```
stof: Can we have a way to turn it ON otherwise too ? If you use query-string versioning with https://packagist.org/packages/incenteev/hashed-asset-bundle, having a stable file hash is important too.

stof: maybe a `.useStableModuleNames()` method, which is enabled automatically when using versioning ?

weaverryan: @stof Unless you enable versioning, your output filenames will *always* be static and not changing (e.g. `main.js`). This just controls the internal names used for the modules. Or am I missing a detail? Basically, I don't think the module naming scheme should be important externally.

stof: @weaverryan my issue is stable module names, not stable output files.

hashed-asset-bundle uses a (truncated) hash of the file content as version. So if you use numeric index for modules, the hash of the vendor module may change just because of reindexing module, busting the cache for nothing.

you use the HashedModuleIdsPlugin when versioning is enabled to ensure that the chunkHash stay consistent, but we have the same need when using hashed-asset-bundle

weaverryan:
@stof Ah, I totally understand now. I've updated the PR to *always* use one of the plugins, instead of adding a new option. Regardless of where you do versioning, it seems like an edge case to me to "be ok with" your asset contents often changing... even when you didn't make any code changes to it. So, at the cost of a slight increase in asset size (in one example I saw, the asset size went from 296kb to 300kb), I'd like to be consistent and always give this "expected" behavior. If someone later *really* wants to minimize their asset size at the expense of the content sometimes changing, they can make a request (and we can add a method to allow that to be configured). This feels like a more sensible default to me.

stof:
@weaverryan fine with me

lib/config-generator.js
```diff
@@ -324,20 +324,25 @@ class ConfigGenerator {
...
+            // makes the final assets *slightly* larger, but prevents contents
+            // from sometimes changing when nothing really changed
+            moduleNamePlugin = new webpack.HashedModuleIdsPlugin();
+        } else {
+            // human-readable module names, helps debug in HMR
+            // enable always when ! production for consistency
+            moduleNamePlugin = new webpack.NamedModulesPlugin();
+        }
+
+        if (moduleNamePlugin) {
```
stof: useless check. It will always be set

lib/config-generator.js
```diff
@@ -324,20 +324,25 @@ class ConfigGenerator {
...
+            plugins.push(moduleNamePlugin);
```
stof: you could put this directly above instead of using a `moduleNamePlugin` variable.

lib/config-generator.js
```diff
@@ -324,20 +324,25 @@ class ConfigGenerator {
...
-        // when not using versioning, none of this is important
+        let moduleNamePlugin;
+        if (this.webpackConfig.isProduction()) {
+            // shorter, and obfuscated module ids (versus NamedModulesPlugin)
+            // makes the final assets *slightly* larger, but prevents contents
+            // from sometimes changing when nothing really changed
+            moduleNamePlugin = new webpack.HashedModuleIdsPlugin();
+        } else {
+            // human-readable module names, helps debug in HMR
+            // enable always when ! production for consistency
```
stof: I would say `when not production` instead of putting an operator in the middle of an English sentence

test/config-generator.js
```diff
@@ -466,6 +465,33 @@ describe('The config-generator function', () => {
             const actualConfig = configGenerator(config);
             expect(actualConfig.devServer).to.be.undefined;
         });
+
+        it('devServer no hot mode', () => {
+            const config = createConfig();
+            config.runtimeConfig.useDevServer = true;
+            config.runtimeConfig.devServerUrl = 'http://localhost:8080/';
+            config.runtimeConfig.useHotModuleReplacement = false;
+            config.outputPath = '/tmp/public';
```
stof: shouldn't this use a windows path on windows ?

weaverryan: it doesn't matter here, because it's not what I'm testing for. I just need a path that won't cause the code to throw an exception internally (`Unable to determine contentBase option...`) - and that logic for that exception *is* tested elsewhere with the proper paths. We could probably clean this file up a bit with some mocks.

weaverryan: pff, nevermind - I immediately saw the test failure after posting this. I'm trying a Windows path now to see if it'll keep things happy :)

PR #22, opened 2017-06-15
Refactoring loaders into re-usable modules

weaverryan:
This is intended to be an internal cosmetic change. The purpose (other than organization) is to help enable the Vue.js loader... which needs to be able to re-use most of this configuration.


weaverryan:
This is ready - it shouldn't contain any real changes, but will help enable the vue-loader

PR #27, opened 2017-06-16
Add appveyor configuration file

DaRamirezSoto:
Added a configuration file for AppVeyor tried to mimic how it was defined in travis configuration file.
This is first time for me using AppVeyor so it defiantly need some more experienced eyes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,9 @@
-# Should we cache local modules ?
-# cache:
-#  - node_modules
+cache:
+  - node_modules
 
 environment:
   matrix:
     - nodejs_version: "6"
-    - nodejs_version: "7"
-    - nodejs_version: "8"
 
 platform:
   - x86
@@ -19,7 +16,6 @@ install:
 test_script:
   - node --version
   - npm --version
-  - IF %nodejs_version% EQU 7 npm run travis:lint 
   - npm run test
 
 build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,25 @@
+# Should we cache local modules ?
+# cache:
+#  - node_modules
+
+environment:
+  matrix:
+    - nodejs_version: "6"
+    - nodejs_version: "7"
+    - nodejs_version: "8"
+
+platform:
+  - x86
+  - x64
+
+install:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+  - x64
+
+install:
+  - ps: Install-Product node $env:nodejs_version $env:platform
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - IF %nodejs_version% EQU 7 npm run travis:lint
```
stof: linting does not need to be done on both Travis and Appveyor. It only depends on the source code, which is the same. I would remove it from here

appveyor.yml
```diff
@@ -0,0 +1,25 @@
+# Should we cache local modules ?
+# cache:
+#  - node_modules
+
+environment:
+  matrix:
+    - nodejs_version: "6"
+    - nodejs_version: "7"
+    - nodejs_version: "8"
```
stof: Do we actually need to test all node versions on Appveoyr as well ? I don't think it is worth it. AFAIK, Appveoyr is not as good as Travis at handling multiple jobs (especially regarding cache, even though you don't use it yet).

DaRamirezSoto: @stof I agree, I just followed the specs as given by `package.json` which states >= 6.
I can change to only test on 6 if that is what we desire

appveyor.yml
DaRamirezSoto: @stof I just followed the configuration that was given in travis config file.
If we decide on not including this I can remove it

fabpot:
@weaverryan I've just enabled Appveyor on symfony/webpack-encore

weaverryan:
@DaRamirezSoto Awesome! The build is running great - it looks like it's the one failure you mentioned over and over again. I'll merge so that I can open a PR to get rid of that :). Thanks!

@fabpot I tried to enabled it yesterday through my account, but it gave the project a `weaverryan/webpack-remix` name. And I still don't see https://ci.appveyor.com/project/symfony/webpack-encore after you've enabled it.

fabpot:
@weaverryan hehe, AppVeyor works very differently from Travis. So, the URL is https://ci.appveyor.com/project/fabpot/webpack-encore

PR #28, opened 2017-06-16
Fixing Windows Support

weaverryan:
In a few places, we rely (relied) on doing some "string replace" on paths. Obviously, that doesn't play well with the windows `\` directory separators. This fixes the few places that rely on this and includes tests that exposed those failures on AppVeyor.

### DIFF
--- a/lib/config/path-util.js
+++ b/lib/config/path-util.js
@@ -87,7 +87,7 @@ module.exports = {
 
         let outputPath = webpackConfig.outputPath;
         // for comparison purposes, change \ to / on Windows
-        outputPath = outputPath.split('\\').join('/');
+        outputPath = outputPath.replace(/\\/g, '/');
 
         // remove trailing slash on each
         outputPath = outputPath.replace(/\/$/, '');
--- a/lib/config/path-util.js
+++ b/lib/config/path-util.js
@@ -1,0 +1,110 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+const path = require('path');
+
+module.exports = {
+    /**
+     * Determines the "contentBase" used for the devServer.
+     *
+     * @param {WebpackConfig} webpackConfig
+     * @return {String}
+     */
+    getContentBase(webpackConfig) {
+        // strip trailing slash (for Unix or Windows)
+        const outputPath = webpackConfig.outputPath.replace(/\/$/,'').replace(/\\$/, '');
+        // use the manifestKeyPrefix if available
+        const publicPath = webpackConfig.manifestKeyPrefix ? webpackConfig.manifestKeyPrefix.replace(/\/$/,'') : webpackConfig.publicPath.replace(/\/$/,'');
+
+        /*
+         * We use the intersection of the publicPath (or manifestKeyPrefix) and outputPath
+         * to determine the "document root" of the web server. For example:
+         *   * outputPath = /var/www/public/build
+         *   * publicPath = /build/
+         *      => contentBase should be /var/www/public
+         */
+
+        // start with outputPath, then join publicPath with it, see if it equals outputPath
+        // in loop, do dirname on outputPath and repeat
+        // eventually, you (may) get to the right path
+        let contentBase = outputPath;
+        while (path.dirname(contentBase) !== contentBase) {
+            contentBase = path.dirname(contentBase);
+
+            if (path.join(contentBase, publicPath) === outputPath) {
+                return contentBase;
+            }
+        }
+
+        throw new Error(`Unable to determine contentBase option for webpack's devServer configuration. The publicPath (${webpackConfig.publicPath}) string does not exist in the outputPath (${webpackConfig.outputPath}), and so the "document root" cannot be determined.`);
+    },
+
+    /**
+     * Returns the output path, but as a relative string (e.g. web/build)
+     *
+     * @param {WebpackConfig} webpackConfig
+     * @return {String}
+     */
+    getRelativeOutputPath(webpackConfig) {
+        return webpackConfig.outputPath.replace(webpackConfig.getContext() + '/', '');
+    },
+
+    /**
+     * If the manifestKeyPrefix is not set, this uses the publicPath to generate it.
+     *
+     * Most importantly, this runs some sanity checks to make sure that it's
+     * ok to use the publicPath as the manifestKeyPrefix.
+     *
+     * @param {WebpackConfig} webpackConfig
+     * @return {String}
+     */
+    generateManifestKeyPrefix(webpackConfig) {
+        if (webpackConfig.publicPath.includes('://')) {
+            /*
+             * If publicPath is absolute, you probably don't want your manifests.json
+             * keys to be prefixed with the CDN URL. Instead, we force you to
+             * choose your manifestKeyPrefix.
+             */
+
+            throw new Error('Cannot determine how to prefix the keys in manifest.json. Call Encore.setManifestKeyPrefix() to choose what path (e.g. build/) to use when building your manifest keys. This is happening because you passed an absolute URL to setPublicPath().');
+        }
+
+        let outputPath = webpackConfig.outputPath;
+        // for comparison purposes, change \ to / on Windows
+        outputPath = outputPath.split('\\').join('/');
+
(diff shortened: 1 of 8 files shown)


lib/config/path-util.js
```diff
@@ -0,0 +1,37 @@
...
+         * to determine the "document root" of the web server. For example:
+         *   * outputPath = /var/www/public/build
+         *   * publicPath = /build/
+         *      => contentBase should be /var/www/public
+         */
+
+        // start with outputPath, then join publicPath with it, see if it equals outputPath
+        // in loop, do dirname on outputPath and repeat
+        // eventually, you (may) get to the right path
+        var contentBase = outputPath;
```
stof: should be `let` rather than `var`

weaverryan:
Tests are green 🎆 - this is ready :)

lib/config/path-util.js
```diff
@@ -0,0 +1,110 @@
...
+        // strip trailing slash (for Unix or Windows)
+        const outputPath = webpackConfig.outputPath.replace(/\/$/,'').replace(/\\$/, '');
+        // use the manifestKeyPrefix if available
+        const publicPath = webpackConfig.manifestKeyPrefix ? webpackConfig.manifestKeyPrefix.replace(/\/$/,'') : webpackConfig.publicPath.replace(/\/$/,'');
+
+        /*
+         * We use the intersection of the publicPath (or manifestKeyPrefix) and outputPath
+         * to determine the "document root" of the web server. For example:
+         *   * outputPath = /var/www/public/build
+         *   * publicPath = /build/
```
stof: what if publicPath is `/subdir/build/` ? AFAIK `path.join(contentBase, publicPath) === outputPath` will never be true, as `outputPath` will use `\` everywhere, while `path.join(contentBase, publicPath)` will have some `/`

lib/config/path-util.js
```diff
@@ -0,0 +1,110 @@
...
+             * keys to be prefixed with the CDN URL. Instead, we force you to
+             * choose your manifestKeyPrefix.
+             */
+
+            throw new Error('Cannot determine how to prefix the keys in manifest.json. Call Encore.setManifestKeyPrefix() to choose what path (e.g. build/) to use when building your manifest keys. This is happening because you passed an absolute URL to setPublicPath().');
+        }
+
+        let outputPath = webpackConfig.outputPath;
+        // for comparison purposes, change \ to / on Windows
+        outputPath = outputPath.split('\\').join('/');
```
stof: is splti+join faster than `.replace(/\\/g, '/')` ?

lib/webpack/webpack-manifest-plugin.js
```diff
@@ -45,6 +45,13 @@ ManifestPlugin.prototype.apply = function(compiler) {
           path.dirname(file),
           path.basename(module.userRequest)
       );
+      /* *** MODIFICATION START *** */
+        // for Windows, we want the keys to use /, not \
+        // (and path.join will obviously use \ in Windows)
+        if (process.platform === 'win32') {
+            moduleAssets[file] = moduleAssets[file].split('\\').join('/');
+        }
+      /* *** MODIFICATION END *** */
```
stof: Can this be contributed upstream too ?

test/config/path-util.js
```diff
@@ -0,0 +1,155 @@
...
+    return new WebpackConfig(runtimeConfig);
+}
+
+/**
+ * Some tests are very specific to different operating systems.
+ * We use this to only run them when needed.
+ *
+ * @returns {boolean}
+ */
+function isWindows() {
```
stof: this could be a const rather than a function, as the value being returned will always be the same:

```js
const isWindows = process.platform === 'win32'

// ...

        if (isWindows) {
            this.skip();
        }
```

test/config/path-util.js
```diff
@@ -0,0 +1,155 @@
...
+            config.setPublicPath('/build/');
+            config.addEntry('main', './main');
+
+            const actualContentBase = pathUtil.getContentBase(config);
+            expect(actualContentBase).to.equal('C:\\projects\\webpack-encore\\public');
+        });
+    });
+
+    describe('generateManifestKeyPrefix', () => {
+        it('manifestKeyPrefix is correctly not required on windows', () => {
```
stof: shouldn't this be skipped on Unix ?

test/config/path-util.js
```diff
@@ -0,0 +1,155 @@
...
+            const config = createConfig();
+            config.outputPath = 'C:\\projects\\webpack-encore\\web\\build';
+            config.setPublicPath('/build/');
+            config.addEntry('main', './main');
+
+            const actualPrefix = pathUtil.generateManifestKeyPrefix(config);
+            expect(actualPrefix).to.equal('build/');
+        });
+
+        it('with absolute publicPath, manifestKeyPrefix must be set', () => {
```
stof: shouldn't this be skipped on windows ?

test/config/path-util.js
```diff
@@ -0,0 +1,155 @@
...
+            config.setPublicPath('https://cdn.example.com');
+
+            expect(() => {
+                pathUtil.generateManifestKeyPrefix(config);
+            }).to.throw('Cannot determine how to prefix the keys in manifest.json. Call Encore.setManifestKeyPrefix() to choose what path (e.g. build/) to use');
+        });
+
+        it('when outputPath and publicPath are incompatible, manifestKeyPrefix must be set', () => {
+            const config = createConfig();
+            config.outputPath = '/tmp/public/build';
```
stof: We need to have a windows test for that too (or just change the output path here depending on the OS instead of duplicating everything)

lib/config/path-util.js
weaverryan: replace only replaces the *first* occurrence :/

lib/config/path-util.js
weaverryan: You're correct. But, if the user has a `publicPath` that is not a sub-string of the `outputPath`, they will already have received an error forcing them to set `manifestKeyPrefix` (you do not need to set this manually... until you deploy under a subdirectory or set publicPath to a CDN). So, this situation won't happen. I've just pushed a commit to hopefully clarify this - it's a bit complex since different parts of the system are working together. It *is* possible that the user will set a `manifestKeyPrefix` that is not a substring of the `outputPath`, and then they would see this error (because this is not allowed)

test/config/path-util.js
weaverryan: +1

test/config/path-util.js
weaverryan: Good idea to alternate the path based on OS - doing that!

test/config/path-util.js
weaverryan: Actually, based on your other suggestion - I've removed all of the skips and just add a bit of `isWindows` logic in each function to change the paths based on OS. This means all tests will run on both OS's - best of both worlds (and less duplication as you mentioned)

lib/config/path-util.js
stof: no, when using a regex, it depends whether you pass a regex matching the first place or all of them (think `preg_match` vs `preg_match_all` in PHP). I added the `g` modifier on my regex.

test/config/path-util.js
```diff
@@ -0,0 +1,118 @@
...
+    runtimeConfig.babelRcFileExists = false;
+
+    return new WebpackConfig(runtimeConfig);
+}
+
+/**
+ * Some tests are very specific to different operating systems.
+ * We use this to only run them when needed.
+ *
+ * @returns {boolean}
```
stof: `@returns` is wrong for a constant

lib/config/path-util.js
weaverryan: You're right! I was mixing this up with the non-regex behavior or replace, which only replaces the first. I'll update these to regex (it's less "weird" than this)

stof: I still suggest doing a benchmark (especially before contributing to webpack-manifest-plugin

lib/webpack/webpack-manifest-plugin.js
```diff
@@ -45,6 +45,13 @@ ManifestPlugin.prototype.apply = function(compiler) {
           path.dirname(file),
           path.basename(module.userRequest)
       );
+      /* *** MODIFICATION START *** */
+        // for Windows, we want the keys to use /, not \
+        // (and path.join will obviously use \ in Windows)
+        if (process.platform === 'win32') {
+            moduleAssets[file] = moduleAssets[file].replace(/\\/g).join('/');
```
stof: this is broken. proper code is `.replace(/\\/g, '/')`

PR #34, opened 2017-06-19
Add vue loader

julienj:
This PR is a patch to provide a vue.js support

before : 
`
    .addLoader({
        test: /\.vue$/,
        loader: 'vue-loader'
    })
`

After
`
     .enableVueLoader()
`

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -44,6 +44,8 @@ class WebpackConfig {
             resolve_url_loader: true
         };
         this.useLessLoader = false;
+        this.useVueLoader = false;
+        this.vueOptions = {};
         this.cleanupOutput = false;
         this.sharedCommonsEntryName = null;
         this.providedVariables = {};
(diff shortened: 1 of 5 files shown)


weaverryan:
Thanks @julienj! But this needs a bit more work so that the styles can be extracted. I'm making that easier in #22. Would you mind if I used this PR as a starting point and finished the loader?

lib/WebpackConfig.js
```diff
@@ -222,6 +224,11 @@ class WebpackConfig {
         this.useReact = true;
     }
 
+    enableVueLoader(options) {
+        this.useVueLoader = true;
+        this.vueOptions = options;
+    }
```
weaverryan: Can you change `options` to be a callback, like we do with `configureBabel()`? The reason is that we *will* (in order to extract styles) add a bunch of options to the loader automatically. If the user wants to add even *more* options, we'll want them to do that in callback (we'll pass them our options, then they can change/add them).

julienj:
Thank you.
No problem to continue on your side

PR #41, opened 2017-06-20
Forcing babel-loader 7.1 and updating resolve-rc for new argument

weaverryan:
In babel-loader 7.1, they added a new argument to `resolve-rc`. This updates Encore to require 7.1 and pass in that new argument.

Of course, the underlying problem is that we're relying on an internal function - so it's entirely valid for them to break BC. We should obviously minimize doing this, but I still think it's a good idea: I don't want to try to re-implement their logic. For example, there are now 2 new places where you can store the Babel configuration... and by using `resolve-rc`, we're now checking for those automatically.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -176,7 +176,7 @@ class WebpackConfig {
         }
 
         if (this.doesBabelRcFileExist()) {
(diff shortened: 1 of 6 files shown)


weaverryan:
Tests are passing now - the appveyor failures will be fixed via #28

robfrawley:
This change makes using `babel-loader` v7.0 *incompatible* with Encore. There *must* be a changelog that conveys when breaking changes are introduced with use of dependencies!

PR #45, opened 2017-06-21
Update appveyor to use yarn instead of npm and auto fix windows line …

DaRamirezSoto:
- Added fix to force windows line endings
- Added yarn to cache
- Replaced npm with yarn for install
- Removed npm version check

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,12 @@
 cache:
   - node_modules
+  - "%LOCALAPPDATA%\\Yarn"
+
+init:
+  - git config --global core.autocrlf true
 
 environment:
-  matrix:
-    - nodejs_version: "6"
+  nodejs_version: "6"
 
 platform:
   - x86
@@ -11,11 +14,11 @@ platform:
 
 install:
   - ps: Install-Product node $env:nodejs_version $env:platform
-  - npm install
+  - yarn
 
 test_script:
   - node --version
-  - npm --version
-  - npm run test
+  - yarn --version
+  - yarn run test
 
 build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,6 @@ cache:
   - node_modules
   - "%LOCALAPPDATA%\\Yarn"
 
-init:
-  - git config --global core.autocrlf true
-
 environment:
   nodejs_version: "6"
 


appveyor.yml
```diff
@@ -1,21 +1,24 @@
 cache:
   - node_modules
+  - "%LOCALAPPDATA%\\Yarn"
+
+init:
+  - git config --global core.autocrlf true
```
stof: -1 for that. We need to make sure that encore works *without* altering the source code (because normal usage will not do such conversion)

DaRamirezSoto: @stof Could you explain what you mean?
I got this from the yarn website: https://yarnpkg.com/lang/en/docs/install-ci/#appveyor-tab
Looking at other projects they use the same line

stof: The doc you linked does not have this line

DaRamirezSoto: @stof You could have been a bit more clear as to what you were talking about.
Your feedback leaves allot of room for misunderstanding.

This is not best practices considering windows.There are cases where you will run into trouble with this. Since I don't think we will have those here I'll remove `autocrlf` line endings on your request.

stof: Well, my comment is on the exact line to which it applies.

And converting the code to CRLF on checkout always *created* bug in my experience (I'm not using Windows anymore for development, but I did it during years), because it means you are not running the source code of the project but a different one. The meaningful value IMO is `input`, not `true` (but this is irrelevant on CI as it will never commit anything). The only valid use case for this setting AFAIK is being able to open the code in Notepad (which is the only text editor I know of which does not support Unix line endings, but I would never use it to write code anymore).
Thus, it is even worse for CI: people installing Encore on Windows will *not* get the line endings converted (as npm/yarn will not use git for that, but download a tarball, and so won't be affected by your git config), meaning that converting line endings on CI would mean that the CI does not test the code that users will use.

weaverryan:
Thanks for all your help with this @DaRamirezSoto! The Windows support should be fixed/merged soon - I'm very happy with how it's all setup - I want Windows to work *great* :).

DaRamirezSoto:
@weaverryan No problem, I like to help more but for now I cannot because of current projects.
I will when I am sure I can give proper support to my contributions; until then I go for low hanging fruit.

PR #46, opened 2017-06-21
Removing window.jQuery from autoProvidejQuery

### LINKED ISSUE #32: jQuery not exposed
I have problem with providing `jQuery` as var to global scope for legacy js code.

main.js
```
var $ = require('jquery');
window.$ = window.jQuery = $;
```

webpack-config.js
```

Encore
    // directory where all compiled assets will be stored
    .setOutputPath('mainsites/build/')

    // what's the public path to this directory (relative to your project's document root dir)
    .setPublicPath('/build')

    // empty the outputPath dir before each build
    .cleanupOutputBeforeBuild()

    // will output as web/build/app.js
    .addEntry('app', './assets/js/main.js')

    // allow legacy applications to use $/jQuery as a global variable
    .autoProvidejQuery()

    // create hashed filenames (e.g. app.abc123.css)
    .enableVersioning()
;

// export the final configuration
```

package.js
```
{
  "name": "schoolm",
  "version": "0.0.1",
  "license": "UNLICENSED",
  "private": true,
  "dependencies": {
    "jquery": "^2.1.4",
  },
  "devDependencies": {
    "@symfony/webpack-encore": "github:symfony/webpack-encore"
  }
}
```

html
```
<?php
$webpackConfig = json_decode(file_get_contents(__DIR__ . '/../mainsites/build/manifest.json'), true);
?>
<script src="<?php echo $webpackConfig['build/app.js']; ?>"></script>
```

chrome console
```
> jQuery
VM117:1 Uncaught ReferenceError: jQuery is not defined
    at <anonymous>:1:1
(anonymous) @ VM117:1
> $
function ( selector, context ) {
		// The jQuery object is actually just the init constructor 'enhanced'
		// Need init if jQuery is called (just allow error to be thrown if not included)
		return new jQuery.…
```

weaverryan:
Fixes #32 

This is because - when this is enabled - it's impossible to expose jQuery as a global
variable on window (without use the expose loader). Specifically, window.jQuery = ...
is re-written in your JS code.

A docs PR also needs to be opened to document how to fix problems if a library depends on `window.jQuery` specifically (is there a good re-world example?)

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -244,7 +244,6 @@ class WebpackConfig {
         this.autoProvideVariables({
             $: 'jquery',
             jQuery: 'jquery',
-            'window.jQuery': 'jquery'
         });
     }
 


weaverryan:
I'm going to merge this so that the next release contains the same code (i.e. without `window.jQuery`) as the current release. Then we can see what we want to do going forward.

PR #49, opened 2017-06-21
Vue.js Loader

weaverryan:


weaverryan:
This is ready to go 🍰  Notes:

* HMR works for JavaScript (wooohoo!) but not for styles, because we're using ExtractTextPlugin like the rest of the library (see #3)

* I added the `.vue` extension always, not just when the vue-loader is enabled. This should give better errors - if I use vue and require a `.vue` file without the extension, instead of not finding the file, you'll get an error about the `vue-loader` needing to be enabled.

lib/loaders/sass.js
```diff
@@ -35,10 +37,10 @@ module.exports = {
 
         sassLoaders.push({
             loader: 'sass-loader',
-            options: {
+            options: Object.assign(sassOptions, {
```
stof: we should not alter the sassOptions object passed as argument. It could have side effects. Use this instead: `Object.assign({}, sassOptions, { sourceMap: ... })` instead

lib/loaders/vue.js
```diff
@@ -0,0 +1,105 @@
...
+             * The optional loaders are always added here. Without these,
+             * if the user forgets to enable an optional loader but uses
+             * (for example) lang="scss" inside vue, it *will* work, becaus
+             * vue-loader will automatically register the sass-loader. But,
+             * it will completely skip our configuration (e.g. ExtractTextPlugin).
+             */
+
+            scss: extractText.extract(
+                webpackConfig,
+                sassLoaderUtil.getLoaders(webpackConfig, {}, true),
```
stof: wouldn't this require having sass-loader installed as soon as you use vue-loader, even if you never use Sass in your project, but less (same in the opposite way for the `less` config of course) ?

test/functional.js
```diff
@@ -603,5 +603,84 @@ module.exports = {
...
+            });
+
+            const appDir = testSetup.createTestAppDir();
+            /*
+             * Most of the time, we don't explicitly use chdir
+             * in order to set the cwd() to the test directory.
+             * That's because, in theory, you should be able to
+             * run Encore from other directories, give you set
+             * the context. However, in this case, the vue-loader
+             * users load-postcss-config to load the postcss config,
```
stof: typo: `uses`

test/functional.js
```diff
@@ -603,5 +603,84 @@ module.exports = {
...
+             * but it uses process.cwd() to find it, instead of the
```
stof: is the context an Encore thing or a webpack one ? If it is a webpack one, it would be worth discussing this upstream

lib/loaders/sass.js
weaverryan: 👍

test/functional.js
weaverryan: context is a webpack thing... and it's meant to be the root from which entry files are loaded. It's unclear to me whether context should also be used for other things, like the directory where postcss.config.js should be loaded (we *do* give it a bit more importance in Encore)

weaverryan:
I added one more commit to change the vue-loader options to a callback - consistent with #50 and `configureBabelLoader()`

PR #50, opened 2017-06-21
Add support to enableTypescriptLoader

davidmpaz:
Hi,

please would you consider adding this to main stream?

this add public API to enable TypeScript loader ([ts-loader](https://github.com/TypeStrong/ts-loader)). It follows much of what is described in #9 

Cuurrently only add vanilla loader as described in [here](https://github.com/TypeStrong/ts-loader/blob/master/examples/vanilla/webpack.config.js) playing side by side with babel.

Future work would be to include also support to `happypack` and `fork-ts-checker-webpack-plugin`

Thanks in advance


weaverryan:
Hey @davidmpaz! I would be very happy to accept this :). Can you rebase - I recently merged some PR's?

> Future work would be to include also support to happypack and fork-ts-checker-webpack-plugin

Indeed - happypack is totally unrelated to TypeScript, correct? And `fork-ts-checker-webpack-plugin` simply adds extra type-checking (and so therefore, build errors/warnings), correct? I just want to minimize BC breaks - so I want to make sure "future things" won't be actually difficult to add later :).

index.js
```diff
@@ -342,6 +342,21 @@ module.exports = {
...
+     * Call this if you plan on loading TypeScript files.
+     *
+     * Supported options:
+     * @see https://github.com/TypeStrong/ts-loader/blob/master/README.md#available-options
+     *
+     * @param {object} options
+     * @return {exports}
+     */
+    enableTypeScriptLoader(options = {}) {
+        webpackConfig.enableTypeScriptLoader(options);
```
weaverryan: I think `options` should be a callback - like we do with `configureBabel()`, so that we don't need to worry about merging about default options and their options.

lib/WebpackConfig.js
```diff
@@ -222,6 +237,18 @@ class WebpackConfig {
         this.useReact = true;
     }
 
+    enableTypeScriptLoader(options = {}) {
+        this.useTypeScriptLoader = true;
+
+        for (const optionKey of Object.keys(options)) {
+            if (!(optionKey in this.typeScriptOptions)) {
+                throw new Error(`Invalid option "${optionKey}" passed to enableTypeScriptLoader(). Valid keys are ${Object.keys(this.typeScriptOptions).join(', ')}`);
+            }
```
weaverryan: I usually like adding nice validation like this, but I'm afraid maintaining this will be a pain (we need to add a new option whenever they add a new option). The typescript-loader should ideally validate these keys themselves.

package.json
```diff
@@ -38,12 +38,12 @@
...
     "lodash": ">=3.5 <5",
     "pkg-up": "^1.0.0",
-    "pretty-error": "^2.1.0",
+    "pretty-error": "^2.1.1",
     "resolve-url-loader": "^2.0.2",
     "style-loader": "^0.13.2",
     "webpack": "^2.2.0",
     "webpack-chunk-hash": "^0.4.0",
-    "webpack-dev-server": "^2.4.5",
+    "webpack-dev-server": "^2.5.0",
```
weaverryan: Let's not bump any versions in this PR - it seems unrelated

test/functional.js
```diff
@@ -563,6 +563,23 @@ module.exports = {
...
+            config.setPublicPath('/build');
+            config.addEntry('main', ['./js/render.ts', './js/index.ts']);
+            config.enableTypeScriptLoader();
+
+            testSetup.runWebpack(config, (webpackAssert) => {
+                // check that ts-loader transformed the ts file
+                webpackAssert.assertOutputFileContains(
+                    'main.js',
+                    'document.getElementById(\'wrapper\').innerHTML = "<h1> Hello World!</h1>";'
+                );
```
weaverryan: great way to test it's working - I like that!

fixtures/js/render.ts
```diff
@@ -0,0 +1,5 @@
+function render() {
```
weaverryan: What about renaming this to `render.tsx` - it would slightly enhance the test, since it would take advantage of the resolve.extensions change and make sure the loader applies to both extensions

lib/WebpackConfig.js
```diff
@@ -51,6 +51,21 @@ class WebpackConfig {
...
+            logLevel: 'info',
+            silent: false,
+            ignoreDiagnostics: [],
+            compiler: 'typescript',
+            configFileName: 'tsconfig.json',
+            visualStudioErrorFormat: false,
+            compilerOptions: {},
+            entryFileIsJs: false,
+            appendTsSuffixTo: []
+        };
```
weaverryan: When we remove the validation below, we should be able to remove these, correct (since these are just the default values)?

test/functional.js
weaverryan: Btw, the test page is currently blank, but for the `vue` loader, I'm adding a `<div id="app"></div>` to it in `setup.js` where `testing.html` is created for the functional tests. You could do the same

index.js
davidmpaz: Good point and will update that.

lib/WebpackConfig.js
davidmpaz: These are the same default values taken from `ts-loader` I add them just to follow the same scheme of implementation Encore had, but if that validation will be removed we can remove also those defaults. 

Also if configuration by callback is added then makes no sense since we are delegating all of that to users. Make sense to make other validations for sanity check but not these currently.

lib/WebpackConfig.js
davidmpaz: More about what we talked already. Is a good point and I will update the PR to take this insto account.

package.json
davidmpaz: That was me being neglectful, sorry about that ^ ^ it was not intended.

test/functional.js
davidmpaz: The original intend was to check that we got something expected in the compiled module. It was just about compilation going through and check that the file was imported.

Did you mean that inside that functional test one can make reference to `document.getElementById('app')` ? It can be done of course. But how is this checked later on? and how will this be usefull for the tests ?

fixtures/js/render.ts
davidmpaz: Very good idea! :+1:

davidmpaz:
Hey @weaverryan  I have updated PR. To answer some of your questions:

> `happypack` is totally unrelated to TypeScript, correct?

Yes. These are the loaders currently now to be supported: https://github.com/amireh/happypack/wiki/Loader-Compatibility-List. This plugin IMHO is critical when having slow build, for now it can be added seperated to Encore but will be nice to have this one also included. I havent check yet other solutions to reduce build time.

> And `fork-ts-checker-webpack-plugin` simply adds extra type-checking (and so therefore, build errors/warnings), correct?

This one makes the types checking in separate thread, meaning that TypeScript compiler doesn't need to do it, making compilation time faster by setting [`transpileOnly = true`](https://github.com/TypeStrong/ts-loader#transpileonly-boolean-defaultfalse) to loader. Actually I did not add this one as "default" setup in loaders because `eslint` was complaining about the module not being published, not sure what that was.

But actually this could be included as bonus on Encore alongside with `ts-loader`. Encore could handle transparently the configuration in case users decide to enable the *type checking in separate* thread to speed up compilation/build time. In my case I reduced several (~2-3k) ms.

test/functional.js
weaverryan: Yes! Zombie executes the JavaScript, so we can assert that the final elements exist on the page correctly (i.e. after the TypeScript code has executed). I've just done this for the vuejs loader - here's a quick gist: https://gist.github.com/weaverryan/64cae3f603529641dcad781eeb0cdcd9

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 npm-debug.log*
 /test_tmp
+.idea
```
weaverryan: You can remove this - it should be set in people's global .gitignore

index.js
```diff
@@ -342,6 +342,25 @@ module.exports = {
...
+     * });
+     *
+     * Supported configuration options:
+     * @see https://github.com/TypeStrong/ts-loader/blob/master/README.md#available-options
+     *
+     * @param {function} callback
+     * @return {exports}
+     */
+    configureTypeScript(callback) {
+        webpackConfig.configureTypeScript(callback);
```
weaverryan: Sorry to go back and forth, but I was thinking more about this - the callback is more difficult to understand and quite a bit harder to use (especially if you want to add a bunch of options at once - you can't just copy some `{foo: bar, baz: true}` code from somewhere and paste it in). I now think we should have this as a normal options object (like you had before). If we eventually add more default options and merging gets weird, we could deprecate the options object and force a callback.

lib/config-generator.js
```diff
@@ -67,8 +68,13 @@ class ConfigGenerator {
 
         config.stats = this.buildStatsConfig();
 
+        let extensions = ['.js', '.jsx'];
+        if (this.webpackConfig.useTypeScriptLoader) {
+            extensions = extensions.concat(['.json', '.ts', '.tsx']);
+        }
```
weaverryan: See my 2nd bullet point here: https://github.com/symfony/webpack-encore/pull/49#issuecomment-310548331. Perhaps we should add `.ts` and `.tsx` to extensions *always*.

lib/loaders/typescript.js
```diff
@@ -0,0 +1,50 @@
...
+    getLoaders(webpackConfig) {
+        loaderFeatures.ensureLoaderPackagesExist('typescript');
+
+        // some defaults
+        let config = {
+            transpileOnly: false,
+            happyPackMode: false,
+            logInfoToStdOut: false,
+            logLevel: 'info',
+            silent: true,
```
weaverryan: It looks like these are already all of the default values (except for `silent`), so can we just remove them instead? And why `silent: true`?

lib/loaders/typescript.js
```diff
@@ -0,0 +1,50 @@
...
+        // allow for ts-loader config to be controlled
+        webpackConfig.tsConfigurationCallback.apply(
+            // use config as the this variable
+            config,
+            [config]
+        );
+
+        // use ts alongside with babel
+        // @see https://github.com/TypeStrong/ts-loader/blob/master/README.md#babel
+        let loaders = babelLoader.getLoaders(webpackConfig);
```
weaverryan: love the note + link 👍

package-lock.json
```diff
@@ -0,0 +1,5187 @@
+{
```
weaverryan: remove this `package-lock.json` file - we're using yarn internally in the library

test/functional.js
```diff
@@ -583,6 +583,50 @@ module.exports = {
...
+            });
+        });
+
+        it('When enabled, react JSX and TypeScript work nice together!', (done) => {
+            const config = createWebpackConfig('www/build', 'dev');
+            config.setPublicPath('/build');
+            config.addEntry('main', ['./js/CoolReactComponent.jsx', './js/render2.tsx']);
+            config.enableReactPreset();
+            config.configureTypeScript(function(tsConfig) {
+                tsConfig.compilerOptions = { 'jsx': 'preserve' };
```
weaverryan: Can you tell me about this option? And why would react and typescript cause problems when enabled together? Is there some known issue that we want to make sure isn't a problem?

.gitignore
davidmpaz: Good point.

index.js
davidmpaz: Don't worry about that, the idea is to get to the best solution possible.

Regarding why I did accept your callback suggestion before. I indeed thought the same as you, on seconds thoughts, after looking the code again, Is it better to keep it like options!?

My conclusion was: no. Why? Because I think it is more explicit to have a callback which accept parameters to configure instead having to know the structure of an object configuration that configure several aspects of application. for example.

This is how i thought on how could be integrated other stuff  better in future:
* `fork-ts-checker-webpack-plugin`:  When tried to add this in `ts-loader` configuration did not feel natural since it is another plugin that can be configured by its own. Still is only useful if you use `ts-loader` and with the option: `transpileOnly: true`. In future this could be passed as another config object to callback.
* I also thought on the `happypack` plugin. This one is more general and can be used with many loaders, but still, having the call back would help at the time of configuration because you will have to configure a bit different the loaders when using it.

Because of this I did not question the callback idea. And that's why I think it would be better like this. Maybe thinking about adding parameters to a configuration callback could not be an optimal solution either, in which case I am open to suggestions :)

What do you think about it?

lib/config-generator.js
davidmpaz: I see, that's very good to have. Could you please point me where this validation occur? I just can not find it now :(

lib/loaders/typescript.js
davidmpaz: You are right! I thought on being more verbose there to have it documented in code but it is right.

> And why silent: true?

Avoiding to be verbose with CLI output. As [show here](https://github.com/TypeStrong/ts-loader/blob/master/README.md#silent-boolean-defaultfalse) it doesn't affect processing of files, only output. For example without `silent` it is always printed messages about which `tsconfig.json` is used. I think it was too much and was making also the output of test not nice looking since it was printing out of context. 

Basically it is more a cosmetic change.

package-lock.json
davidmpaz: Better! I don't know if it is me, but I have faced problems with `npm`. The package management is not consistent. Some times I don't get all my dependencies installed.

I am wondering now whether this is the reason why you use yarn? :)

test/functional.js
davidmpaz: It is a TypeScript compilation issue. Remove it and you get compilation errors about not having configure this option. This only happen when compiling `.tsx` files. I did not add a check/validation for that when configuring loader because:
* Not sure where to add it.
* Difficult to know before hand which kind of files will be parsed ? Not sure here though.
* The compilation error is pretty clear. Developers should configure the option, that's it.

Take a look to [this example configuration](https://github.com/TypeStrong/ts-loader/blob/master/examples/react-babel-karma-gulp-happypack/tsconfig.json) I just tooked from there. I am not an expert so most of test/functional/usage I have took them from  `ts-loader` project itself

davidmpaz:
Hey @weaverryan I have added some comments, let me know your thoughts. I need to leave the updates for the weekend probably, if not, on Monday I make a new round.

thanks for the feedback and looking forward.

cheers

package-lock.json
stof: Well, yarn was a better solution than npm 3. And npm 5 was not yet released when we started this project

test/functional.js
davidmpaz: Thanks! :+1:  I could reproduce it for TS. Just one thing. I needed to add the div#app to html generated for functional test. Is this correct? I was not sure if the html was added by some of the scripts included.

lib/friendly-errors/transformers/missing-loader.js
```diff
@@ -49,6 +49,10 @@ function transform(error) {
         case 'jsx':
             error.loaderName = 'react';
             break;
+        case 'tsx':
+        case 'ts':
+            error.loaderName = 'typescript';
+            break;
```
weaverryan: Good catch on this! As it looks like you discovered, this is what gives us the good error when you try to load a `.tsx` file and you haven't enabled its loader :)

lib/test/setup.js
```diff
@@ -140,6 +140,7 @@ function requestTestPage(webRootDir, scriptSrcs, callback) {
 <head>
 </head>
 <body>
+    <div id="app"></div>
```
weaverryan: +1 this is the correct way to do that (I did it in #49 too - we'll just merge them together)

index.js
weaverryan: Oh, and one more thing: let's rename this to `enableTypeScriptLoader()` to be consistent with the other loaders (`configureBabel()` is the outlier - because Babel is always enabled, so we're *just* configuring it).

test/WebpackConfig.js
```diff
@@ -278,6 +278,23 @@ describe('WebpackConfig object', () => {
         });
     });
 
+    describe('configureTypeScript', () => {
+        it('Calling method sets it', () => {
+            const config = createConfig();
+            const testCallback = () => {};
+            config.configureTypeScript(testCallback);
+            expect(config.tsConfigurationCallback).to.equal(testCallback);
```
weaverryan: We should have 1 test that passes no callback, to make sure that works

test/functional.js
weaverryan: Interesting... I just tried your code locally - and the test passed without the `.jsx: 'preserve'` option. If you comment out setting that option, does this test fail for you? Perhaps we would actually need to add some JSX into the .ts files (e.g. https://github.com/TypeStrong/ts-loader/blob/master/examples/react-babel-karma-gulp-happypack/src/main.tsx). And if the error is really clear and we're not doing anything special to handle it, does the test add anything? Or, if the babel preset has already been activated, should we set this option for them (or perhaps too magic)? If we're not sure, I'd prefer to do nothing and let the user configure this themselves.

test/functional.js
```diff
@@ -583,6 +583,74 @@ module.exports = {
...
+                    path.join(config.getContext(), 'www'),
+                    [
+                        'build/main.js'
+                    ],
+                    (browser) => {
+
+                        // assert that the ts module rendered
+                        browser.assert.text('#app h1', 'Welcome to Your TypeScript App');
+                        // make sure the styles are not inline
+                        browser.assert.elements('style', 0);
```
weaverryan: I think we can remove this for this test - I had it for vue, because vue actually allows you to put styles inline in your `.vue` files, so it was relevant to check that those had been processed correctly.

weaverryan:
One more thing - can you run `yarn install` - it should create an updated `yarn.lock` file that needs to be committed.

lib/config-generator.js
```diff
@@ -67,8 +68,13 @@ class ConfigGenerator {
 
         config.stats = this.buildStatsConfig();
 
+        let extensions = ['.js', '.jsx', '.ts', '.tsx'];
+        if (this.webpackConfig.useTypeScriptLoader) {
+            extensions.push('.json');
+        }
```
davidmpaz: Odd!... I recall having issues with loader not resolving the `tsconfig.json` file for TypeScript compilation if not added the `json` extension. But now I see on my project everything is fine without it and also the tests are fine. So I'll remove this.

test/WebpackConfig.js
davidmpaz: Added!

test/functional.js
davidmpaz: > Interesting... I just tried your code locally - and the test passed without the .jsx: 'preserve' option.

For me I was getting an error, clear how to solve, but error at last. Currently is not like that. Also odd! Maybe I was just tired :). I will remove the option since it is not throwing any error and on build time users will be able to notice what to do clearly.

> I think we can remove this for this test

done

test/functional.js
```diff
@@ -583,6 +583,70 @@ module.exports = {
...
+
+                        // assert that the ts module rendered
+                        browser.assert.text('#app h1', 'Welcome to Your TypeScript App');
+                        done();
+                    }
+                );
+            });
+        });
+
+        it('When enabled, react JSX and TypeScript work nice together!', (done) => {
```
weaverryan: I think we should remove this test entirely. Afaik, I think you will get an error if you add some JSX into a .tsx file (that clear error you talked about). But since we're not doing anything in Encore to work around this or help in any way, this test isn't adding any value.

davidmpaz: Sounds good, could you please take care of removing when merging ? From my side will take bit longer now.

thanks in advance :+1:

weaverryan:
Thanks for your awesome and hard work on this David! I also created a docs issue to document this - would you be interested in going at that too? :)

davidmpaz:
It was my pleasure :) ... I am interested to having it available built in as soon as possible.

Regarding docs, sure. Please reference the issue and if possible/exist point me to some guidelines on how do you tackle docs in the project.

weaverryan:
Yea, sorry about that comma - bad merge on my part indeed! The docs issue is here: https://github.com/symfony/symfony-docs/issues/8100. And here's the PR I made (as an example) for vue: https://github.com/symfony/symfony-docs/pull/8101

PR #53, opened 2017-06-23
[WIP] upgrading to webpack 3

### LINKED ISSUE #38: Update dependencies for webpack 3 (when possible)
- Webpack 3 
- For usage with next LTS version of node (8); which now include a lock file similar to yarn
- sync additional dependencies ( probably need to wait until some been updated to work with Webpack 3 )

weaverryan:
Fix #38 

Encore should already support webpack 3, but I don't want to merge this until all of the loaders support it properly (because I don't want new Encore users to get webpack 3 and have any unnecessary issues).


weaverryan:
Deps upgraded to webpack 3.1, including extract-text-plugin 3.0. All dependencies are now advertising support for webpack 3.1 :)

Issue #52, opened 2017-06-22
"$ is undefined" with .autoProvidejQuery() and jQuery plugin


heitorvrb:
I'm trying to use the "foundation-datepicker" package and getting "TypeError: $ is undefined".

`app/Resources/js/app.js`:
```
require('foundation-sites');
require('foundation-datepicker/js/foundation-datepicker.js');
[...]
```
`.autoProvidejQuery()` is enabled in `webpack.config.js`. `.autoProvidejQuery()` works with other jQuery-dependant packages, but when I add this one, it gives the error.

Could it be something in the package? According to the browser profiler, the guilty lines in `foundation-datepicker.js` are:
```
! function($) {

[...]

$.fn.fdatepicker = function(option) {

[...]
```

What should I do? Sorry if this is not very webpack-encore specific.

kcivey:
It's not working for me either. I had to make a separate file containing lines like
```
window.$ = window.jQuery = require('jquery');
window.moment = require('moment');
window._ = require('underscore');
```
and include it in the entry.

weaverryan:
Hi guys!

I know the problem for `foundation-datepicker` (at least). Add this:

```js
Encore
    // ...
    autoProvideVariables({
        'window.jQuery': 'jquery'
    })
```

There was a PR to add this to `autoProvidejQuery` (#10), but it has some unfortunate side effects (#32). This should fix your issue. Internally, we'll debate the best "default" option.

@kcivey You may have been having the same issue... but I would need a bit more code and the error details to know for sure :).

heitorvrb:
It works. Thanks for your help

peyman-mohamadpour:
same problem here with Bootstrap 3.3.7 and on page, arbitrary jQuery stuff.

matudelatower:
in your entry point define:

```js
// require jQuery normally
const $ = require('jquery');

// create global $ and jQuery variables
global.$ = global.jQuery = $;
```

PR #54, opened 2017-06-25
Re-adding window.jQuery to autoProvidejQuery

weaverryan:
By adding `window.jQuery` to `autoProvidejQuery()`, it caused problems by making `window.jQuery = require('jquery')` not work (#32). But, *not* having it causes other obvious problems (see #52).

This re-adds `window.jQuery` back. As a fix, we would document exposing jQuery as a global variable by doing this:

```js
global.$ = global.jQuery = require('jquery');
```

ping @alOneh

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -244,6 +244,7 @@ class WebpackConfig {
         this.autoProvideVariables({
             $: 'jquery',
             jQuery: 'jquery',
+            'window.jQuery': 'jquery',
         });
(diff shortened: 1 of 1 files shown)


jdreesen:
Don't you mean jQuery should be exposed by doing this:

```js
global.$ = global.jQuery = require('jquery');
```

weaverryan:
Thanks @jdreesen - I copied the wrong version :). Updated my comment

PR #66, opened 2017-06-30
Allow usage of devServer combined with an absolute URL as publicPath

### LINKED ISSUE #59: dev-server inside docker?
Hi,

I'm trying to get `encore dev-server` to work inside a docker container. 

I noticed at first that it listens only on the container's loopback interface ('localhost'), which was OK, because I could use `--host 0.0.0.0` to get it to listen on the proper network interface...

But when I do this the manifest.json ends up using `http://0.0.0.0:8080/....` as it's url prefix, and this stops the assets from loading. 

I tried the `--public ...` flag to set a hostname for the public URL, but that doesn't seem to affect manifest generation.

How can I do this?

Thanks.

robertfausk:
This fixes #59 when webpack encore is used inside a docker container and webpack-dev-server needs an absolute URL.

I'm not sure about printing and message of warning.
I used console.log for printing analogous to "bin/encore.js" but maybe FriendlyErrorsWebpackPlugin should be used... There aren't any other warnings yet whjch I could take for reference.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -9,6 +9,7 @@
 
 'use strict';
 
+const chalk = require('chalk');
 const path = require('path');
 const fs = require('fs');
 
@@ -93,7 +94,8 @@ class WebpackConfig {
          */
         if (publicPath.includes('://')) {
             if (this.useDevServer()) {
-                throw new Error('You cannot pass an absolute URL to setPublicPath() and use the dev-server at the same time. Try using Encore.isProduction() to only configure your absolute publicPath for production.');
+                console.log(chalk.bgYellow.black(' WARNING ') + chalk.yellow(' You should not pass an absolute URL to setPublicPath() and use the dev-server at the same time'));
+                console.log();
             }
         } else {
             if (publicPath.indexOf('/') !== 0) {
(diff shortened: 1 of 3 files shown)


weaverryan:
Yes! Thanks for the PR! 

+1 for how you've done the warning. It's quite likely we'll need to refactor into more of a "logging" system... but that can be done easily later.

I think we should add a new flag to activate this behavior - e.g. `encore dev-server --without-dev-server-public-path`... or... something shorter? :). The current behavior is to override the `publicPath` with the URL to the dev server (and to throw an exception for clarity if the user is setting `publicPath` to some other host name). What I really want, is to tell Encore to stop doing that.

With this new flag, if I have `Encore.setPublicPath('/build')`, I don't need to change that to `Encore.setPublicPath('http://localhost/build')`... I can just leave it alone... and Encore will now continue to just use `/build` as the publicPath for all assets. With this flag passed, we *would* still *allow* someone to set an absolute URL to their `publicPath`, but it wouldn't be necessary.

I think it'll also simplify things - we won't need a warning anymore: we'll throw an exception like now... unless they have the new `--without-dev-server-public-path` flag (then we will totally be ok with allowing them to set the publicPath to an absolute URL).

Wdyt?

weaverryan:
I'm working on this locally - I've incorporated your commits. So, I'll close this and make a new PR soon :)

PR #75, opened 2017-07-02
Making enableTypeScriptLoader() options callback optional

weaverryan:
The `enableTypeScriptLoader()` should allow for no callback function - it's probably not needed in most cases.

Ping @davidmpaz


davidmpaz:
I did noticed also :+1:

PR #83, opened 2017-07-06
Allow resetting the state of the Encore singleton

### LINKED ISSUE #48: Support for multiple webpack configs
Webpacks supports 3 different kinds of exports for its config:

- a config object, which is what Encore generates
- an array of config objects, which will then be similar to running webpack multiple times (once per config) but in a single run. This is handy when you want to isolate different shared entries entirely for instance.
- a promise resolving to one of the previous format (handy to allow building the config asynchronously if you need to load it from elsewhere)

Encore does not need to do anything special to support promises. People needing this could simply use Encore *inside* their promise.

However, Encore currently cannot be used for the array case (at least not without accessing internal APIs). This is because the public API of Encore is currently built around a stateful singleton.
Note that all internal APIs are ready for that though, as they rely on a non-singleton WebpackConfig object.

stof:
Closes #48

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ if (!runtimeConfig) {
     throw new Error('Are you trying to require index.js directly?');
 }
 
-const webpackConfig = new WebpackConfig(runtimeConfig);
+let webpackConfig = new WebpackConfig(runtimeConfig);
 
 module.exports = {
     /**
@@ -396,7 +396,7 @@ module.exports = {
     /**
      * Is this currently a "production" build?
      *
-     * @returns {*}
+     * @returns {boolean}
      */
     isProduction() {
         return webpackConfig.isProduction();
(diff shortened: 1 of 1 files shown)


stof:
AFAIK, we have no tests for the Encore layer itself currently (which might be related to the fact that it was not resettable btw), which is why there is no test changes currently.

weaverryan:
> AFAIK, we have no tests for the Encore layer itself currently

Correct: specifically there is no test for `index.js`. I think it would be a good idea to create a smoke test - e.g. configure Encore with a few basic things, call `getWebpackConfig()`, reset it, configure a few other things and call `getWebpackConfig()` again (asserting very basic things on the config in both cases). If you can add that here, awesome. Otherwise, we can add it later - this is a pretty simple patch and doesn't affect any existing code.

stof:
`getWebpackConfig` is hard to test cleanly if you want to cover it all, as it calls `process.exit`. This would require actually using a subprocess.

weaverryan:
It only calls `process.exit()` if there is a validation error. So yea, testing that case is not worth it. But testing a successful case as a "smoke test" should not be too hard.

weaverryan:
Thanks Stof!

Issue #78, opened 2017-07-03
Relax validation upon outputPath and publicPath


davidmpaz:
In [`getContentBase(webpackConfig)`](https://github.com/symfony/webpack-encore/blob/v0.9.1/lib/config/path-util.js#L21) some validation is done against `outputPath` and  `publicPath`.

I think this validation is checking too much and restricting use cases possible. Could we relax this to be a warning and not error?

My use case is:
Project with custom publishing strategy. Webpack is compiling into intermediate path which is then "published" to real `publicPath`. As consequence I am facing the error thrown in [line 52](https://github.com/symfony/webpack-encore/blob/v0.9.1/lib/config/path-util.js#L52)

I think the validation should not be affected by outputPath since this can (or shall we allow it?) be anything.

I have found some other issue #59 and its PR #66 that might be related to same problem.

Does all this makes sense?

davidmpaz:
As an update here I can say that by commenting the [line](https://github.com/symfony/webpack-encore/blob/v0.9.1/lib/config/path-util.js#L52) throwing the error I can get my webpack-dev-server working normally.

It is possible to get a warning or some message instead stopping execution?

thanks in advance

weaverryan:
@davidmpaz I'm certainly open to this - Encore purposely is "restrictive" now, to see if there are any legit use-cases.

Could you post the important parts of your `webpack.config.js` file and tell me a bit more about what exactly you're doing with respect to the intermediate path? And is the only issue when using the dev-server, or always?

davidmpaz:
Hi @weaverryan 

thanks for answering. I can post relevant part but it is not more than that explained. See:
```
let publicPath = '/static/theme.dir'
Encore
    .setOutputPath('./static')
    .setPublicPath(publicPath)
    .setManifestKeyPrefix(publicPath)
```
the scenario is a framework that have many modules, themes are also modules. Every module can export their own static assets to an **`static` directory inside module**. Build/Deploy process take cares of putting those assets from every module in public directory accessible by application server, under `webRooot/static/[module.name]`. That's basically all.

I am affected only when trying to use dev-server. If I try to satisfy validation for making dev-server work, then I have a side effect under the manifest plugin, in manifest all urls are wrong.

Currently the workaround is to modify configuration after encore generation. I need to modify dev-server and manifest plugin settings as well to make all work like expected. This is ok for me, but when trying to introduce a whole development stack to colleagues this could be a draw back, it is always good when things looks natural and not hacky :)
```
[...]

PR #96, opened 2017-07-15
Relaxed public path requirements with dev-server

### LINKED ISSUE #59: dev-server inside docker?

weaverryan:
Fixes #59 and finishes #66.

* Adds a new `--keep-public-path` option for `dev-server`. When used, your `publicPath` is not prefixed with the dev server URL. For #59, this means you can use `setPublicPath('/build')` with the dev-server, and your assets will remain local (i.e. `/build/main.js` instead of `http://localhost:8080/build/main.js`).

* It is now possible to pass an absolute URL to `setPublicPath()` without an error when using `dev-server`. But, we issue a big warning, because this means your assets will point to to that absolute URL, instead of to the dev-server (which for most setups, is not what you want).

@samjarrett I'd love to confirm that this would solve your issue in Docker :).

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -10,7 +10,6 @@
 
 'use strict';
 
-const path = require('path');
 const parseRuntime = require('../lib/config/parse-runtime');
 const context = require('../lib/context');
 const chalk = require('chalk');
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -62,11 +62,14 @@ function showUsageInstructions() {
     console.log('Commands:');
     console.log(`    ${chalk.green('dev')}        : runs webpack for development`);
     console.log('       - Supports any webpack options (e.g. --watch)');
+    console.log();
     console.log(`    ${chalk.green('dev-server')} : runs webpack-dev-server`);
     console.log(`       - ${chalk.yellow('--host')} The hostname/ip address the webpack-dev-server will bind to`);
     console.log(`       - ${chalk.yellow('--port')} The port the webpack-dev-server will bind to`);
     console.log(`       - ${chalk.yellow('--hot')}  Enable HMR on webpack-dev-server`);
+    console.log(`       - ${chalk.yellow('--keep-public-path')} Do not change the public path (it is usually prefixed by the dev server URL)`);
     console.log('       - Supports any webpack-dev-server options');
+    console.log();
     console.log(`    ${chalk.green('production')} : runs webpack for production`);
     console.log('       - Supports any webpack options (e.g. --watch)');
     console.log();
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ const configGenerator = require('./lib/config-generator');
 const validator = require('./lib/config/validator');
 const PrettyError = require('pretty-error');
 const runtimeConfig = require('./lib/context').runtimeConfig;
+const logger = require('./lib/logger');
 
 // at this time, the encore executable should have set the runtimeConfig
 if (!runtimeConfig) {
@@ -21,6 +22,9 @@ if (!runtimeConfig) {
 }
 
 let webpackConfig = new WebpackConfig(runtimeConfig);
+if (runtimeConfig.verbose) {
+    logger.verbose();
+}
 
 module.exports = {
     /**
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -9,6 +9,7 @@
 
 'use strict';
 
+const chalk = require('chalk');
 const path = require('path');
 const fs = require('fs');
 
@@ -94,7 +95,8 @@ class WebpackConfig {
          */
         if (publicPath.includes('://')) {
             if (this.useDevServer()) {
-                throw new Error('You cannot pass an absolute URL to setPublicPath() and use the dev-server at the same time. Try using Encore.isProduction() to only configure your absolute publicPath for production.');
+                console.log(chalk.bgYellow.black(' WARNING ') + chalk.yellow(' You should not pass an absolute URL to setPublicPath() and use the dev-server at the same time'));
+                console.log();
             }
         } else {
             if (publicPath.indexOf('/') !== 0) {
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -131,8 +131,8 @@ class WebpackConfig {
      * @returns {string}
      */
     getRealPublicPath() {
-        // if we're using webpack-dev-server, use it & add the publicPath
-        if (this.useDevServer()) {
+        // if we're using webpack-dev-server and have no absolute url, use it & add the publicPath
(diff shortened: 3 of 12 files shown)


samjarrett:
Hey @weaverryan thanks for the heads-up! I gave this a go with config like this:

```js
if (!Encore.isProduction()) {
    Encore
        .setPublicPath('http://dev.vcap.me:8080')
        .setManifestKeyPrefix('/assets/build/')
    ;
}
```

and when I run the command with `encore dev-server --host 0.0.0.0`, I get the warning on the command line about using setPublicPath() and dev-server, but it seems like the config was being ignored?

Example - paths to CSS are being generated as relative:

```html
<link href="/assets/build/global.css" rel="stylesheet">
```

if I disable that in my path to CSS is this:

```html
<link href="http://0.0.0.0:8080/assets/build/global.css" rel="stylesheet">
```

have I missed something?

weaverryan:
Thanks @samjarrett! Hmm... I bet I missed one small detail - i'll take a look at this soon and ping back when I'm sure it's ready :)

davidmpaz:
Hi @weaverryan,
just a question, will this PR also address #78 ? ... from:

> When used, your publicPath is not prefixed with the dev server URL

I think this will create some troubles when using live reloading with Browser Sync. Resources are not written in file system but kept in memory. 

So do we need that url: `http://localhost:8080` as prefix when developing or do we need to set it as absolute url in order to get latest updates from compilations?

How do you think i could do with this PR to address #78 ?

davidmpaz:
@weaverryan, got it thanks!!

weaverryan:
My apologies @samjarrett - this is (and was) totally ready to go after all :). There's one mistake in your script:

```js
if (!Encore.isProduction()) {
    Encore
        .setPublicPath('http://dev.vcap.me:8080')
        // I removed the opening slash
        .setManifestKeyPrefix('assets/build/')
    ;
}
```

The manifest key normally doesn't start with a `/` - we make sure of that. So, I'm guessing (assuming you're using Symfony) that your template looks like this:

```twig
<script src="{{ asset('assets/build/main.js') }}">
```

Because you had `.setManifestKeyPrefix('/assets/build/')`, it changed the keys in `manifest.json` to start with a `/`, and so the path (`assets/build/main.js`) wasn't found in `manifest.json` and wasn't transformed (i.e. it was left alone).

I'm going to make a PR to guarantee that `setManifestKeyPrefix()` has no opening `/`. I originally allowed this... to add a bit more flexibility. But it's going to mess people up I think.

Soooo, make this one line change, and let me know how things look. I appreciate it!

samjarrett:
Works well @weaverryan, thanks!

weaverryan:
Cheers - thanks for checking @samjarrett!

adrienbrault:
@weaverryan I can't figure out how to use this feature, I keep getting `Unknown argument: keep-public-path`, see the output below:

```
# ./node_modules/.bin/encore dev-server --keep-public-path --disable-host-check --port 8204 --host 0.0.0.0

webpack-dev-server 2.9.1
webpack 3.6.0
Usage: https://webpack.js.org/configuration/dev-server/

Config options:
  --config       Path to the config file
                         [string] [default: webpack.config.js or webpackfile.js]
...
  --open-page    Open default browser with the specified page           [string]

Unknown argument: keep-public-path
```

even though the keep-public-path option is shown in the --help:
```
# ./node_modules/.bin/encore dev-server --help
usage encore [dev|production|dev-server]

encore is a thin executable around the webpack or webpack-dev-server executables

Commands:
    dev        : runs webpack for development
       - Supports any webpack options (e.g. --watch)

    dev-server : runs webpack-dev-server
       - --host The hostname/ip address the webpack-dev-server will bind to
       - --port The port the webpack-dev-server will bind to
       - --hot  Enable HMR on webpack-dev-server
       - --keep-public-path Do not change the public path (it is usually prefixed by the dev server URL)
       - Supports any webpack-dev-server options

    production : runs webpack for production
       - Supports any webpack options (e.g. --watch)

    encore dev --watch
    encore dev-server
    encore production

webpack-dev-server 2.9.1
webpack 3.6.0
Usage: https://webpack.js.org/configuration/dev-server/

Config options:
...
```

Lyrkan:
@adrienbrault It's caused by the fact that the `--keep-public-path` argument is supposed to be handled by Encore only but `webpack-dev-server` picks it up and doesn't recognize it. Based on webpack/webpack#2254 custom arguments need to be prefixed by `env.` (e.g. `env.keep-public-path`).

Unless I'm missing something too it didn't seem to work in the version it was released in either (v0.12.0), would you mind opening an issue about it ?

weaverryan:
@Lyrkan Are you saying that the `--keep-public-path` is broken everywhere? This definitely worked when I merged - I was testing it directly. But, @adrienbrault's output look like a bug.

Lyrkan:
@weaverryan Looks like it...

```
$ mkdir encore-public-path && cd encore-public-path

$ yarn add --dev @symfony/webpack-encore@0.12.0
yarn add v1.0.2
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
success Saved 603 new dependencies.
├─ @symfony/webpack-encore@0.12.0
(...)
Done in 62.17s.

$ ./node_modules/.bin/encore dev-server --keep-public-path

webpack-dev-server 2.9.1
webpack 3.6.0
Usage: https://webpack.js.org/configuration/dev-server/

Config options:
(...)
Unknown argument: keep-public-path
```

I don't know what could've caused it to break... maybe a dependency that got updated? The issue that I referenced in my previous comment seems quite old though.

pscheit:
glad I found this thread.
@weaverryan you said it's and edge case and for 95%  this should work. Well maybe when 5% only use docker. Because this is what happens:

When I run the wepack-dev-server in the docker container. Because it should not bind to localhost (and would be inaccessible from docker-host)  we HAVE to use --host 0.0.0.0. So that the manifest.json will always contain: http:/0.0.0.0 , This makes no sense in any scenario, right?

Shouldn't setting the publicPath be mentioned here:
https://symfony.com/doc/current/frontend/encore/dev-server.html

best regards
philipp

weaverryan:
Yea - it almost definitely should. And we should mention Docker specifically. Could you open a doc PR or issue and dump your information there? It would be really helpful :)

pscheit:
i will

Issue #87, opened 2017-07-11
Make clean-webpack-plugin configurable


weaverryan:
From a tweet! 
https://twitter.com/DevCircus/status/884574052127518720

Issue #79, opened 2017-07-03
Make UglifyJsPlugin optional/configurable


allfreelancers:
HI!
// todo - options here should be configurable
new webpack.optimize.UglifyJsPlugin({
      sourceMap: this.webpackConfig.useSourceMaps
})

Can you make it optional/configurable? I have an error:
app.b452c69496e2c9f0cbd2.js from UglifyJs

stof:
what is the error message you get ?

allfreelancers:
Error :
app.8eaba82891470d4ca570.js from UglifyJs
Unexpected token: punc (() [app.8eaba82891470d4ca570.js:204,8]


The solution of the problem is possible:
https://github.com/webpack/webpack/issues/1542

stof:
yeah, UglifyJS currently does not support minifying ES6 code. You need to transpile it to ES5 with Babel first.

weaverryan:
Closing - see #139 also for more details

Goatfried:
For anyone else that comes here looking for a solution:
If you use babel-preset-env, set the option [targets.uglify](https://babeljs.io/docs/plugins/preset-env/#optionstargets-uglify) to true in your babel presets.

Issue #65, opened 2017-06-30
Add HappyPack support


davidmpaz:
From README.

> [HappyPack](https://github.com/amireh/happypack) makes webpack builds faster by allowing you to transform multiple files in parallel.

Would be nice if we bring ideas on how to integrate best this, since it changes the way the loaders are declared. And also is good to get agreement on degree of integration.

Some points:
* When `happypack` is enable. All loaders will use it? or do we want to use it for some loaders instead?
* The same apply to `happypack` configuration options. Ex. Same amount of threads for all loaders?
* All of the previous influences on how many plugin instances will be added to configuration.

Some resources for reading
https://github.com/amireh/happypack/tree/master/examples

My five cents goes for a **per loader** configuration. Every loader **can use it or not** when enabled. this will add as many plugin instances as loaders we use, but in my opinion provides flexibility on loaders usage.

thanks in advance.

weaverryan:
Probably "per loader"... though I'm not sure what the config would look like to say "yes ts-loader, no to babel". Also, it looks like some loaders might have issues or be unsupported: https://github.com/amireh/happypack/wiki/Loader-Compatibility-List... and there "may" be issues  with ExtractText? https://github.com/amireh/happypack/issues/119#issuecomment-281892153. We could also apply it to all of the loaders that we can... all the time. And then see if we need the ability to enable/disable it for some loaders.

> The same apply to happypack configuration options. Ex. Same amount of threads for all loaders?

We should have 1 global HappyPack configuration, at least to start.

Overall, my biggest concerns is if this will introduce a lot of WTF, for small performance gains. Since HappyPack works with loader X, but not Y... it gives me a little bit of that impression. But, I've never used it :)

davidmpaz:
I see you point :) ...

The use case I though of was to pass happypack configuration when `enable***Loader()` calls, if configuration is given we set up plugin for that loader, do nothing otherwise. This approach would even allow to introduce happypack incrementally for loaders. 

My experience have been only with TypeScript. Currently my code base is not that big, popular vendors like bootstrap, jquery, etc. plus couple of modules, but this will grow more likely. Even with small code base time got reduced from 3-5 secs to 1s steady. So far for TypeScript works perfectly for me

Maybe we should wait until more support is added to happypack?

weaverryan:
[...]

Issue #63, opened 2017-06-30
Add support for separate type checking in Type Script


davidmpaz:
According to [this section](https://github.com/TypeStrong/ts-loader#loader-options) we can improve significantly the compilation speed by setting `transpileOnly = true`, though this require using another plugin for the type checking which does it in a separate process.

Until now I have been using [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin). I was planning to make a PR for adding it.

Today I also found [ts-checker-webpack-plugin](https://www.npmjs.com/package/ts-checker-webpack-plugin) but plugin doesn't seem to be as healthy as the other one, still exists to solve some issues with the previous.

Could we discuss to bring ideas to the table?

thanks in advance.

weaverryan:
I would use whatever is officially recommended by TS, which is `fork-ts-checker-webpack-plugin` (since it's included in their ts-loader examples): https://github.com/TypeStrong/ts-loader

Other than that, it seems to make sense. If the user opts into this... then we can add the plugin :). To opt in... it seems like we're taking Encore down this path of having Encore-specific options after the callback:

```js
.enableTypeScriptLoader(function() {}, {
    fork_ts_checker_plugin: true
});
```

As perfectly consistent as that is, other ideas are welcome :).

davidmpaz:
Which plugin to use is then clear. Regarding implementation I had in mind 2 solutions:

1. Add the plugin automagically when user configure `ts-loader` with `transpileOnly: true`. Options would have sensible defaults.
2. Add explicit API for `enableForkedTypesChecking(pluginConfig)`. This would enforce settings in `ts-loaders` for instance: `transpileOnly: true`. As well as any validation needed.

My point .1 I think would be problematic the same as your proposed solution in the way that plugin configuration will be hard (point 1) or we will be duplicating options on the way (Encore-specific options after the callback). Still would be the easiest way to use out of the box.

Point 2 add more verbosity and allow for more plugin configuration later on. Calling the API would store options and flag and when configuring `ts-loader` plugin would be added.

weaverryan:
I like option 2!

PR #99, opened 2017-07-19
Refactor plugin configuration

davidmpaz:
Extract encore built in plugin configuration logic into plugin utils.

First step toward allowing to be able to configure plugins through options and have more modular integration for plugins inside encore. From now on only public API with proper options is needed,
for most cases.

This changes are internals and tries to improve plugin architecture inside encore. Please take a look on it and feedback are welcome :). So far I see improvements for several issues:

* #2 - Point .2 ...for plugins we should do the same as for loaders, from point of view of module split (utilities). 
* #63 - Better (more modular) integration of the plugin as well its configuration.
* #65 - Same as #63
* #79 - Now it should be possible to pass options and/or check for flags to apply plugin or not.
* #87 - Some small steps to make clean plugin more configurable. Public API still need to be changed though.

thanks in advance.


weaverryan:
I'll check this out tonight! Excited about this!

weaverryan:
Ok, just checked this out! Breaking things into smaller pieces is a win :). To be honest, I'm struggling on a few of the details about "how best" to do this... but since it's internal stuff, we shouldn't sweat it too much.

> From now on only public API with proper options is needed, for most cases.

What do you mean by this?

About the individual plugins files, naming the method `getPlugins()` and allowing each to return an array of plugins looks weird to me (I realize you based it off of my `getLoaders()`... and it felt weird there too :) ). What if each plugin file exports just a function... and what if we pass `plugins` *to* that function and have it modify the array. So:

```js
// clean.js
module.exports = function(plugins, webpackConfig, paths, cleanUpOptions = {}) {
    let config = Object.assign({}, cleanUpOptions, {
        root: webpackConfig.outputPath,
        verbose: false,
    });

    plugins.push(new CleanWebpackPlugin(paths, config));
};
```

`config-generator.js` would look like this:

```js
if (this.webpackConfig.cleanupOutput) {
    cleanPluginUtil(plugins, this.webpackConfig, ['**/*']);
}
```

Actually, we could go a step further, and include the logic of whether or not a plugin should be enabled, inside of that file:

```js
// clean.js
module.exports = function(plugins, webpackConfig, paths, cleanUpOptions = {}) {
    if (!this.webpackConfig.cleanupOutput) {
        return;
    }

    let config = // ...
    // rest of file
};
```

It still doesn't feel perfect, but maybe good enough for the next step :).

lib/plugins/dev-server.js
```diff
@@ -0,0 +1,28 @@
+/*
```
weaverryan: This isn't really a `dev-server` plugin... it's a "AssetOutputDisplayPlugin", which happens to be disabled when the dev-server is used.

lib/plugins/dev-server.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {WebpackConfig} webpackConfig
+ * @param {Array} paths to clean
+ * @param {Object} cleanUpOptions
+ * @return {Array} of plugins to add to webpack
+ */
+module.exports = {
+    getPlugins(webpackConfig) {
+        const outputPath = pathUtil.getRelativeOutputPath(webpackConfig);
+        const friendlyErrorsPlugin = FriendlyErrorsPlugin.getPlugins();
+        return [new AssetOutputDisplayPlugin(outputPath, friendlyErrorsPlugin[0])];
```
weaverryan: This is clever, but a problem :). Because `FriendlyErrorsPlugin.getPlugins()` actually instantiates a `new FriendlyErrorsWebpackPlugin` internally, when `FriendlyErrorsPlugin.getPlugins()` is called by `config-generator`, it will receive one instance. When it's called here, it will receive a different instance. I "got away: with doing stuff like this with the loaders, because those return static config objects, not new instances of objects. In practice, this might not be a problem, but we can do better :).

The only idea I can think of right now, is to change `plugins` in `config-generator` to a hash - i.e. create keys for each plugin - e.g.

```js
// old
plugins.push(new FriendlyErrorsPlugin());

// now
plugins.push('friendly_errors_plugin', new FriendlyErrorsPlugin());
```

then we could use that internal key to find other plugins. At the bottom of `buildPluginsConfig()` in `config-generator`, we would still return an actual array (not a hash/object). I don't love this magic, invented `friendly_errors_plugin` machine name... but it would work, and would still be an internal implementation detail, at least for now (i.e. not exposed to users in any way).

lib/plugins/dev-server.js
davidmpaz: oops !! my bad. I'll update that.

lib/plugins/dev-server.js
davidmpaz: I see your point. Is this issue present in general for all plugins or it is just this particular case with the  `FriendlyErrorsWebpackPlugin`?

Regarding hashed solution:

>... this magic, invented `friendly_errors_plugin` machine name...

I don't like it either. This case looks to me like a good scenario for a **singleton**. I did some research about the topic on the `nodejs` land. I did found some resources:
* [post explaining approaches](https://derickbailey.com/2016/03/09/creating-a-true-singleton-in-node-js-with-es6-symbols/)
* [module implementing a solution](https://www.npmjs.com/package/create-singleton)
* [another implementation approach](https://simplapi.wordpress.com/2012/05/14/node-js-singleton-structure/)

What do you think if we try to implement one of those solutions for that `FriendlyErrorsPlugin`? 

I think this is the best way we can guaranty the single instance we need.

davidmpaz:
To answer your first comment

Regarding: 

>From now on only public API with proper options is needed, for most cases.

I did mean that with this approach (broken into smaller pieces). Any future plugin (`fork-ts-checker-webpack-plugin`, `browser-sync`, etc..) can be easily integrated by creating another plugin module and adding respective public API. It was not a comment for internally used plugins, those are already integrated it ins owns way.

For:
>What if each plugin file exports just a function... and what if we pass plugins to that function and have it modify the array.

I like this idea far more than current one, actually I was trying to get some result similar to an interface implementation, as long as all plugins integration follow same pattern i don't see any problem. To your suggestion though I would only change one thing, function signature:
```
module.exports = function(webpackConfig, paths, cleanUpOptions = {}) {
```
I think we don't need the plugin array any more because `webpackConfig` already contain plugins on its own, plugins that are added later on by:
```
this.webpackConfig.plugins.forEach(function(plugin) {
    plugins.push(plugin);
});
```
we could push to `webpackConfig.plugins` directly and get rid of that intermediate piece of code. Since plugin execution in webpack matters, we could ask as *convention* to users to add any custom plugin through `addPlugin()` api at the end of `Encore` calls.

What do you think?

lib/plugins/dev-server.js
weaverryan: > This case looks to me like a good scenario for a singleton

Yes.... but no :). We allow you to generate multiple webpack configs. And to be strict about it, each separate webpack config should have its own instances of plugins. 

So, yes, we need to make sure that we don't create multiple `FriendlyErrorsWebpackPlugin` while generating *one* webpack config... but *do* need to create distinct instances, each time we `ConfigGenerator.getWebpackConfig()`. If you can take a shot at that, perfect :).

weaverryan: > In which way a singleton will be different to the original code when we call Encore.reset() to generate a new configuration?

Honestly, in practice, it probably makes not difference for `FriendlyErrorsWebpackPlugin`. In fact, it's not clear to me what the proper behavior is for this plugin specifically when it comes to multi webpack config setups - there's an option question about it: https://github.com/geowarin/friendly-errors-webpack-plugin/issues/56.

Could we just pass in the FriendlyErrorsPlugin instance as a dependency - i.e. add a `friendlyErrorsWebpackPlugin` argument to this plugin and pass the instance in from `config-generator`? It would work and no complexity or magic.

weaverryan:
> I did mean that with this approach (broken into smaller pieces). Any future plugin (fork-ts-checker-webpack-plugin, browser-sync, etc..) can be easily integrated by creating another plugin module and adding respective public API

I understand!

> I think we don't need the plugin array any more because webpackConfig already contain plugins on its own, plugins that are added later on by:

This sounds good... but so far, when we call `ConfigGenerator.getWebpackConfig()`, we never modify the `WebpackConfig`. I think we should continue to *not* modify this object: it's our input only.

Do we agree on enough to get this to a good spot? I feel pretty comfortable - it's internal stuff anyways, which we can continue to iterate on. I don't want to delay this PR too much more :).

Thanks!

davidmpaz:
I feel also comfortable with the result. Thanks for the clarifications, did helped a lot.

weaverryan:
Very nice! Thank you David!

PR #101, opened 2017-07-19
Forked typescript type checking

davidmpaz:
This PR fixes: #63 it should be possible to merge also with #99 directly.

Open points:

Regarding point .1 in #2

> If this plugin is added. we need to check dependency is added using loader-features.js. This module name is getting name from loaders, we are not adding a loader here but a plugin. Maybe some more general module is needed for checking features?

After implementing this I realize that as  soon as you require a module which doesn't exist node throws an error, we don't even go through error catch from encore. So possible solutions I see for this pending point are:

1. We don't refactor `loader-features.js` file and we still will get a loud error when trying to use some plugin not installed. From error is clear whats need to be done.
2. We refactor `loader-features.js` to be named more general and we include plugins dependencies also there as intended initially. But this way we need to check for missing packages earlier in `WebpackConfig.js` like I am doing now in this PR.

I like either solutions the same, I would like to get feedback on what do you guys think about it?

thanks in advance.


weaverryan:
I know what I'd like: option 2:

> We refactor loader-features.js to be named more general and we include plugins dependencies also there as intended initially. But this way we need to check for missing packages earlier in WebpackConfig.js like I am doing now in this PR.

Even if the normal error from Node is pretty good, I'd rather be able to make the error specific to Encore - e.g. "You run yarn XYZ in order to use enableForkedTypeScriptTypesChecking".

I'll leave a comment that might help with the issue of where you place the "check" for the necessary packages.

Cheers!

lib/loaders/typescript.js
```diff
@@ -11,6 +11,7 @@
 
 const loaderFeatures = require('../loader-features');
 const babelLoader = require('./babel');
+const forkedTypesPluginUtil = require('../plugins/forked-ts-types');
```
weaverryan: When you require this file, it requires `fork-ts-checker-webpack-plugin`. And THAT causes the big Node error in case `fork-ts-checker-webpack-plugin` isn't installed. What you can do is move this `require` statement down - it doesn't need to be on top:

if (webpackConfig.useForkedTypeScriptTypeChecking) {
    loaderFeatures.ensureLoaderPackagesExist('forkedtypecheck');
    
    const forkedTypesPluginUtil = require('../plugins/forked-ts-types');
}

require is most commonly done on top, but it's not a requirement.

davidmpaz: That was my first solution. I did not commit it because I was hidding a dependency in the execution path of the library. I guess that since we use it to report back an error it should be fine.

I'll take care of it. thanks!!

davidmpaz:
Btw @weaverryan ... any idea how can I deal with that timeout error I receive from ``functional.js` tests on appveyor ? I am considering now whether remove that test or not :(

davidmpaz:
I believe that this timeout error on **Appveyor** and now **Travis**:
```
1) Functional tests using webpack Basic scenarios. TypeScript is compiled and type checking is done in separate process!:
     Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
 is related to the cached `node_modules` directory. I had same errors locally in non related tests ([like now in travis](https://travis-ci.org/symfony/webpack-encore/builds/256071165?utm_source=github_status&utm_medium=notification)) when changing branches which uses differents modules, after an `npm install` error was gone.

I will ignore them, after merging everything should be gone.

weaverryan:
@davidmpaz Now that I look at this (nice work btw), is there any disadvantage to just *always* adding the forked type checking, if `fork-ts-checker-webpack-plugin` is enabled? I'd like Encore to try to automatically apply as many sane optimizations as it can. So, when you apply ts loader (but don't have `fork-ts-checker-webpack-plugin` installed), we could add a "recommendation" `console.log()` that you'll have better performance if you install it (this would be the first time we have one of these "recommendations" comments - but it's an idea I'd like to do more to help educate the user).

davidmpaz:
thanks. Regarding disadvantages I can see in their docs:
>It's very important to be aware that this plugin uses typescript's, not webpack's modules resolution. It means that you have to setup tsconfig.json correctly. For example if you set files: ['./src/someFile.ts'] in tsconfig.json, this plugin will check only someFile.ts for semantic errors. It's because of performance.

And i did noticed that the plugin expect a `tsconfig.json` [file from here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/blob/v0.2.7/lib/index.js#L90) if not, error is thrown. As far I remember for the `ts-loader`  is recommended but optional, not a requirement.

Besides that, nothing more come to my mind.

index.js
```diff
@@ -377,6 +377,22 @@ module.exports = {
     },
 
     /**
+     * Call this to enable forked type checking for TypeScript loader
+     * https://github.com/TypeStrong/ts-loader/blob/v2.3.0/README.md#faster-builds
+     *
+     * This is a build optimization API to reduce build times.
+     *
+     *
```
weaverryan: extra line here

index.js
```diff
@@ -377,6 +377,22 @@ module.exports = {
...
+     * @param {Object} forkedTypeScriptTypesCheckOptions
+     * @return {exports}
+     */
+    enableForkedTypeScriptTypesChecking(forkedTypeScriptTypesCheckOptions = {}) {
```
weaverryan: We've made these options arguments callbacks everywhere else. Though it feels unnecessary here, I think we should do the same

lib/WebpackConfig.js
```diff
@@ -243,6 +245,16 @@ class WebpackConfig {
         this.tsConfigurationCallback = callback;
     }
 
+    enableForkedTypeScriptTypesChecking(forkedTypeScriptTypesCheckOptions = {}) {
+
+        if (!this.useTypeScriptLoader) {
+            throw new Error('TypeScript loader must be enabled. Call `enableTypeScriptLoader()` first.');
+        }
```
weaverryan: I can't remember if I made this comment before or not, but checks like this need to be moved to be during the config generation process. It should be fine to call `enableTypeScriptLoader()` and `enableForkedTypeScriptTypesChecking()` in whatever order.

lib/loader-features.js
```diff
@@ -41,6 +41,11 @@ const loaderFeatures = {
         packages: ['typescript', 'ts-loader'],
         description: 'process TypeScript files'
     },
+    forkedtypecheck: {
+        method: 'enableForkedTypeScriptTypesChecking()',
+        packages: ['typescript', 'ts-loader', 'fork-ts-checker-webpack-plugin'],
+        description: 'check TypeScript types in separate process'
```
weaverryan: > in a separate...

lib/loaders/typescript.js
```diff
@@ -32,6 +32,17 @@ module.exports = {
...
+        // fork-ts-checker-webpack-plugin integration
+        if (webpackConfig.useForkedTypeScriptTypeChecking) {
+            loaderFeatures.ensureLoaderPackagesExist('forkedtypecheck');
+            // force transpileOnly to speed up
+            config.transpileOnly = true;
+
+            // add forked ts types plugin to the stack
+            const forkedTypesPluginUtil = require('../plugins/forked-ts-types');
+            forkedTypesPluginUtil(webpackConfig);
+        }
```
weaverryan: What about logging a recommendation here to enable this feature for faster type-checking? We now have a `logger.js` module that can be used for this (though I might add a new method on that - `recommendation()` - instead of using `debug()`.

package.json
```diff
@@ -34,6 +34,7 @@
     "css-loader": "^0.26.2",
     "extract-text-webpack-plugin": "^3.0.0",
     "file-loader": "^0.10.1",
+    "fork-ts-checker-webpack-plugin": "^0.2.7",
```
weaverryan: should be in `devDependencies`

test/WebpackConfig.js
```diff
@@ -303,6 +303,30 @@ describe('WebpackConfig object', () => {
...
+        });
+
+        it('Calling without TypeScript loader throws an error', () => {
+            const config = createConfig();
+            expect(() => {
+                config.enableForkedTypeScriptTypesChecking();
+            }).to.throw(
+                'TypeScript loader must be enabled. Call `enableTypeScriptLoader()` first.'
+            );
+        });
```
weaverryan: Good test, but should live somewhere else after the corresponding code moves.

test/functional.js
```diff
@@ -617,6 +617,40 @@ module.exports = {
...
+                    'main.js',
+                    'manifest.json'
+                ]);
+
+                done();
+
+                // TODO find a way to confirm forked is used. Maybe terminal output?
+                // Output:
+                // Starting type checking service...
+                // Using 1 worker with 2048MB memory limit
```
weaverryan: hmm. Well, we have other tests that confirm that the config is setup correctly. Is it enough to know just that this test didn't blow up? Is there any reasonable way it could blow up? I'm probably ok without adding anything additional.

index.js
davidmpaz: What line is surplussing ? :)

lib/loader-features.js
davidmpaz: If you refer to make a feature checker for plugins I have some work related to that. I just want to start including something in main stream before make another PR for the loader features itself. For now I used like it is.

lib/loaders/typescript.js
davidmpaz: When we get that `recommendation()` method we can add it yes.

package.json
davidmpaz: Will get a linting error then. I even reported that in Realytics/fork-ts-checker-webpack-plugin#34 but it seems to be right like that. 

Honestly I can not argue much about it since I still don't get right eslint rules :(

test/functional.js
davidmpaz: I don't know actually. That's why I did add the comment. Will check anyways.

index.js
weaverryan: We usually only have one empty (well `*`) line between the description and the start of the arguments. So, delete line 384 :).

lib/loaders/typescript.js
weaverryan: It's in #116 - we'll probably merge this first. If so, I'll add the warning after #116

package.json
weaverryan: Ah, this makes sense. Congrats! You were the first to hit this issue :). Basically, it's quite often for Encore to have optional features. Think about `sass-loader` - we don't have that in our `dependencies` because we force the user to add this manually. We add it in `devDependencies` just so we can use it in our tests.

But `fork-ts-checker-webpack-plugin` is different, because (unlike all the optional loaders), we explicitly *require* this module (for loaders, you just mention the loader name and webpack requires it later - so the linter doesn't pick up on this).

tl;dr; The linter is correct in a sense: Encore is using this package in its actual *code*, not just in its tests. But it's wrong because we're manually checking to make sure this package exists before using it.

So, add `// eslint-disable-line` after the `require` line in `forked-ts-types.js` to silence this lint problem. THEN put it in `devDependencies` :)

test/functional.js
weaverryan: Let's just remove these comments for now and leave the test as-is.

lib/plugins/forked-ts-types.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+let ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
```
Lyrkan: `const` instead of `let`

test/plugins/forked-ts-types.js
```diff
@@ -0,0 +1,41 @@
...
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+const expect = require('chai').expect;
+const WebpackConfig = require('../../lib/WebpackConfig');
+const RuntimeConfig = require('../../lib/config/RuntimeConfig');
+const tsLoader = require('../../lib/loaders/typescript');
+let ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
```
Lyrkan: `const` instead of `let`

lib/plugins/forked-ts-types.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @param {WebpackConfig} webpackConfig
+ * @return {void}
+ */
+module.exports = function(webpackConfig) {
+    let config = Object.assign({}, webpackConfig.forkedTypeScriptTypesCheckOptions);
```
Lyrkan: `const` instead of `let`

weaverryan:
Thank you David! If I have time this morning, I'll try to add the "recommendation" log to "advertise" this when you use ts-loader

PR #102, opened 2017-07-19
Fix missing tsloader return statement

ldiringer:
This commit (https://github.com/symfony/webpack-encore/commit/35d9e6f5c24f88ba2660d4e58407b1330aeb2640#diff-168726dbe96b3ce427e7fedce31bb0bc) removed inadvertently the `return this;` statement from the `enableTypeScriptLoader` function.

This prevents calling another function chained on `enableTypeScriptLoader` in a `webpack.config.js` file.

This PR just puts the `return this;` statement back.

### DIFF
--- a/index.js
+++ b/index.js
@@ -372,6 +372,8 @@ module.exports = {
      */
     enableTypeScriptLoader(callback = () => {}) {
         webpackConfig.enableTypeScriptLoader(callback);
+
+        return this;
     },
 
     /**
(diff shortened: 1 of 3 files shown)


davidmpaz:
We should definitely make a test for the pubic API, to prevents such cases in future. Do you think you could add some @ldiringer ?

davidmpaz:
what do you think if `api.js` file get renamed to `index.js`, that way it explicitly says what part is testing?

tests look great!. Thank you very much, i hope this get merged soon.

ldiringer:
All done. Thank you for the review @davidmpaz!

weaverryan:
Thank you @ldiringer.

weaverryan:
And thank you also for the test - that was very clearly needed.

PR #110, opened 2017-07-25
Always add a hash to the name of the files created by the default assets loaders

### LINKED ISSUE #73: File-loader not works with different images with same name

Lyrkan:
This PR modifies the configuration of the default assets loaders so a `[hash]` is always added to the output file names, even if versioning is disabled.

It basically prevents having two files with the same name (but in a different directory) overwriting each other during build.

This closes #73 and was further discussed in #103.


lib/config-generator.js
```diff
@@ -127,15 +127,15 @@ class ConfigGenerator {
                 test: /\.(png|jpg|jpeg|gif|ico|svg)$/,
                 loader: 'file-loader',
                 options: {
-                    name: `images/[name]${this.webpackConfig.useVersioning ? '.[hash]' : ''}.[ext]`,
+                    name: 'images/[name].[hash].[ext]',
```
weaverryan: What do you think about using a shorter hash - like `[hash:8]`? The longer seems overkill... I should have done this originally.

weaverryan:
One minor comment - but this looks great - really nice job with the test!

lib/config-generator.js
Lyrkan: Done, no idea what's up with AppVeyor though (same thing for #103)

weaverryan:
Thank you @Lyrkan! Test failure is from master and unrelated - I'll fix it :)

PR #114, opened 2017-07-26
Fixing a bug with webpack 3.4.0 with extra arg as an entry

### LINKED ISSUE #112: Encore 11.0 bug "cannot resolve dev"
Hello,

Using the same config as I did with Encore 0.9.1, when I updated to 0.11.0 and tried to run it, this happened:

```
./node_modules/.bin/encore dev
Running webpack ...

 ERROR  Failed to compile with 1 errors                                                                                                                                                                                              4:02:12 PM


 error

Entry module not found: Error: Can't resolve 'dev'
```

Did I do something wrong ?

weaverryan:
Fixes #112 

The updated way avoids triggering the "singleton" behavior or yargs. This is important so that later, when webpack uses yargs, it uses the updated `process.argv` (because we remove our extra "command" argument so that webpack doesn't see it).

I'm not sure exactly which dep/commit caused the issue, but this a more proper solution anyways.

Also added a smoke test for the encore binary.

### DIFF
--- a/test/bin/encore.js
+++ b/test/bin/encore.js
@@ -29,7 +29,7 @@ describe('bin/encore.js', function() {
             `
 const Encore = require('../../index.js');
 Encore
-    .setOutputPath('${testDir}/build')
+    .setOutputPath('/build')
     .setPublicPath('/build')
     .addEntry('main', './js/no_require')
 ;
--- a/test/bin/encore.js
+++ b/test/bin/encore.js
@@ -29,7 +29,7 @@ describe('bin/encore.js', function() {
             `
 const Encore = require('../../index.js');
 Encore
-    .setOutputPath('/build')
+    .setOutputPath('build/')
     .setPublicPath('/build')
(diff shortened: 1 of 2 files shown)


test/bin/encore.js
```diff
@@ -0,0 +1,50 @@
...
+    .setPublicPath('/build')
+    .addEntry('main', './js/no_require')
+;
+
+module.exports = Encore.getWebpackConfig();
+            `
+        );
+
+        const binPath = path.resolve(__dirname, '../', '../', 'bin', 'encore.js');
+        exec(`node ${binPath} dev --context=${testDir}`, { cwd: testDir }, (err, stdout, stderr) => {
```
stof: the 2 paths need to be escaped as they could contain spaces

weaverryan: Thanks - I just pushed an imperfect escape at sha: 172a670120e79dd6b5f6aaa65845f9892b982ea0 - which is good enough for test paths I think :)

PR #115, opened 2017-07-26
Add Encore.configureRuntimeEnvironment() method to the public API

### LINKED ISSUE #55: Improve error display when thrown inside Webpack.config.js
When an error is thrown inside `WebpackConfig.js`, it falls outside of the try-catch used inside of `getWebpackConfig()` (https://github.com/symfony/webpack-encore/blob/5ca96e94a6af8d9b43e6d6e106b7cf45d743a86e/index.js#L372-L383). This means the errors are really ugly :). If we're going to prettify the errors (which I like), we should do it also around the methods to WebpackConfig.js. We'll probably need to wrap each method in a try-catch (or do something clever to do this automatically).

### LINKED ISSUE #109: Get config for testing configuration (Karma)
Hello,

I'm using the vue-loader and i'd like to test unit my code. 

As shown in the webpack template https://github.com/vuejs-templates/webpack, you need to get the Webpack config for the karma.conf.js.

```js
//karma.conf.js

var webpackConfig = require('../../build/webpack.test.conf')
```

The problem is that getting the webpack config without using the encore binary doesn't work.
```
Invalid config file!
  Error: Are you trying to require index.js directly?
```
Is there any way to get the wepack config or is there an other way to setup karmaJs for webpack-encore ?

Thx

Lyrkan:
This PR fixes #109 by adding an `Encore.configureRuntimeEnvironment` and an `Encore.clearRuntimeEnvironment` method to the public API.

---

Basically, until now if you tried requiring `@symfony/webpack-encore` without using `./node_modules/.bin/encore`, it would throw an `Are you trying to require index.js directly?` error.

This prevented retrieving the Webpack configuration object easily, which can be needed in some cases (e.g. if you use [karma-webpack](https://github.com/webpack-contrib/karma-webpack#usage)).

With these changes, requiring `index.js` doesn't create a `WebpackConfig` instance anymore if the runtime environment isn't available, which removes the need of having a check at that time.

In order to deal with methods of the public API that need the `WebpackConfig` to be available, a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) of the API is exported instead of the real object. This proxy only allows calls to `configureRuntimeEnvironment` and `clearRuntimeEnvironment` if the `WebpackConfig` isn't initialized, or act as a passthrough if it is.

This also fixes #55 since it allows to add a `try { ... }  catch { ... }` quite easily around all the methods of the API.

The `Encore.clearRuntimeEnvironment()` (that nullifies `runtimeConfig` and `webpackConfig`) was needed to write tests but since I doubt it'll really be useful outside of that case let me know if you see another way to do it.

---

**Examples:**

```js
Encore
    .setOutputPath('build/')
    .setPublicPath('/')
    .addEntry('main', './src/index.ts');
```
If executed directly would result in the following error:
```
Error: Encore.setOutputPath() cannot be called yet because the runtime environment doesn't appear to be configured. Try calling Encore.configureRuntimeEnvironment() first.
```

Whereas the following would work:

```js
Encore
    .configureRuntimeEnvironment('dev-server')
    .setOutputPath('build/')
    .setPublicPath('/')
    .addEntry('main', './src/index.ts');
```

It is also possible to pass the same options that are available using the usual CLI tool using the second argument of `configureRuntimeEnvironment` and camel-cased names:

```js
Encore
    .configureRuntimeEnvironment('dev-server', {
        keepPublicPath: true,
        https: true,
    })
    .setOutputPath('build/')
    .setPublicPath('/')
    .addEntry('main', './src/index.ts');
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,20 +13,22 @@ const WebpackConfig = require('./lib/WebpackConfig');
 const configGenerator = require('./lib/config-generator');
 const validator = require('./lib/config/validator');
 const PrettyError = require('pretty-error');
-const runtimeConfig = require('./lib/context').runtimeConfig;
 const logger = require('./lib/logger');
+const parseRuntime = require('./lib/config/parse-runtime');
 
-// at this time, the encore executable should have set the runtimeConfig
-if (!runtimeConfig) {
-    throw new Error('Are you trying to require index.js directly?');
+let webpackConfig = null;
+let runtimeConfig = require('./lib/context').runtimeConfig;
+
+// If runtimeConfig is already set webpackConfig can directly
+// be initialized here.
+if (runtimeConfig) {
+    webpackConfig = new WebpackConfig(runtimeConfig);
+    if (runtimeConfig.verbose) {
+        logger.verbose();
+    }
 }
 
-let webpackConfig = new WebpackConfig(runtimeConfig);
-if (runtimeConfig.verbose) {
-    logger.verbose();
-}
-
-module.exports = {
+const publicApi = {
     /**
      * The directory where your files should be output.
      *
@@ -431,17 +433,9 @@ module.exports = {
      * @returns {*}
      */
     getWebpackConfig() {
-        try {
-            validator(webpackConfig);
+        validator(webpackConfig);
 
-            return configGenerator(webpackConfig);
-        } catch (error) {
-            // prettifies errors thrown by our library
-            const pe = new PrettyError();
-
-            console.log(pe.render(error));
-            process.exit(1); // eslint-disable-line
-        }
+        return configGenerator(webpackConfig);
     },
 
     /**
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -454,5 +448,99 @@ module.exports = {
...
+     *
+     * @param {string} environment
+     * @param {object} options
+     * @returns {exports}
+     */
+    configureRuntimeEnvironment(environment, options = {}) {
+        runtimeConfig = parseRuntime(
+            Object.assign(
+                {},
+                require('yargs/yargs')([environment]).argv,
```
weaverryan: I'm not sure I understand this. Is there a need to continue to read the command line options? The command line options could be *totally* different (since there is a different executable being run).

I see that parseRuntime() looks for `argv._[0]` for the environment/command. Is that what you're trying to supply here? If so, I think we could set that manually - e.g. `options._ = [environment]`. Or, we could re-work `parse-runtime.js` so that the `argv._[0]` is actually done in `encore.js` and `parse-runtime` looked like `function(command, options, cwd) {`.

index.js
```diff
@@ -454,5 +448,99 @@ module.exports = {
...
+                options
+            ),
+            process.cwd()
+        );
+
+        if (runtimeConfig.verbose) {
+            logger.verbose();
+        }
+
+        webpackConfig = new WebpackConfig(runtimeConfig);
```
weaverryan: Since this (and the logger lines) above are repeated at the top of the file, let's warp these in some function (i.e. some simple `function initializeWebpackConfig()` at the top of the file) that we call in both places...

index.js
Lyrkan: I added that part to make sure that the first parameter passed to `parseRuntime` had the same structure than the one usually supplied by yargs, it doesn't actually keep reading the CLI options since we provide a fake `process.argv` (see https://github.com/yargs/yargs/blob/master/docs/api.md#api).

I did `options._ = [environment]` at first but switched to calling the API after realizing that `yargs` adds other things to the `argv` array that may be used later by `parseRuntime` (e.g.: `$0`)

index.js
```diff
@@ -454,5 +448,99 @@ module.exports = {
...
+                'configureRuntimeEnvironment',
+                'clearRuntimeEnvironment',
+            ];
+
+            if (!webpackConfig && (safeMethods.indexOf(prop) === -1)) {
+                throw new Error(`Encore.${prop}() cannot be called yet because the runtime environment doesn't appear to be configured. Try calling Encore.configureRuntimeEnvironment() first.`);
+            } else {
+                // Either a safe method has been called or the webpackConfig
+                // object is already available. In this case act as a passthrough.
+                return (...parameters) => {
```
Lyrkan: It's needed in order to pass all the parameters to the real method: `const res = target[prop](...parameters);`

```js
// Real API method
const inner = (param1, param2) => {
    console.log(`param1 = ${param1}`);
    console.log(`param2 = ${param2}`);
};

// Method returned by the proxy
const wrapper = (...parameters) => {
    inner(...parameters);
};

wrapper(1, 2);

// Output:
// param1 = 1
// param2 = 2
```

test/index.js
```diff
@@ -44,7 +65,12 @@ describe('Public API', () => {
 
     describe('addEntry', () => {
 
+        it('should not be callable before the runtime environment has been configured', () => {
+            expect(() => api.addEntry('entry', 'main.js')).to.throw();
+        });
```
Lyrkan: Exactly what I thought after copy/pasting that part for every method of the API.

I wanted a 2nd opinion before removing everything :)

index.js
```diff
@@ -454,5 +448,99 @@ module.exports = {
...
+        runtimeConfig = null;
+        webpackConfig = null;
+    },
 };
+
+// Proxy the API in order to prevent calls to most of its methods
+// if the webpackConfig object hasn't been initialized yet.
+const publicApiProxy = new Proxy(publicApi, {
+    get: (target, prop) => {
+        if (typeof target[prop] === 'function') {
```
Lyrkan: Yep, I missed that part, I added a `return target[prop]` just in case.

weaverryan:
Merged! This is really, really great. Thanks @Lyrkan!

index.js
```diff
@@ -454,5 +448,99 @@ module.exports = {
...
+            // These methods of the public API can be called even if the
+            // webpackConfig object hasn't been initialized yet.
+            const safeMethods = [
+                'configureRuntimeEnvironment',
+                'clearRuntimeEnvironment',
+            ];
+
+            if (!webpackConfig && (safeMethods.indexOf(prop) === -1)) {
+                throw new Error(`Encore.${prop}() cannot be called yet because the runtime environment doesn't appear to be configured. Try calling Encore.configureRuntimeEnvironment() first.`);
```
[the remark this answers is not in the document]
cybernet: that message brought me here today, that was shown by phpstorm :)

PR #130, opened 2017-08-04
Add options callback to Encore.enablePostCssLoader()

### LINKED ISSUE #129: Question: How to set a custom path to the postcss.config.js file?
I'm using Symfony and I would like to move all config files to the `app/config` . I figured out that I can set the path to the `webpack.config.js` file by adding `--config [PATH]` to the encore script. Unfortunately I cannot currently figure out  how to set the path to the `postcss.config.js` file of `postcss-loader`. Without Encore I would set a `config` option in the config of `postcss-loader`, but how can I do that when Encore is managing the whole configuration? Is there any way?

Lyrkan:
This PR allows to add a callback as the first argument of `Encore.enablePostCssLoader()` in order to be able to modify the options of the `postcss-loader`.

As an example (fixes #129):

```js
Encore.enablePostCssLoader((options) => {
    options.config = {
        path: 'app/config/postcss.config.js'
    };
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -275,10 +275,20 @@ module.exports = {
      *
      * https://github.com/postcss/postcss-loader
      *
+     *     Encore.enablePostCssLoader();
+     *
+     * Or pass options to the loader
+     *
+     *     Encore.enablePostCssLoader(function(options) {
+     *         // https://github.com/postcss/postcss-loader#options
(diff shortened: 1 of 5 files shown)


weaverryan:
This looks perfect. Very nice work! Thank you @Lyrkan!

PR #131, opened 2017-08-04
Remove duplicate addPlugin test case

Lyrkan:
This PR removes a duplicate test case in `WebpackConfig.js`:

* [first instance](https://github.com/symfony/webpack-encore/blob/v0.12.0/test/WebpackConfig.js#L335-L346)
* [second instance](https://github.com/symfony/webpack-encore/blob/v0.12.0/test/WebpackConfig.js#L368-L379)

### DIFF
--- a/test/WebpackConfig.js
+++ b/test/WebpackConfig.js
@@ -375,19 +375,6 @@ describe('WebpackConfig object', () => {
         });
     });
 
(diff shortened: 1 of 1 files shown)


weaverryan:
We were just being extra careful ;). Good catch and thank you!

PR #134, opened 2017-08-05
Add options callback to Encore.enableLessLoader()

### LINKED ISSUE #118: Less-Loader, we need the "relativeUrls" option !
The Less-loader has the option ''.
I need to use it to use a library where assets(fonts, images...) are not relative.

to make it work, I can modify your file 'lib/loaders/less.js' and add `relativeUrls: false,`.
Example : 
```
...cssLoader.getLoaders(webpackConfig, ignorePostCssLoader),
            {
                loader: 'less-loader',
                options: {
                    relativeUrls: false,
                    sourceMap: webpackConfig.useSourceMaps
                }
            },
``` 

I would be nice if that option could be an option of your 'enableLessLoader' function ! 
Example : 
```
.enableSassLoader(function(sassOptions) {}, {
        relativeUrls: false
    })
```

Lyrkan:
This PR allows to add a callback as the first argument of `Encore.enableLessLoader()` (fixes #118).
I didn't notice that one was missing too when working on #130.

Example :

```js
Encore.enableLessLoader((options) => {
    options.relativeUrls = false;
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -330,10 +330,21 @@ module.exports = {
     /**
      * Call this if you plan on loading less files.
      *
+     *     Encore.enableLessLoader();
+     *
+     * Or pass options to the loader
+     *
+     *     Encore.enableLessLoader(function(options) {
+     *         // https://github.com/webpack-contrib/less-loader#examples
+     *         // http://lesscss.org/usage/#command-line-usage-options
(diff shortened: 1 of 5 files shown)


weaverryan:
Brilliant! Thank you @Lyrkan!

Issue #136, opened 2017-08-09
Consider reverting #110


jderusse:
Versionning may be disabled at the user choice. Encore shouldn't ignore this choice to fix a technical naming issue.

For the final user, getting the bad file, may be better than no file at all (think about favicon.ico for instance). And IMO it shouldn't be a choice made be encore for you.

Please consider other alternatives like using the `[path]` when versioning is disabled or using a hash based on the filepath instead of the file content

Lyrkan:
Hi @jderusse, 

A bit of context about why we did that change: with the old behavior if you had two file with the same name in different subfolders (e.g. `test1/test.png` and `test2/test.png`) one would end up overwriting the other silently during builds (bug #73). The fix was to add the `[hash]` of the file to filenames.

Is there really a case where having a hash in the filename would cause an issue? Which problem would you have with a favicon for instance?

About considering alternatives:
* `[path]`: we talked about it in #103 and didn't see any benefit over just adding `[hash]`
* using a hash based on the filepath: I don't think you can do that using `file-loader` but if I'm wrong feel free to correct me

Note that for very specific cases you can also require your images using the `file-loader?name=[name].[ext]!./image.png` syntax.

weaverryan:
Yo @jderusse!

I'm open to reverting if that's the right choice. But I'm also curious about your setup: are you referencing `favicon.ico` from somewhere in webpack (just to use that example)? Usually, if you're referencing an image or font from somewhere processed by Webpack (a CSS file or even a JS file), you don't care about the final path: Webpack rewrites your CSS to point to the new path, wherever it is. Even in JavaScript, if you reference an image, the new path to that image is returned:

```js
// will be something like build/main.abc123.png
const imagePath = require('../images/main.png');
```

In other words, you shouldn't care about the final filename. And that's why we're curious about your setup :).

Cheers!

jderusse:
Thank you for your feedback.

Actually, I don't care about the file path for a developer point of view. We both agree on that point :-)

But as soon as I expose a path to my users this path belong to the legacy and have to be maintained. You never known when this file will be requested by the user. 
It could be few second between the time the user download the html page and the browser request the asset. Or several day if the image has been indexed by Google Image and a user click on it. Or several month if you include that image in an email.

[...]

PR #137, opened 2017-08-10
Add Encore.configureFilenames() method to the public API

### LINKED ISSUE #125: Versioning file format configuration
Hello,

Sorry if it is not the right place for this recommandation.
We are trying Webpack Encore and wanted to avoid having double "." in the output filename when using versioning assets and let the manifest file out of the web (or public) folder.

First, here is how we could resolve this issues currently as it may come in handy for others :
```javascript
const ManifestPlugin = require('@symfony/webpack-encore/lib/webpack/webpack-manifest-plugin');
const ExtractTextPlugin = require('extract-text-webpack-plugin');

Encore
// (...) Your Encoreconfiguration

// export the final configuration

// JavaScript output file format
config.output.filename = '[name]-[chunkhash].js';

// Assets (fonts and images) output file format
for (let i = 0; i < config.module.rules.length; i++) {
    let rule = config.module.rules[i];
    if ('options' in rule && 'name' in rule.options) {
        rule.options.name = rule.options.name.replace('.[hash]', '-[hash]');
    }
}

// WebPack plugins configuration override
for (let i = 0; i < config.plugins.length; i++) {
    let plugin = config.plugins[i];

    // Custom manifest path
    if (plugin instanceof ManifestPlugin) {
        plugin.opts.fileName = "../../data/manifest.json";
    }

    // StyleSheets output file format
    if (plugin instanceof ExtractTextPlugin) {
        plugin.filename = '[name]-[contenthash].css';
    }
}
```
I think it would be really nice to be able to configure this as an Encore.setVersioningFileFormat() that would take an object with extension as key to configure the default formats :
```javascript
Encore
// (...) Your Encoreconfiguration
    .setVersioningFileFormat({
        js: '[name]-[chunkhash].js',
        css: '[name]-[contenthash].css',
        rules_suffix: '-[hash]',
    )
```
That object could be preset with current static values and implemented as : 

[@symfony/webpack-encore/lib/config-generator.js](https://github.com/symfony/webpack-encore/blob/master/lib/config-generator.js)
```javascript
// Currently
${this.webpackConfig.useVersioning ? '.[hash]' : ''}
// Could be 
${this.webpackConfig.useVersioning ? this.webpackConfig.versioningFormats.rules_suffix : ''}
...
// Currently
filename: this.webpackConfig.useVersioning ? '[name].[chunkhash].js' : '[name].js',
// Could be 
filename: this.webpackConfig.useVersioning ? this.webpackConfig.versioningFormats.js: '[name].js',
```
[@symfony/webpack-encore/lib/plugins/extract-text.js](https://github.com/symfony/webpack-encore/blob/master/lib/plugins/extract-text.js)
```javascript
// Currently
 filename: webpackConfig.useVersioning ? '[name].[contenthash].css' : '[name].css',
// Could be
 filename: webpackConfig.useVersioning ? webpackConfig.versioningFormats.css: : '[name].css',
```
Thanks for the work it was really nice to play with.
If there is a better place to submit this recommandation please let me now.

Lyrkan:
This PR adds the `Encore.configureFilenames()` method to the public API (fixes #125).

**Example:**

```js
Encore.configureFilenames({
    js: '[name].[chunkhash].js',
    css: '[name].[contenthash].css',
    images: 'images/[name].[hash:8].[ext]',
    fonts: 'fonts/[name].[hash:8].[ext]'
});
```

If one key is defined in the first parameter it will be used for the filenames of the related category.  
If it isn't the behavior will stay the same as before.

For now I kept the old naming strategies as the default values. We could eventually change that following the discussion in #136.

### DIFF
--- a/index.js
+++ b/index.js
@@ -472,6 +472,34 @@ const publicApi = {
     },
 
     /**
+     * Call this to change how the name of each output
+     * file is generated.
+     *
+     *     Encore.configureFilenames({
+     *         js: '[name].[chunkhash].js',
+     *         css: '[name].[contenthash].css',
+     *         images: 'images/[name].[hash:8].[ext]',
+     *         fonts: 'fonts/[name].[hash:8].[ext]'
+     *     });
+     *
+     * Only the filenames defined in the first parameter
+     * of this method will be modified.
+     *
+     * If you are using Encore.enableVersioning()
+     * make sure that your "js" filenames contain
+     * "[chunkhash]" and your "css" filenames contain
+     * "[contenthash]".
+     *
+     * @param {object} filenames
+     * @returns {exports}
+     */
+    configureFilenames(filenames) {
+        webpackConfig.configureFilenames(filenames);
+
+        return this;
+    },
+
+    /**
      * If enabled, the output directory is emptied between
      * each build (to remove old files).
      *
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -62,6 +62,7 @@ class WebpackConfig {
         this.forkedTypeScriptTypesCheckOptionsCallback = () => {};
         this.useImagesLoader = true;
         this.useFontsLoader = true;
+        this.configuredFilenames = {};
     }
 
     getContext() {
@@ -283,6 +284,22 @@ class WebpackConfig {
 
     disableFontsLoader() {
         this.useFontsLoader = false;
+    }
+
+    configureFilenames(configuredFilenames = {}) {
+        if (typeof configuredFilenames !== 'object') {
+            throw new Error('Argument 1 to configureFilenames() must be an object.');
+        }
+
+        // Check allowed keys
+        const validKeys = ['js', 'css', 'images', 'fonts'];
+        for (const key of Object.keys(configuredFilenames)) {
+            if (validKeys.indexOf(key) === -1) {
+                throw new Error(`"${key}" is not a valid key for configureFilenames(). Valid keys: ${validKeys.join(', ')}.`);
+            }
+        }
+
+        this.configuredFilenames = configuredFilenames;
     }
 
     cleanupOutputBeforeBuild() {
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -472,6 +472,34 @@ const publicApi = {
...
+     *
+     *     Encore.configureFilenames({
+     *         js: '[name].[chunkhash].js',
+     *         css: '[name].[contenthash].css',
+     *         images: 'images/[name].[hash:8].[ext]',
+     *         fonts: 'fonts/[name].[hash:8].[ext]'
+     *     });
+     *
+     * Only the filenames defined in the first parameter
+     * of this method will be modified.
```
weaverryan: How about:

> It's safe to omit a key (e.g. css): the default naming strategy
> will be used for those file types.

index.js
```diff
@@ -472,6 +472,34 @@ const publicApi = {
...
+     *
+     * If you are using Encore.enableVersioning()
+     * make sure that your "js" filenames contain
+     * "[chunkhash]" and your "css" filenames contain
+     * "[contenthash]".
```
weaverryan: nice

lib/WebpackConfig.js
```diff
@@ -285,6 +286,22 @@ class WebpackConfig {
...
+            throw new Error('Argument 1 to configureFilenames() must be an object.');
+        }
+
+        // Check allowed keys
+        const validKeys = ['js', 'css', 'images', 'fonts'];
+        for (const key of Object.keys(configuredFilenames)) {
+            if (validKeys.indexOf(key) === -1) {
+                throw new Error(`"${key}" is not a valid key for configureFilenames(). Valid keys: ${validKeys.join(', ')}.`);
+            }
+        }
```
weaverryan: well, I love this ;)

weaverryan:
Thank you @Lyrkan!

Seikyo:
Thank you for your time and the feature @Lyrkan @weaverryan

Issue #15, opened 2017-06-15
Document how javascript is being minified in production


trsteel88:
I can't find how javascript is being minified in production or what additional options we can give it (e.g. stripping special comments).

trajedy:
Looks like it is still a TODO. https://github.com/symfony/webpack-encore/blob/master/lib/config-generator.js#L390

weaverryan:
Indeed - we need a way to directly set the uglify options. What about:

```js
Encore
    .setUglifyJsOptions({...})
```

The options are here: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/README.md. I can't think of any reason why it would need to be any more complex than this :)

stof:
If we expose the options, we should decide how this interacts with the default options provided by Encore (and we should document them so that people can know what to expect)

weaverryan:
@stof In this case, I was thinking it would be an array merge, where (obviously) the user's config takes precedence on conflicts. But, I did something a bit different for the babel config. In that case, since the default babel config has a deeper structure (and maybe you want to just tweak on embedded key), it's configured with a callback, where we pass you the default config and you modify it. We could stay consistent and use that approach everywhere.

fabpot:
I would avoid merging configurations if this is not just key => value pairs. Not sure if that's the case here though :)

weaverryan:
The options (that we add) are a simple, scalar key-pair values now... but that may not be true forever :). 

Let's go with the callback way of doing it:

```js
Encore
    .configureUglifyJsPlugin(function(options) {
        // options already has our default values
    });
```

elmariachi111:
I tried to strip out license comments and minfify the results further in production by doing this (encore 0.12 / webpack 3.5):

```
Encore.configureBabel(function(babelConfig) { //this is actually documented.
    babelConfig.comments = false;
    return babelConfig;
})
```

but it **doesn't** strip the comments (tried a little bit but I cannot make it work :( ). 

So in a 2nd attempt I tried using the webpack.UglifyJS plugin:

```
if (Encore.isProduction()) {
    Encore.addPlugin(new webpack.optimize.UglifyJsPlugin({
        comments: false,
        compress: {
            warnings: false,
            drop_console: true,
        }
    }))
}
```

[...]

PR #141, opened 2017-08-17
Remove comments from javascript files in production environment

Seikyo:
This PR removes comments from compiled javascript files in production

Please let me know if this was intended to be the default behaviour I will work on implementing a method to allow default options to be overridden

### DIFF
--- a/lib/plugins/uglify.js
+++ b/lib/plugins/uglify.js
@@ -24,7 +24,10 @@ module.exports = function(plugins, webpackConfig, uglifyOptions = {}) {
     }
 
     let uglifyConfig = Object.assign({}, uglifyOptions, {
-        sourceMap: webpackConfig.useSourceMaps
+        sourceMap: webpackConfig.useSourceMaps,
+        output: {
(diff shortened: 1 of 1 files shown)


Lyrkan:
Hi @Seikyo,

Isn't that the default value for UglifyJS (cf. https://github.com/mishoo/UglifyJS2#output-options)?

There is another open issue (#15) kind of related to how comments are handled during minification. From what I read in it the only comments left should be the ones containing `@licence`  or `@preserve`, which is an intended behavior.

Seikyo:
Hi @Lyrkan 

After some research here is the [default value for uglifyJS comments](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/v0.4.6/src/index.js#L101) : `/^\**!|@preserve|@license/`

I think you can cancel this PR as it is indeed the expected behaviour. It would be a better idea to allow overriding uglifyJsPlugin as suggested in #15 and will resolve my issue.

For information, with the options `comments: false` all comments are properly removed.

PR #142, opened 2017-08-17
Feature configure manifest options

Seikyo:
This PR adds `configureManifestPlugin()` as discussed in #125 

**Example:**
```javascript
Encore.configureManifestPlugin({
    fileName: '../../var/assets/manifest.json'
})
```

These options will override any default option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,6 +103,25 @@ const publicApi = {
     },
 
     /**
+     * Allows to set ManifestPlugin options and override default options
+     * List of available options can be found at https://github.com/danethurber/webpack-manifest-plugin
+     *
+     * For example:
+     *
+     *      .configureManifestPlugin({
+     *          fileName: '../../var/assets/manifest.json'
+     *      })
+     *
+     * @param {Object} options
+     * @returns {publicApi}
+     */
+    configureManifestPlugin(options) {
+        webpackConfig.configureManifestPlugin(options);
+
+        return this;
+    },
+
+    /**
      * Adds a JavaScript file that should be webpacked:
      *
      *      // final output file will be main.js in the output directory
@@ -294,7 +313,8 @@ const publicApi = {
      * @param {function} postCssLoaderOptionsCallback
      * @return {exports}
      */
-    enablePostCssLoader(postCssLoaderOptionsCallback = () => {}) {
+    enablePostCssLoader(postCssLoaderOptionsCallback = () => {
+    }) {
         webpackConfig.enablePostCssLoader(postCssLoaderOptionsCallback);
 
         return this;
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -103,6 +103,25 @@ const publicApi = {
...
+     * List of available options can be found at https://github.com/danethurber/webpack-manifest-plugin
+     *
+     * For example:
+     *
+     *      .configureManifestPlugin({
+     *          fileName: '../../var/assets/manifest.json'
+     *      })
+     *
+     * @param {Object} options
+     * @returns {publicApi}
```
Lyrkan: `@returns {exports}`

lib/plugins/manifest.js
```diff
@@ -24,11 +24,14 @@ module.exports = function(plugins, webpackConfig) {
...
-    plugins.push(new ManifestPlugin({
+    let defaultConfig = {
         basePath: manifestPrefix,
         // guarantee the value uses the public path (or CDN public path)
         publicPath: webpackConfig.getRealPublicPath(),
         // always write a manifest.json file, even with webpack-dev-server
         writeToFileEmit: true,
-    }));
+    };
+    let manifestConfig = Object.assign(defaultConfig, webpackConfig.manifestPluginOptions);
```
Lyrkan: Just a minor remark here, `Object.assign` modifies its first argument, so in this case `defaultConfig` (which doesn't really makes sense since it won't be the default config anymore after that).

You probably could either do:

```js
const defaultConfig = { /* ... */ };
const manifestConfig = Object.assign({}, defaultConfig, webpackConfig.manifestPluginOptions);
```

Or only use a single variable (which looks a bit less clear in my opinion):

```js
const manifestConfig = Object.assign({}, { /* ... */ }, webpackConfig.manifestPluginOptions);

**Comment by Seikyo** [comment]
@Lyrkan I tried to looked into the other features pull request but honestly I am kind of lost with how this works. Could you tell me if this is more suitable ?

File: lib/WebpackConfig.js [hunk 3]
```diff
@@ -117,6 +118,16 @@ class WebpackConfig {
         this.manifestKeyPrefix = manifestKeyPrefix;
     }
 
+    configureManifestPlugin(manifestPluginOptionsCallback = () => {}) {
+        this.usePostCssLoader = true;
```
Lyrkan: You can probably remove that line :)

lib/WebpackConfig.js
```diff
@@ -117,6 +118,16 @@ class WebpackConfig {
...
+
+        if (typeof manifestPluginOptionsCallback !== 'function') {
+            throw new Error('Argument 1 to enablePostCssLoader() must be a callback function.');
```
Lyrkan: `Argument 1 to configureManifestPlugin() must be a callback function.`

Maybe you could also add a test for that part (in `test/WebpackConfig.js`).

Seikyo:
Should be good now, thank you for your help

weaverryan:
This is excellent work! Thank you @Seikyo!

PR #143, opened 2017-08-18
Intercept Webpack output during functional tests

### LINKED ISSUE #47: Functional tests should intercept the output
The output of the command running in functional tests should not be sent to the test runner output.

Lyrkan:
This PR replaces `process.stdout.write` by an empty function when running webpack in each functional test (fixes #47).

It also sets the `silent` option of the Zombie.js browser to true to avoid displaying things like:

```
Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
See more tips at https://vuejs.org/guide/deployment.html
```

The only "noisy" message left should be `DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead` since it is sent to stderr (should be fixed when alexindigo/webpack-chunk-hash#9 is released).

### DIFF
--- a/lib/test/setup.js
+++ b/lib/test/setup.js
@@ -60,32 +60,47 @@ function createWebpackConfig(testAppDir, outputDirName = '', command, argv = {})
 }
 
 function runWebpack(webpackConfig, callback, allowCompilationError = false) {
-    validator(webpackConfig);
-    const compiler = webpack(configGenerator(webpackConfig));
-    compiler.run((err, stats) => {
-        if (err) {
-            console.error(err.stack || err);
-            if (err.details) {
-                console.error(err.details);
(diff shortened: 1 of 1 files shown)


stof:
The fact that stderr messages are still displayed is fine to me. Usages of deprecated APIs should still be reported in our CI (even though it is not our fault this time)

weaverryan:
Awesome! Thank you @Lyrkan! Sorry for the delay - August is almost completely travel... with my infant. Will be back in a week for more complex stuff. I very much appreciate the attention for Encore :)

PR #144, opened 2017-08-21
Add Preact preset (with preact-compat support)

Lyrkan:
This PR adds a `enablePreactPreset()` method to the public API.

Basically it allows to use [Preact](https://github.com/developit/preact) (which is an alternative to react), optionally in conjunction with [`preact-compat`](https://github.com/developit/preact-compat) (which makes it possible to use the same code you would have in a React app).

It doesn't enable a babel preset per-se but it's still some kind of preset and it makes it consistent with the already existing `enableReactPreset()` method.

The method takes as an argument an option object that allows to enable/disable the use of `preact-compat`:

```js
// Without preact-compat (default)
// - Adds transform-react-jsx to the plugins used by babel
//   and configures it to wrap JSX into Preact h() calls
Encore.enablePreactPreset()

// With preact-compat
// - Adds transform-react-jsx to the plugins used by babel
//   but with its default configuration (that uses React.createElement)
// - Aliases react and react-dom to use preact-compat instead
Encore.enablePreactPreset({ preactCompat: true })
```

I also created the following project as a PoC and it seems to be working as intended:

* [without `preact-compat`](https://github.com/Lyrkan/encore-preact)
* [with `preact-compat`](https://github.com/Lyrkan/encore-preact/tree/preact-compat)

### DIFF
--- a/index.js
+++ b/index.js
@@ -394,16 +394,15 @@ const publicApi = {
      *     Encore.enablePreactPreset()
      *
      * If you wish to also use preact-compat (https://github.com/developit/preact-compat)
-     * call that method with its first parameter set
-     * to true:
+     * you can enable it by setting the "preactCompat" option to true:
      *
-     *     Encore.enablePreactPreset(true)
+     *     Encore.enablePreactPreset({ preactCompat: true })
      *
-     * @param {boolean} usePreactCompat
+     * @param {object} options
      * @returns {exports}
      */
-    enablePreactPreset(usePreactCompat = false) {
-        webpackConfig.enablePreactPreset(usePreactCompat);
+    enablePreactPreset(options = {}) {
+        webpackConfig.enablePreactPreset(options);
 
         return this;
     },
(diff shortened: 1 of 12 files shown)


index.js
```diff
@@ -388,6 +388,27 @@ const publicApi = {
...
+     * If you wish to also use preact-compat (https://github.com/developit/preact-compat)
+     * call that method with its first parameter set
+     * to true:
+     *
+     *     Encore.enablePreactPreset(true)
+     *
+     * @param {boolean} usePreactCompat
+     * @returns {exports}
+     */
+    enablePreactPreset(usePreactCompat = false) {
```
weaverryan: This is the only part I'm not sure about. When I saw your code examples, I assumed the true/false was turning the entire feature on/off (like some of our other `enable()` methods). It's more verbose, but I'd prefer `enablePreactPreset({preactCompat: true})` style (with our normal validating of any options passed).

Lyrkan: I definitely agree with you :)

weaverryan:
Great work! Another excellent addition! :)

PR #150, opened 2017-08-30
Add a 'Did you mean ...' error message in case of an unrecognized API property

Lyrkan:
This PR aims to improve the error message displayed when an user makes a typo in the name of one of the API methods he's trying to use.

Since pictures speak louder than words:

**Before**

**After**

Note that the error is triggered when trying to access the property, not when the method is called (since it doesn't exist), hence the "is not a recognized property" message. I'm aware that the use of the "property" word in it may not be ideal, but using "method" instead feels wrong too since we could theoretically have other things than methods in the API object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -645,7 +645,7 @@ const publicApiProxy = new Proxy(publicApi, {
                 }
             }
 
-            const error = new Error(`${chalk.red(`Encore.${prop}`)} is not a recognized property, did you mean ${chalk.green(`Encore.${similarProperty}`)}?`);
+            const error = new Error(`${chalk.red(`Encore.${prop}`)} is not a recognized property or method, did you mean ${chalk.green(`Encore.${similarProperty}`)}?`);
             console.log(new PrettyError().render(error));
             process.exit(1); // eslint-disable-line
         }
--- a/index.js
+++ b/index.js
@@ -633,7 +633,9 @@ const publicApiProxy = new Proxy(publicApi, {
                     process.exit(1); // eslint-disable-line
                 }
             };
-        } else if (typeof target[prop] === 'undefined') {
+        }
+
+        if (typeof target[prop] === 'undefined') {
             // Find the property with the closest Levenshtein distance
             let similarProperty;
             let minDistance = Number.MAX_VALUE;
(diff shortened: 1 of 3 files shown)


stof:
@Lyrkan what about saying `unrecognized property or method` ?

Lyrkan:
I was so focused on trying to find a single word that I didn't even think of using both of them :)

index.js
```diff
@@ -631,6 +633,21 @@ const publicApiProxy = new Proxy(publicApi, {
                     process.exit(1); // eslint-disable-line
                 }
             };
+        } else if (typeof target[prop] === 'undefined') {
```
weaverryan: nit pick: since we return in the first `if`, this doesn't need to be a nested `else if`, it can just be a fresh, un-nested `if`.

index.js
```diff
@@ -631,6 +633,21 @@ const publicApiProxy = new Proxy(publicApi, {
...
+            // Find the property with the closest Levenshtein distance
+            let similarProperty;
+            let minDistance = Number.MAX_VALUE;
+            for (const apiProperty in target) {
+                const distance = levenshtein.get(apiProperty, prop);
+                if (distance <= minDistance) {
+                    similarProperty = apiProperty;
+                    minDistance = distance;
+                }
+            }
```
weaverryan: If you type something totally crazy that is completely not close to anything, it will suggest *something*, right? In Symfony's core, we use "3" as the maximum distance (if it's more than 3, we don't recommend anything): https://github.com/symfony/symfony/blob/1bb2bc322bc1c2d0e6e70c2a0fa9fc0be90757fd/src/Symfony/Component/ExpressionLanguage/SyntaxError.php#L34.

It probably makes sense to do the same thing here

weaverryan:
<3 this feature. Thank you @Lyrkan!

index.js
stof: to be exact, we don't use 3 most of the time. We use `strlen(input) / 3` (we have a few cases using 3)

weaverryan: Ah, thanks! I've just changed that at sha: 2e216a465fdd42071e4d781b5b84ad47e5442267... and indeed, it is a bit more "forgiving" in appropriate situations.

PR #152, opened 2017-09-04
Add various methods to configure default plugins

### LINKED ISSUE #15: Document how javascript is being minified in production

### LINKED ISSUE #87: Make clean-webpack-plugin configurable

### LINKED ISSUE #148: Expose mangle option for UglifyJsPlugin
Greetings!

Came across the problem of AgularJs minification. Some third party modules are not minified correctly with mangle option enabled. It throws error like ``Error: Unknown provider: tProvider <- t``. Unfortunately, there is no way to disable mangle via Encore interface. Please, add the option. Thank you in advance!

P.S. For those, experiencing the same problem, here's how I solved it:

```js
...

let config = Encore.getWebpackConfig(),
    webpack = require('webpack');

if (Encore.isProduction()) {
    let index;

    config.plugins.forEach((plugin, i) => {
        if (plugin instanceof webpack.optimize.UglifyJsPlugin) {
            index = i;
        }
    });

    if (index) {
        config.plugins.splice(index, 1);
    }

    config.plugins.push(new webpack.optimize.UglifyJsPlugin({
        mangle: false,
        sourceMap: !Encore.isProduction()
    }));
}
```

Lyrkan:
This PR adds some methods to configure default plugins (closes #148, closes #87 and closes #15):

* `Encore.configureDefinePlugin(callback)`
* `Encore.configureExtractTextPlugin(callback)`
* `Encore.configureFriendlyErrorsPlugin(callback)`
* `Encore.configureLoaderOptionsPlugin(callback)`
* `Encore.configureUglifyJsPlugin(callback)`

Other changes:

* Allows the `clean-webpack-plugin` to be configured using `Encore.cleanupOutputBeforeBuild(paths, callback)`
* Fixed a small mistake in the `Encore.configureManifestPlugin()` jsdoc
* Reorganized flags/callbacks in the constructor of `webpack.config.js` since it was starting to be a bit hard to read. I'm not sure about the way I splitted things though, let me know what you think about it.
* Renamed `common-chunks.js` to `commons-chunks.js` since the name of the plugin is `CommonsChunkPlugin`

@weaverryan: Not directly related but while doing all of that I noticed that `sassOptions` uses snake-case whereas I used camel-case in #144 for `enablePreactPreset()` and in #115 for `configureRuntimeEnvironment()`, should we do something about that?

***Edit 1:** Added `Encore.configureCleanWebpackPlugin()`*
***Edit 2:** Removed `Encore.configureCleanWebpackPlugin()` to use `Encore.cleanupOutputBeforeBuild(paths, callback)` arguments instead*


weaverryan:
> Not directly related but while doing all of that I noticed that sassOptions uses snake-case whereas I used camel-case in #144 for enablePreactPreset() and in #115 for configureRuntimeEnvironment(), should we do something about that?

Hmm, yea, good catch. We should probably "get this right" now. I'm leaning towards using camelCase... so we would "fix" `sassOptions` (I'm leaning towards this in part because the options in #115 are camelCase just because that's how the CLI options are normalized internally.

index.js
```diff
@@ -103,13 +103,89 @@ const publicApi = {
...
+     * @param {function} loaderOptionsPluginOptionsCallback
+     * @returns {exports}
+     */
+    configureLoaderOptionsPlugin(loaderOptionsPluginOptionsCallback = () => {}) {
+        webpackConfig.configureLoaderOptionsPlugin(loaderOptionsPluginOptionsCallback);
+
+        return this;
+    },
+
+    /**
```
weaverryan: This all looks beautiful

index.js
```diff
@@ -522,10 +618,20 @@ const publicApi = {
      * If enabled, the output directory is emptied between
      * each build (to remove old files).
```
weaverryan: It's not from your commit, but this could probably fit all on one line.

index.js
```diff
@@ -522,10 +618,20 @@ const publicApi = {
...
      *
+     * A list of available options can be found at https://github.com/johnagan/clean-webpack-plugin
+     *
+     * For example:
+     *
+     *      Encore.cleanupOutputBeforeBuild(['*.js'], (options) => {
```
weaverryan: Maybe we should show the default `['**/*']` as the first option. The issue is that, if I *only* want to specify some options for the plugin... I need to know what this first value is by default, so I can repeat it. That kind of sucks... but I can't think of a good way around it (so showing the default here might make sense)

lib/WebpackConfig.js
```diff
@@ -311,8 +371,18 @@ class WebpackConfig {
         this.configuredFilenames = configuredFilenames;
     }
 
-    cleanupOutputBeforeBuild() {
+    cleanupOutputBeforeBuild(paths = ['**/*'], cleanWebpackPluginOptionsCallback = () => {}) {
+        if (!Array.isArray(paths)) {
+            throw new Error('Argument 1 to cleanupOutputBeforeBuild() must be an Array');
```
weaverryan: What do you think about this:

> ... must be an Array of paths - e.g. ['**/*']

index.js
Lyrkan: I wanted to but `'**/*'` breaks the multi-line comment :(

weaverryan: Ha! What a crazy edge case :p

weaverryan:
Thank you @Lyrkan!

PR #159, opened 2017-09-13
Replace "resolve_url_loader" by "resolveUrlLoader"

Lyrkan:
This PR deprecates the `resolve_url_loader` option used by the `enableSassLoader` and replaces it by `resolveUrlLoader`.

This is done to be consistent with other methods that use camel-case instead of snake-case (see [#152](https://github.com/symfony/webpack-encore/pull/152#issuecomment-329266198)).

`resolve_url_loader` will still work for now but will display the following deprecation message:

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -57,16 +57,16 @@ class WebpackConfig {
         this.useSassLoader = false;
         this.useReact = false;
         this.usePreact = false;
-        this.preactOptions = {
-            preactCompat: false
-        };
         this.useVueLoader = false;
         this.useTypeScriptLoader = false;
         this.useForkedTypeScriptTypeChecking = false;
 
         // Features/Loaders options
         this.sassOptions = {
-            resolve_url_loader: true
+            resolveUrlLoader: true
+        };
+        this.preactOptions = {
+            preactCompat: false
         };
 
         // Features/Loaders options callbacks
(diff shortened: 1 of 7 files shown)


lib/WebpackConfig.js
```diff
@@ -251,11 +251,17 @@ class WebpackConfig {
         this.sassLoaderOptionsCallback = sassLoaderOptionsCallback;
 
         for (const optionKey of Object.keys(options)) {
-            if (!(optionKey in this.sassOptions)) {
-                throw new Error(`Invalid option "${optionKey}" passed to enableSassLoader(). Valid keys are ${Object.keys(this.sassOptions).join(', ')}`);
+            let normalizedOptionKey = optionKey;
+            if (optionKey === 'resolve_url_loader') {
+                logger.warning('enableSassLoader: "resolve_url_loader" is deprecated. Please use "resolveUrlLoader" instead.');
```
weaverryan: How about a new `logger.deprecation` method? It will probably look the same, but can be a nice, standard way to start doing this

lib/WebpackConfig.js
```diff
@@ -57,16 +57,16 @@ class WebpackConfig {
...
         this.useVueLoader = false;
         this.useTypeScriptLoader = false;
         this.useForkedTypeScriptTypeChecking = false;
 
         // Features/Loaders options
         this.sassOptions = {
-            resolve_url_loader: true
+            resolveUrlLoader: true
+        };
+        this.preactOptions = {
```
Lyrkan: Not related to this PR but it wasn't at the right place since #152 was merged after #144

weaverryan:
Awesome. Thank you @Lyrkan!

PR #164, opened 2017-09-21
Trying out the new webpack-manifest-plugin

### LINKED ISSUE #140: How to manually add entries to manifest.json?
Hi there,

I've got somewhat legacy (huge dumped JS arrays) code. Without Encore I can do this:

```js
new ConcatPlugin({
    uglify: true,
    useHash: true,
    sourceMap: false,
    fileName: 'shipdev-data.js',
    filesToConcat: ['./frontend/shipdev/data/modules.js', './frontend/shipdev/data/ships.js']
})
```

I would like to keep it this way for now, because it works. I tried using `addEntry` but this runs for ages (I cancelled after 10minutes). Now using Encore, I am adding this:

```js
addPlugin(new ConcatPlugin({
    uglify: true,
    useHash: true,
    sourceMap: false,
    fileName: 'shipdev-data.js',
    filesToConcat: ['./frontend/shipdev/data/modules.js', './frontend/shipdev/data/ships.js']
}))
```

Which works fine to the point that the file is available and contains the content as it should. Now I would like to have this entry being added to the `manifest.json`. Is there an elegant way with Encore?

My current workaround is just defining `useHash: false`, which works for `asset('shipdev-data.js')` with `framework.assets.base_path` configured.

weaverryan:
This is not ready to be merged yet... as we should wait for a stable release. But, let's see if the tests pass :). We originally "forked" this library internally due to 2 issues, both of which should have been fixed in version 2 of the library.

Should fix #140.

TODOS:

- Remove the old `webpack-manifest-plugin.js` file entirely and update paths to require the actual library instead of this

- Stop passing the `publicPath` option to the plugin, as this is not supported anymore (I believe it's just being ignored right now)


weaverryan:
Failures are due to: https://github.com/salesforce/tough-cookie/issues/92... which is unrelated to us, and will hopefully get resolved shortly.

stof:
yeah, too bad that JS regexp does not have possessive quantifiers. In PHP, they would allow fixing the issue very easily (we did such fixes in Symfony).

Lyrkan:
> Remove the old webpack-manifest-plugin.js file entirely and update paths to require the actual library instead of this

I'd keep the file for now with only the export of the real plugin and a `logger.deprecation(...)` in it in case someone currently uses it to do something like:

```js
const ManifestPlugin = require('@symfony/webpack-encore/lib/webpack/webpack-manifest-plugin.js');

// (...)

const config = Encore.getWebpackConfig();

for (const plugin of config.plugins) {
    if (plugin instanceof ManifestPlugin) {
        // Do something with its config
    }
}

module.exports = config;
```

weaverryan:
Hmm, interesting point - I'm cool with that :). In general, however, we should discourage people from requiring specific files from the lib - these shouldn't be considered part of our public API, unless we advertise using them (like the new PluginPriorities module).

Lyrkan:
@weaverryan  It looks like Webpack Manifest Plugin 2.0.0 is out :)

Issue #163, opened 2017-09-21
Control over plugin order


stof:
Currently, custom plugins are registered last. but this can cause issues:

- when using the copy-webpack-plugin (assuming plugins are releasing versions containing the support for manifest and copy together), you need to register it before the manifest plugin AFAIK (otherwise the manifest won't see the new entries)
- when using some plugins altering the JS code, you are likely to want them registered before the UglifyJS plugin, so that the final file is minified.

Other examples may come.

I don't know whether we should build a way to control the order precisely, or whether we should just change the position of custom plugins.

Lyrkan:
Hey @stof,

What do you think about something like this?

```js
const webpack = require('webpack');
const ManifestPlugin = require('webpack-manifest-plugin'); // Not the one currently used
const Plugin1 = require('plugin1');
const Plugin2 = require('plugin2');
const Plugin3 = require('plugin3');

Encore
  .addPlugin(new Plugin1())
  .addPlugin(new Plugin2())
  .addPlugin(new Plugin3())
  .sortPlugins([
    Plugin1,
    webpack.optimize.UglifyJsPlugin,
    Plugin2,
    Plugin3,
    ManifestPlugin
  ]);
```

Then just before [this return](https://github.com/symfony/webpack-encore/blob/master/lib/config-generator.js#L243) we would reorder `webpackConfig.plugins` so:

* all instances of Plugin1 are added before instances of UglifyJsPlugin
* all instances of UglifyJsPlugin are added before instances of Plugin2
* all instances of Plugin2 are added before instances of Plugin3
* etc.

This should work if we don't have multiple instances of the same plugin that need to be sorted relatively to each other (which can already be done for custom plugins by calling `addPlugin` in the right order).

I would also move the manifest-plugin after the custom plugins by default in order to avoid the issue with the copy-webpack-plugin.

PR #165, opened 2017-09-22
Add Encore.sortPlugins() method

Lyrkan:
This PR tries to solve #163 (ping @stof) by adding a `sortPlugins` method to the public API.

Usage:
```js
const webpack = require('webpack');
const Plugin1 = require('plugin1');
const Plugin2 = require('plugin2');

Encore.sortPlugins([
    Plugin2,
    webpack.optimize.UglifyJsPlugin,
    Plugin1
])
```

Since it relies on `instanceof` checks it won't allow to sort two instances of a given plugin relatively to eachother but I'm not sure that case needs to be handled anyway...

I also changed the default position of the `ManifestPlugin` so it is added after custom plugins. This is done in order to prevent issues with other plugins that emit assets (e.g. `CopyWebpackPlugin` and `ConcatPlugin` after #164 is merged).

### DIFF
--- a/index.js
+++ b/index.js
@@ -279,6 +279,37 @@ const publicApi = {
     },
 
     /**
+     * Sorts plugins so they are added to Webpack in the given order.
+     *
+     * This method takes an array of the constructor functions used to
+     * initialize the plugins you want to sort.
+     *
+     * You only need to provide the plugins that need to be sorted:
+     * the other ones will keep their default position or respect
+     * the order in which the Encore.addPlugin() method has been called.
+     *
+     * For example:
+     *
+     *      const webpack = require('webpack');
+     *      const Plugin1 = require('plugin1');
+     *      const Plugin2 = require('plugin2');
+     *
+     *      Encore.sortPlugins([
+     *          Plugin2,
+     *          webpack.optimize.UglifyJsPlugin,
+     *          Plugin1
+     *      ])
+     *
+     * @param {Array} sortOrder
+     * @return {exports}
+     */
+    sortPlugins(sortOrder) {
+        webpackConfig.sortPlugins(sortOrder);
+
+        return this;
+    },
+
+    /**
      * Adds a custom loader config
      *
      * @param {object} loader The loader config object
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -45,6 +45,7 @@ class WebpackConfig {
         this.sharedCommonsEntryName = null;
         this.providedVariables = {};
         this.configuredFilenames = {};
+        this.sortPluginsOrder = [];
 
         // Features/Loaders flags
         this.useVersioning = false;
@@ -254,6 +255,14 @@ class WebpackConfig {
 
     addPlugin(plugin) {
         this.plugins.push(plugin);
+    }
+
+    sortPlugins(sortOrder = []) {
+        if (!Array.isArray(sortOrder)) {
+            throw new Error('Argument 1 to sortPlugins() must be an Array of constructor functions (e.g.: [ExtractTextPlugin, CleanWebpackPlugin])');
+        }
+
+        this.sortPluginsOrder = sortOrder;
     }
 
     addLoader(loader) {
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -250,19 +250,17 @@ class ConfigGenerator {
             const currentPlugin = plugins.shift();
             const sortedIndex = sortOrder.findIndex((type) => currentPlugin instanceof type);
 
-            if (sortedIndex >= 0) {
-                // If the current plugin type has been found in the sortOrder array
-                // we move every instances of the types that are supposed to be placed
-                // before it.
-                for (let i = 0; i < sortedIndex; i++) {
-                    const currentType = sortOrder.shift();
+            // If the current plugin type has been found in the sortOrder array
+            // we move every instances of the types that are supposed to be placed
+            // before it.
+            for (let i = 0; i < sortedIndex; i++) {
+                const currentType = sortOrder.shift();
 
-                    // Add plugins of the current type to the sorted array
-                    sortedPlugins.push(...plugins.filter((plugin) => plugin instanceof currentType));
+                // Add plugins of the current type to the sorted array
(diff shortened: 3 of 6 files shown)


weaverryan:
Yo!

I’m not against this solution... but I want to exhaust all other possibilities, just be sure this solution is inherently a bit complex.

First, is the manifest plugin & copy plugin ordering issue confirmed? Do we know for sure that order matters? Second, do we know any other issues? Stop mentioned one about Uglify, can we find a real example?

Also, what about adding a prependPlugin() method. Along with the existing method, you could then add plugins before the core plugins or after... which doesn’t give you complete flexibility, but much more.

What do you think?

Lyrkan:
Hi Ryan,

I have no idea if that's the right solution either... see this more as a POC of what I described in the comments of the issue :)

I tried to find a middle ground between something that wouldn't be flexible enough and something that would do the same than `config.plugins = config.plugins.sort((plugin) => { ... })`) (which would be a real PITA to configure).

> First, is the manifest plugin & copy plugin ordering issue confirmed? Do we know for sure that order matters?

I think it is, I've also seen it mentionned in [this thread](https://github.com/danethurber/webpack-manifest-plugin/pull/75#issuecomment-325623094) (ping @davidmpaz).

From my undestanding the order of plugins does matter. Let's say you add two plugins that both have a callback on the same event:

```js
function Plugin1() {}
Plugin1.prototype.apply = function(compiler) {
  compiler.plugin("emit", function(compilation, callback) {
    // ...
  }
}

function Plugin2() {}
Plugin2.prototype.apply = function(compiler) {
  compiler.plugin("emit", function(compilation, callback) {
    // ...
  }
}
```

If you add `[new Plugin1(), new Plugin2()]` and run webpack, the callback from Plugin1 will be called before the one from Plugin2. So if Plugin1 does something that is needed by Plugin2 during that step it won't work properly.

I don't know if that's the case for UglifyJsPlugin since it highly depends on how the plugin is implemented but I definitely think that's something that users should be able to configure.

> Also, what about adding a prependPlugin() method. Along with the existing method, you could then add plugins before the core plugins or after... which doesn’t give you complete flexibility, but much more.

I thought about that too, but I'm not sure that it would be flexible enough... what if a plugin needs to be added between two of the plugins added by Encore?

weaverryan:
> If you add [new Plugin1(), new Plugin2()] and run webpack, the callback from Plugin1 will be called before the one from Plugin2. So if Plugin1 does something that is needed by Plugin2 during that step it won't work properly.

Yes, this is certainly true. The question is: how often will this be a problem? Since it *will* be a problem at least *sometimes* (even if it's an edge-case), it makes sense to add a solution for it (and most people won't need this solution anyways, so it's not something they will even need to know about).

> but I'm not sure that (`prependPlugin`) would be flexible enough... what if a plugin needs to be added between two of the plugins added by Encore?

Very true again :)

I have 2 counter-proposals:

A) Add a `copyFiles()` method to help alleviate this problem

and 

B) Internally, give each plugin a "priority" number - e.g. 10, 20, -10, 20. Then, add a 2nd argument to `addPlugin()` which is an optional priority number. This would allow you to "place" a plugin between 2 other plugins, if you care to.

Wdyt? This PR is perfect... but it's trying to work around an inflexibility in Encore itself... so it's not super user friendly. Maybe (B) is a bit nicer?

Lyrkan:
> A) Add a copyFiles() method to help alleviate this problem

We should probably add it anyway but that'll not be enough if other plugins have the same kind of issue (e.g. ConcatPlugin).

> B) Internally, give each plugin a "priority" number - e.g. 10, 20, -10, 20. Then, add a 2nd argument to addPlugin() which is an optional priority number.

I thought about that before starting to work on this PR. My main concern was that it wouldn't be easy enough to use/understand either:

```js
// We would have to add constants so we can still change priorities
// if needed later on without introducing breaking changes... but
// that'll only work if the users use it.
const PluginPriorities = require('@symfony/webpack-encore/lib/plugin-priorities');

Encore
  .addPlugin(new Plugin1())

  // Since we'll have the same priority than Plugin1 here (default),
  // are we sure that it won't end-up being added before it when sorting
  // all the plugins? AFAIK JavaScript sort is not guaranteed to be stable.
  .addPlugin(new Plugin2())
  
  // Unless we provide with constants like "PluginPriorities.BeforeManifestPlugin"
  // or "PluginPriorities.AfterUglifyJsPlugin" the user will have to do it manually
  .addPlugin(new Plugin3(), PluginPriorities.ManifestPlugin - 1)
  .addPlugin(new Plugin4(), PluginPriorities.UglifyJsPlugin + 1)
;
```

weaverryan:
A) Let's plan to add the `copyFiles()` method then. See #175

About the priority stuff, I'm happy you were also thinking about it :)

> We would have to add constants so we can still change priorities
> if needed later on without introducing breaking changes... but
> that'll only work if the users use them.

It would still be a BC break to change the orders, even if we use the constants. This is exactly how Symfony does it with event listeners. It's a bit odd to choose random integers for priority. But, in practice, it works fine. I would expect to start with everything as a 0 priority, except (unless I'm missing something) the plugin in the future `copyFiles()` - it would maybe have a priority of -100.

> Since we'll have the same priority than Plugin1 here (default),
> are we sure that it won't end-up being added before it when sorting
> all the plugins? AFAIK JavaScript sort is not guaranteed to be stable.

This is an implementation detail... but yea... we would need to make sure the sort is maintained when the priorities are equal. We just need to figure that out :)

I was picturing something like this:

```js
Encore
  .addPlugin(new Plugin1())
  .addPlugin(new Plugin2())
  
  .addPlugin(new Plugin3(), -50)
  .addPlugin(new Plugin4(), 50)
;
```

Internally, if we had a file where all the plugin priorities were set to constants (even if most were 0), this would be a cool way of documenting them. In the future, if we realize that people need to be able to add plugins *between* two core plugins that have a 0 priority, we'll need to break BC (in a minor way) by giving those existing plugins different priority. We're not even sure if this is going to happen, so that's why I like this simple solution.

Lyrkan:
If you don't mind the BC breaks I'm ok with that solution too :+1:

Closing this PR!

PR #172, opened 2017-10-04
Add an instance of DefinePlugin even in development mode

### LINKED ISSUE #171: Why DefinePlugin is configured only in production?
https://github.com/symfony/webpack-encore/blob/929a253f26b5b5875a1486f9c2d4552933ac947c/lib/plugins/define.js#L21

Lyrkan:
This PR fixes #171 by adding an instance of the `DefinePlugin` to the generated Webpack config no matter what the current environment is.


weaverryan:
Ah yes, definitely a good fix! Thanks!

PR #177, opened 2017-10-08
Add a priority parameter to the Encore.addPlugin() method

### LINKED ISSUE #163: Control over plugin order

Lyrkan:
Following the discussion in #165 this PR adds a `priority` parameter to the `Encore.addPlugin()` method in order to allow people to choose where their custom plugins will be added in the generated Webpack config (fixes #163).

__Examples:__

```js

// Without setting the priority a plugin will be
// appended to the other plugins.
Encore.addPlugin(new CustomPlugin());
```

```js

// If two plugins need to be sorted relatively to
// eachother (CustomPluginB will be placed before
// CustomPluginA):
Encore
    .addPlugin(new CustomPluginA(), 100);
    .addPlugin(new CustomPluginB(), 50)
;
```

```js
const PluginPriorities = require('@symfony/webpack-encore/lib/plugins/plugin-priorities.js');

// For now all plugins added by Encore have a
// priority of 0 (the same as the default priority for
// custom plugins), but constants are available in
// order to avoid BC breaks:
Encore.addPlugin(new CustomPlugin(), PluginPriorities.UglifyJsPlugin);
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -269,11 +269,35 @@ const publicApi = {
      * For example, if you want to add the "webpack.IgnorePlugin()", then:
      *      .addPlugin(new webpack.IgnorePlugin(requestRegExp, contextRegExp))
      *
+     * By default custom plugins are added after the ones managed by Encore
+     * but you can also set a priority to define where your plugin will be
+     * added in the generated Webpack config.
+     *
+     * For example, if a plugin has a priority of 0 and you want to add
+     * another plugin after it, then:
+     *
+     *      .addPlugin(new MyWebpackPlugin(), -10)
+     *
+     * The priority of each plugin added by Encore can be found in the
+     * "lib/plugins/plugin-priorities.js" file. It is recommended to use
+     * these constants if you want to add a plugin using the same priority
+     * as one managed by Encore in order to avoid backward compatibility
+     * breaks.
+     *
+     * For example, if you want one of your plugins to have the same priority
+     * than the DefinePlugin:
+     *
+     *      const Encore = require('@symfony/webpack-encore');
+     *      const PluginPriorities = require('@symfony/webpack-encore/lib/plugins/plugin-priorities.js');
+     *
+     *      Encore.addPlugin(new MyWebpackPlugin(), PluginPriorities.DefinePlugin);
+     *
      * @param {string} plugin
+     * @param {number} priority
      * @return {exports}
      */
-    addPlugin(plugin) {
-        webpackConfig.addPlugin(plugin);
+    addPlugin(plugin, priority = 0) {
+        webpackConfig.addPlugin(plugin, priority);
 
         return this;
     },
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -252,8 +252,15 @@ class WebpackConfig {
         this.styleEntries.set(name, src);
     }
 
-    addPlugin(plugin) {
-        this.plugins.push(plugin);
+    addPlugin(plugin, priority = 0) {
+        if (typeof priority !== 'number') {
+            throw new Error('Argument 2 to addPlugin() must be a number.');
+        }
+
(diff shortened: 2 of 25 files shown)


index.js
```diff
@@ -269,11 +269,35 @@ const publicApi = {
...
      *      .addPlugin(new webpack.IgnorePlugin(requestRegExp, contextRegExp))
      *
+     * By default custom plugins are added after the ones managed by Encore
+     * but you can also set a priority to define where your plugin will be
+     * added in the generated Webpack config.
+     *
+     * For example, if a plugin has a priority of 0 and you want to add
+     * another plugin before it, then:
+     *
+     *      .addPlugin(new MyWebpackPlugin(), -10)
```
weaverryan: We need to reverse the sorting: priority means that a *higher* number wins. So `-10` should run *after* plugins with priority 0. This is also consistent with Symfony's priorities.

index.js
```diff
@@ -269,11 +269,35 @@ const publicApi = {
...
+     * breaks.
+     *
+     * For example, if you want one of your plugins to have the same priority
+     * than the DefinePlugin:
+     *
+     *      const Encore = require('@symfony/webpack-encore');
+     *      const PluginPriorities = require('@symfony/webpack-encore/lib/plugins/plugin-priorities.js');
+     *
+     *      Encore.addPlugin(new MyWebpackPlugin(), PluginPriorities.DefinePlugin);
+     *
```
weaverryan: Great docs, btw. And nice idea to introduce the optional PluginPriorities.

index.js
Lyrkan: Whoops, fixed :)

weaverryan:
Awesome, thank you @Lyrkan!

Lyrkan:
@stof The reason for that choice is explained in https://github.com/symfony/webpack-encore/pull/165#issuecomment-334862597: only the ManifestPlugin seems to require being placed after (nearly) everything else.

I'm not against setting a different priority for each plugin (I even thought about doing so initially) but is it really necessary?

stof:
but then, at least the manifest plugin  should have a different priority

Lyrkan:
Adding a plugin before the manifest plugin can already be done by doing:

```js
Encore.addPlugin(
    new MyPlugin(),
    PluginPriorities.WebpackManifestPlugin + 1
);
```

If the order doesn't matter for the current other built-in plugins (I may be wrong on that though) it shouldn't be an issue to add the new plugin before all of them.

As en example #221 is going to use the CopyWebpackPlugin which needs to be added before the manifest plugin. In this case it makes sense to use a different priority.

Lyrkan:
Just realized I said that the manifest plugin needed to be added after everything else in my first comment... I blame tiredness because that's definitely not what I meant.

It must only be placed after some plugins that emit additional assets, which is the case for the `copy-webpack-plugin` and the `webpack-concat-plugin`.

Lyrkan:
Currently yes since the order doesn't seem to matter for the core plugins we provide right now.

The idea behind that PR is to have something flexible enough in case other plugins (not necessarily core ones) require to be added before/after each others. And if later on we find out that an external plugin needs to be placed between two core plugins we could still modify those two plugins' priority.

#221 will also introduce the `copy-webpack-plugin` which will not have a priority of 0 since it has to be added before the manifest plugin (but yeah, that's still just an "add this one before anything else" situation).

Issue #85, opened 2017-07-08
How should image assets be handled?


MPLLC:
I have a couple image files that are referenced by CSS.  How should I handle these with Webpack Encore, if at all?  The documentation doesn't say anything about image assets.  Can I use `addEntry`?  Or will they be automatically detected/handled if they're in a location like `/path/to/project/app/Resources/assets/images`?

It's just a bit confusing coming from using Assetic in all my old projects.

davidmpaz:
Maybe this is a duplicate of #24 ? Could you confirm @MPLLC? Work is done in that direction, not yet ready though.

MPLLC:
Ah, didn't see #24.  Yes, my question/issue is more or less the same.

weaverryan:
Closing as a duplicate - thanks for your question! It's one of our top issues to clarify for people :)

PR #179, opened 2017-10-11
Fix issue incorrect paths for fonts. Add methods to Use relative paths

nikrou:
The idea is to allow to use relative path for fonts and/or images.


css/style_with_fonts.scss
```diff
@@ -0,0 +1,9 @@
+$font-path: '../fonts';
```
Lyrkan: This file should probably be in the `fixture/css` folder.

index.js
```diff
@@ -75,6 +75,18 @@ const publicApi = {
         return this;
     },
 
+    setFontsPublicPath(publicPath) {
```
Lyrkan: I'd add at least a small comment here with some use cases.

index.js
```diff
@@ -75,6 +75,18 @@ const publicApi = {
...
+        webpackConfig.setFontsPublicPath(publicPath);
+
+        return this;
+    },
+
+    setImagesPublicPath(publicPath) {
```
Lyrkan: Missing comment for that one too.

Lyrkan: By the way, do we really need two methods? is there a use case in which we would only call one of them or call them with different values?

Lyrkan (commented):
Some comments I have after a quick look at your changes (one of them probably being related to the failing tests).

I would also add a test case for each method in `test\index.js` to be sure that the `return this` doesn't silently disappear later on.

As I said in #85 though, I'm not sure that's the right way to handle it. The `file-loader` seems to provide a `useRelativePath` option that could do the job (didn't have the time to test it, so I may be wrong).

weaverryan:
Yea, we need to do some playing with this. I'm also interested in `useRelativePath`, but it may have some issues: https://github.com/webpack-contrib/file-loader/pull/150... though I think just using the relative paths for `publicPath` also has issues... @Lyrkan we need to do some experimenting with this. I think it's unlikely that we'll be able to change Encore to have `useRelativePath` by default (again, because I think there are issues), but we might make a way to opt into this. I think we need experimenting to know more, then we can find the best solution.

scaytrase:
@weaverryan any news on relative paths?

LeoShivas:
Hello,

I also need relative paths for fonts. Any news ?

Kocal:
Closing this one. Webpack Encore is now in low-maintenance mode and only accepts bug fixes, and relative font paths is a new feature, so it's out of scope. It's also from 2017 and the patch no longer applies to the current codebase. Thanks for the contribution back then, @nikrou.

PR #182, opened 2017-10-13
Upgrade webpack-chunk-hash dependency

### LINKED ISSUE #91: Webpack 3.1 deprecation warning
#53 moved us to Webpack 3.1 without any issues.

But, we do have 1 deprecation we should fix before releasing:

> (node:32501) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.

This comes from the `addStyleEntry .js files are removed` functional test.

Lyrkan:
Fixes #91 by using `webpack-chunk-hash@^0.5.0` instead of `webpack-chunk-hash@^0.4.0` ([caret ranges](https://yarnpkg.com/lang/en/docs/dependency-versions/#toc-caret-ranges) don't allow to change the first non-zero digit).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "resolve-url-loader": "^2.0.2",
     "style-loader": "^0.13.2",
     "webpack": ">=2.2.0 <4",
-    "webpack-chunk-hash": "^0.4.0",
+    "webpack-chunk-hash": "^0.5.0",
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks @Lyrkan :)

PR #190, opened 2017-10-29
Add Encore.enableBuildNotifications() method

### LINKED ISSUE #187: Add webpack-notifier
When your build finishes (or fails), there's not currently any notification, other than on the terminal itself. So, it's easy to not realize that you have a build problem. 

This fix is easy - use `webpack-notifier`. I only originally didn't add this, because I don't love having notifications on build success, but I've talked to others that would like this (and I think having it on errored builds definitely makes sense). Perhaps something like:

```js
Encore.enableBuildNotifications();
```

Lyrkan:
This PR adds an `enableBuildNotifications()` method to the public API in order to use the `webpack-notifier` plugin (closes #187):

```js
// With default options
Encore.enableBuildNotifications();

// Conditionally
Encore.enableBuildNotifications(!Encore.isProduction());

// With custom options
Encore.enableBuildNotifications(true, (options) => {
    options.title = 'Foo';
});
```

The plugin is disabled by default and needs to be installed manually using `yarn add --dev webpack-notifier`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -609,6 +609,30 @@ const publicApi = {
     },
 
     /**
+     * If enabled, display build notifications using
+     * webpack-notifier.
+     *
+     * https://github.com/Turbo87/webpack-notifier
+     *
+     *     Encore.enableBuildNotifications();
+     *
+     *     // or configure the webpack-notifier options
+     *     // https://github.com/Turbo87/webpack-notifier#configuration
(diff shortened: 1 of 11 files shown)


weaverryan:
Wow! Once again, this is just *really* nice work. I'm super excited about this feature :)

PR #195, opened 2017-10-31
Add Stylus loader

mneuhaus:


index.js
```diff
@@ -498,6 +498,26 @@ const publicApi = {
...
 
     /**
+     * Call this if you plan on loading stylus files.
+     *
+     *     Encore.enableStylusLoader();
+     *
+     * Or pass options to the loader
+     *
+     *     Encore.enableStylusLoader(function(options) {
+     *         // https://github.com/shama/stylus-loader
```
Lyrkan: Could you add another line with an example there so people know how to use `options` (using a real one)?

lib/config-generator.js
```diff
@@ -185,6 +186,13 @@ class ConfigGenerator {
             });
         }
 
+        if (this.webpackConfig.useStylusLoader) {
+            rules.push({
+                test: /\.stylus/,
```
Lyrkan: Is the `.stylus` file extension the recommended one? I saw `.styl` used on multiple pages:

* [Stylus Github](https://github.com/stylus/stylus)
* [Stylus documentation](http://stylus-lang.com/docs/#basic-usage)
* [Stylus-loader Github](https://github.com/shama/stylus-loader).

mneuhaus:
i'll try to take care of those tonight :) 👍

lib/config-generator.js
mneuhaus: you're absolutely right, that is a typo, which i'll fix right away :)

index.js
mneuhaus: to be honest, there isn't much that can be configured, i can not think of a good example off the top of my head. (https://github.com/shama/stylus-loader/blob/master/index.js#L33)
The primary reason, for putting this in encore is, that adding a css preprosessor that properly works with encore is a bit much boilerplate: 
```
.addLoader({
    test: /\.styl/,
    use: [
        { loader: './node_modules/extract-text-webpack-plugin/dist/loader.js',
            options: { omit: 1, remove: true } },
        { loader: 'style-loader?sourceMap' },
        { loader: 'css-loader',
            options: {
                minimize: false, sourceMap: true, importLoaders: 0
            }
        },
        { loader: 'stylus-loader'}
    ]
})
```

Lyrkan: Hey @mneuhaus ,

I agree that not many options are available for that loader.
Maybe : `options.import = ['~library/index.styl'];` ?
There is an example [here](https://github.com/shama/stylus-loader#using-nib-with-stylus) in which they do something like that in order to use `nib`.

mneuhaus: sure, sounds fine, i added it :)

mneuhaus:
i added the tests and fixed the mentioned comments :)

lib/features.js
```diff
@@ -26,6 +26,11 @@ const features = {
         packages: ['less-loader'],
         description: 'load LESS files'
     },
+    stylus: {
+        method: 'enableStylusLoader()',
+        packages: ['stylus-loader'],
```
weaverryan: I believe we also need `stylus`, correct?

mneuhaus:
@weaverryan there you go :)

Issue #88, opened 2017-07-11
Incorrect paths for fonts


nikrou:
My context:
I host my webiste under a given path. I do not use symfony but we don't care here.
I used (of course) webpack-encore to manage my assets. It works well except for fonts because of my extra path.

My website is hosted under /my-path. I generate my stylesheets using sass. Stylesheets paths are composed of my base path and the path found in manifest.json.
I've got for example  /my-path/build/css/style.css
I use font-awesome to add web-fonts. The url() in the generated stylesheets is url(/build/fonts/fontawesome-webfont.eot); and so one.
I don't know (just start to investigate) how to fix it. I usually used webpack (without webpack-encore) and webpack used a relative path to fonts from stylesheets instead of an absolute path; something like:

```css
@font-face {
  font-family: 'FontAwesome';
  src: url(./fonts/fontawesome-webfont.eot);
  src: url(./fonts/fontawesome-webfont.eot?#iefix&v=4.7.0) format("embedded-opentype"), url(./fonts/fontawesome-webfont.woff2) format("woff2"), url(./fonts/fontawesome-webfont.woff) format("woff"), url(./fonts/fontawesome-webfont.ttf) format("truetype"), url(./images/fontawesome-webfont.svg#fontawesomeregular) format("svg");
  font-weight: normal;
  font-style: normal;
}
```

And I've got 

```css
@font-face {
  font-family: 'FontAwesome';
  src: url(/my/public/conf/path/build/fonts/fontawesome-webfont.eot);
  src: url(/my/public/conf/path/build/fonts/fontawesome-webfont.eot?#iefix&v=4.7.0) format("embedded-opentype"), url(/my/public/conf/path/build/fonts/fontawesome-webfont.woff2) format("woff2"), url(/my/public/conf/path/build/fonts/fontawesome-webfont.woff) format("woff"), url(/my/public/conf/path/build/fonts/fontawesome-webfont.ttf) format("truetype"), url(/my/public/conf/path/build/images/fontawesome-webfont.svg#fontawesomeregular) format("svg");
  font-weight: normal;
  font-style: normal;
}
```

Any idea to solve that issue is appreciate. If you have any question, don't hesitate.

weaverryan:
Hey @nikrou!

[...]

Issue #26, opened 2017-06-16
Wrong path fonts


dayze:
When I include semantic or bootstrap in my main.js I end up with a 404 not found error for the fonts

Apparently, the output path of fonts go to the root of my web server like this

`http://localhost/build/fonts/icons.woff2
`

weaverryan:
Hey @dayze!

Can you post your webpack.config.js file and also the code for any relevant other files (like main.js)?

dayze:
Hi, 

**webpack.config.js**

```
var Encore = require('@symfony/webpack-encore');

Encore
// directory where should all compiled assets will be stored
    .setOutputPath('web/build/')

    // what's the public path to this directory (relative to your project's document root dir)
    .setPublicPath('/build')

    // empty the outputPath dir before each build
    .cleanupOutputBeforeBuild()

    // will output as web/build/app.js
    .addEntry('app', './assets/js/main.js')


    // will output as web/build/global.css
    .addStyleEntry('global', './assets/css/global.scss')


    // allow sass/scss files to be processed
    .enableSassLoader()

    // allows legacy applications to use $/jQuery as a global variable
    .autoProvidejQuery()


    .enableSourceMaps(!Encore.isProduction())



// create hashed filenames (e.g. app.abc123.css)
// .enableVersioning()
;

var webpackConfig = Encore.getWebpackConfig();
webpackConfig.module.rules.push({
    test: /\.vue$/,
    loader: 'vue-loader'
});
webpackConfig.module.rules.push({
    test: /\.js$/,
    loader: 'babel-loader',
    exclude: /node_modules/
});

// export the final configuration
module.exports = webpackConfig;
```

**main.js**
```
import Vue from 'vue/dist/vue'
import App from './app.vue'

require('semantic-ui/dist/semantic.min.css')

```

**Errors**
```
GET http://localhost/build/fonts/icons.woff2 
GET http://localhost/build/fonts/icons.woff 
GET http://localhost/build/fonts/icons.ttf 

```

heitorvrb:
Same issue here, with similar configuration. The situation seems to be that my app doesn't run on a ground-level domain, but on a alias on apache:
```
Alias /mySymfonyApp /var/www/html/mySymfonyApp/web
```
So webpack is trying to create the `src: url()` paths with `/build/fonts/myFont.eot`, when it should be `/mySymfonyApp/build/fonts/myFont.eot`.

webpack.config.js:
```
var Encore = require('@symfony/webpack-encore');

[...]

PR #196, opened 2017-11-08
Replace error throwing with logger warning when publicPath is not absolute

andyexeter:
There are valid use cases for setting the publicPath to something not starting with a forward slash as discussed in #88 and #26 

Throwing an error and stopping compilation is overkill for this and beyond the remit of webpack encore, so I propose this PR where a warning is given to the user instead and the `setPublicPath` method falls in line with behaviour of `setManifestKeyPrefix`

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -127,7 +127,7 @@ class WebpackConfig {
             // technically, not starting with "/" is legal, but not
             // what you want in most cases. Let's warn the user that
             // they might be making a mistake.
-            logger.warning('The value passed to setPublicPath() should usually start with "/" or be a full URL (http://...)');
+            logger.warning('The value passed to setPublicPath() should *usually* start with "/" or be a full URL (http://...). If you\'re not sure, then you should probably change your public path and make this message disappear.');
         }
 
         // guarantee a single trailing slash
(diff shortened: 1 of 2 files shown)


weaverryan:
Since a lot of people are asking about the relative path stuff, I also think we should work on support for it (we just need some time to play with it & experiment - Webpack itself may have some issues with relative paths).

The question is - should we merge this now when we know there is a legit issue (https://github.com/symfony/webpack-encore/issues/26#issuecomment-341977357) that needs to be figured out?

Lyrkan:
In my opinion merging it now won't be a big issue since the user would still be warned, and as @andyexeter said we already kinda do the same thing in `setManifestKeyPrefix`:

https://github.com/symfony/webpack-encore/blob/bcfbbfd6ab684197b2315f850dcb2db56172ea34/lib/WebpackConfig.js#L140-L152

andyexeter:
Hey @weaverryan - do you think you could take a look at merging this?

The issue you referenced in #26 wouldn't change whether this was merged or not, and the warning would still let people know they could experience issues when using a relative path.

It should be down to the end user to decide whether they want to use a relative path or not rather than being forced not to by the plugin.

kevinpapst:
Would be great if you look into this issue again @andyexeter and @weaverryan.
I am working on migrating an open source app to Symfony 4 and webpack-encore, but I am not able to release a build off my app that runs inside a subdirectory and on domain-level. 
I simply don't know where my user will deploy the app and no matter how I generate the files, 50% of my user will complain about missing assets.

weaverryan:
Merged and I've just tagged 0.17.2. If you use this feature, I would appreciate it if you posted any feedback or "findings" on #26 so that we can determine of the relative paths work, or if it causes issues.

Thanks!

PR #207, opened 2017-11-17
Update webpack-manifest-plugin.js

Detract:
Version 1.1.0 is outdated. It took me some time to figure out why configuration options like map and filter does not work with encore.

### DIFF
--- a/lib/webpack/webpack-manifest-plugin.js
+++ b/lib/webpack/webpack-manifest-plugin.js
@@ -1,6 +1,6 @@
 /*
  * File is originally a direct copy from https://github.com/danethurber/webpack-manifest-plugin
- * at version 1.1.0.
+ * at version 1.3.2.
  *
  * The library is licensed as MIT.
  *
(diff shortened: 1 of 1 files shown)


Lyrkan:
Hi @Detract,

We're waiting for the new major version or webpack-manifest-plugin to be released (see #164) so we can remove that file and use the real plugin.

We needed that fork because of some issues that were directly patched in the file you modified and are should be fixed in v2.0.0 (currently in rc1, no idea if there is an ETA for a stable version, ping @mastilver). I guess these missing patches are the cause of the failing tests.

Detract:
Hi @Lyrkan,

thanks for response. I'll then close the pr and wait for the new version.

Dirk

mastilver:
Hi @Lyrkan 

I don't want to release v2 before https://github.com/danethurber/webpack-manifest-plugin/issues/100 is fixed, and I don't have much time at the moment, I will try later next week

Lyrkan:
Hey @mastilver!

No worries, there is still some work to be done on #164 anyway :)

PR #215, opened 2017-11-25
Fix output when using "--json" or "-j" option

### LINKED ISSUE #204: Fix support for --json option
The `--json` option (e.g. `encore production --json`) is useful for visualizing your assets: https://chrisbateman.github.io/webpack-visualizer/

There are at least 2 things that Encore currently breaks:

1) We dump some output, which breaks the JSON
2) We silence a bunch of output, which the visualizer expects

Lyrkan:
This PR fixes #204 by detecting the `--json` and `-j` options and:

* removing some `console.log` calls
* removing the `FriendlyErrorsWebpackPlugin`
* Using default `stats` values

As said in [that comment](https://github.com/symfony/webpack-encore/issues/204#issuecomment-346950361) of the related issue the `--silent` option has to be used with `yarn run encore` or `yarn encore` and must be placed before the name of the script.

For instance:

```bash
# Works
./node_modules/.bin/encore production --json
yarn run --silent encore production  --json
yarn --silent encore production  --json

# Doesn't work (some lines being added by yarn)
yarn run encore production --json
yarn encore production --json

# Doesn't work (since Webpack doesn't know the --silent option)
yarn run encore production --silent --json
yarn encore production --silent --json
```

Edit: Forgot to add a test yesterday, should be good now :)

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -45,8 +45,10 @@ if (runtimeConfig.useDevServer) {
 
     return require('webpack-dev-server/bin/webpack-dev-server');
 } else {
-    console.log('Running webpack ...');
-    console.log();
+    if (!runtimeConfig.outputJson) {
+        console.log('Running webpack ...');
+        console.log();
+    }
 
     return require('webpack/bin/webpack');
 }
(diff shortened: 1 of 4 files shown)


weaverryan:
Thanks @Lyrkan :). Sorry for the very slow merge - will start catching up soon (this was a very easy merge!)

PR #217, opened 2017-11-25
Add Encore.addAliases and Encore.addExternals methods

### LINKED ISSUE #200: How to add alias and externals in encore
Hello,

Since webpack is obscure to me, I'm giving `encore` a chance but i'm strugling with how to add an alias and externals entry

```
resolve: {
      extensions: ['.js', '.json', '.vue'],
      alias: {
        Translator: 'node_modules/bazinga-translator/js',
      },
    },
    externals: {
      vue: 'Vue',
      vuex: 'Vuex',
      google: 'google',
      leaflet: 'L',
      markerclustererplus: 'MarkerClusterer',
      routing: 'Routing',
      app: 'app',
      translator: 'Translator',
    },
```

I tried `ctrl+f` in `encore` API but didn't found it :(

Lyrkan:
This PR adds two new methods to the API (closes #200):

* **`Encore.addAliases`**: Adds the given values to the `resolve.alias` setting of the generated config
* **`Encore.addExternals`**: Adds the given values to the `external` setting of the generated config

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -77,7 +77,7 @@ class ConfigGenerator {
 
         config.resolve = {
             extensions: ['.js', '.jsx', '.vue', '.ts', '.tsx'],
-            alias: {}
+            alias: this.webpackConfig.aliases
         };
 
         if (this.webpackConfig.useVueLoader) {
@@ -88,6 +88,8 @@ class ConfigGenerator {
             config.resolve.alias['react'] = 'preact-compat';
             config.resolve.alias['react-dom'] = 'preact-compat';
         }
+
+        config.externals = this.webpackConfig.externals;
 
         return config;
     }
--- a/index.js
+++ b/index.js
@@ -329,6 +329,51 @@ const publicApi = {
     },
 
     /**
+     * Allow you to add aliases that will be used by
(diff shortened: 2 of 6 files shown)


lambertbeekhuis:
Can you merge this!!

lib/config-generator.js
```diff
@@ -77,7 +77,7 @@ class ConfigGenerator {
 
         config.resolve = {
             extensions: ['.js', '.jsx', '.vue', '.ts', '.tsx'],
-            alias: {}
+            alias: this.webpackConfig.aliases
```
stof: you should create a copy of it, so that modifying the generated webpack config object cannot modify the Encore config (which might be used again, even if it is probably only an edge case to reuse it without resetting).
this also applies to externals.

Lyrkan: Definitely, nice catch, I'll open a new PR for that.

I'm a bit worried that there is the same issue with other objects though... for instance when calling `Encore.addLoader()` and `Encore.addPlugin()`.

stof: Well, if we don't mutate the object ourselves, it might be less of an issue. but here, we are doing the mutation ourselves in some cases

PR #221, opened 2017-12-04
Add a copyFiles() method to the public API (using CopyWebpackPlugin)

### LINKED ISSUE #175: Add a copyFiles() Method
See https://github.com/symfony/webpack-encore/pull/165#issuecomment-331996320

Lyrkan:
**Note: Can't be merged yet as it needs #164 for the tests to pass. Right now everything should work apart from the copied files not being added to the manifest.json (see https://github.com/danethurber/webpack-manifest-plugin/pull/75).**

This PR adds a `copyFiles()` method to the API (closes #175):

```js
Encore.copyFiles([
    'file.txt'
     { from: 'from/file.txt' },
     { from: 'from/file.txt', to: 'to/file.txt' },
     { from: 'from/file.txt', to: 'to/directory' },
     { from: 'from/directory' },
], (options) => {
    options.ignore = ['*.txt']
});
```

Multiple calls will add new instances of the plugin, each one being configured using the patterns and callback it has been added with.

### DIFF
--- a/index.js
+++ b/index.js
@@ -799,6 +799,32 @@ class Encore {
     }
 
     /**
+     * Add files or directories to be copied to the build directory.
+     *
+     * A list of valid pattern properties and available options can be
+     * found at https://github.com/webpack-contrib/copy-webpack-plugin#usage
+     *
+     *     Encore.copyFiles([
+     *       'file.txt'
+     *       { from: 'from/file.txt' },
+     *       { from: 'from/file.txt', to: 'to/file.txt' },
+     *       { from: 'from/file.txt', to: 'to/directory' },
+     *       { from: 'from/directory' },
+     *     ], (options) => {
+     *       options.ignore = ['*.txt']
+     *     });
+     *
+     * @param {object} patterns
+     * @param {function} copyWebpackPluginOptionsCallback
+     * @returns {Encore}
+     */
+    copyFiles(patterns, copyWebpackPluginOptionsCallback) {
+        webpackConfig.copyFiles(patterns, copyWebpackPluginOptionsCallback);
+
+        return this;
+    }
+
+    /**
      * If enabled, the output directory is emptied between each build (to remove old files).
      *
      * A list of available options can be found at https://github.com/johnagan/clean-webpack-plugin
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -90,6 +90,7 @@ class WebpackConfig {
 
         // Plugins options
         this.cleanWebpackPluginPaths = ['**/*'];
+        this.copyWebpackPluginConfigs = [];
 
         // Plugins callbacks
         this.cleanWebpackPluginOptionsCallback = () => {};
@@ -484,6 +485,21 @@ class WebpackConfig {
         }
 
         this.urlLoaderOptions = urlLoaderOptions;
+    }
+
+    copyFiles(patterns = [], copyWebpackPluginOptionsCallback = () => {}) {
+        if (!Array.isArray(patterns)) {
+            throw new Error('Argument 1 to copyFiles() must be an Array of patterns - e.g. [{ from: \'from/file.txt\' }]');
+        }
+
+        if (typeof copyWebpackPluginOptionsCallback !== 'function') {
+            throw new Error('Argument 2 to copyFiles() must be a callback function');
+        }
+
+        this.copyWebpackPluginConfigs.push({
+            patterns: patterns,
+            optionsCallback: copyWebpackPluginOptionsCallback,
+        });
     }
 
     cleanupOutputBeforeBuild(paths = ['**/*'], cleanWebpackPluginOptionsCallback = () => {}) {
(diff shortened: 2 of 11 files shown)


weaverryan:
We're just waiting on the stable release: https://github.com/danethurber/webpack-manifest-plugin/releases

Technically, I believe we could use the RC1 as a package and NOT wait for stable... but hopefully it'll happen soon - I pinged the maintainers.

JohanLopes:
@weaverryan @Lyrkan The new version of Webpack manifest version has been released : https://github.com/danethurber/webpack-manifest-plugin/releases

Lyrkan:
@JohanLopes Yep, waiting for #164 to be updated and merged before rebasing that one :)

weaverryan:
Ok, #164 is merged! go go go! :)

Lyrkan:
@weaverryan I rebased it but I'm a bit annoyed with some of the test results:

```js
Encore.copyFiles([{ from: 'images/symfony_logo.png', to: 'symfony_logo.[hash:8].png' }]);

// Expected manifest.json
{"build/symfony_logo.png": "/build/symfony_logo.ea1ca6f7.png"}

// Actual manifest.json
{"build/symfony_logo.ea1ca6f7.png": "/build/symfony_logo.ea1ca6f7.png"}
```

I'm not sure what to do about that...

Also (but less of an issue so I didn't include it in my tests):

```js
Encore.copyFiles([{ from: 'images/symfony_logo.png', to: './symfony_logo.png' }]);

// Expected manifest.json
{"build/symfony_logo.png": "/build/symfony_logo.png"}

// Actual manifest.json
{"build/./symfony_logo.png": "/./build/symfony_logo.png"}
```

Lyrkan:
I just noticed that the first issue was already discussed there: https://github.com/danethurber/webpack-manifest-plugin/pull/75#issuecomment-325614092

So... nothing can really be done for now since the Copy Webpack Plugin doesn't expose the original filename (that PR could help though: https://github.com/webpack-contrib/copy-webpack-plugin/pull/198).

weaverryan:
Boooooooooooo

Just to throw out an idea... if we copied that plugin (yet again) to our code, are there some small changes that would allow us to have this feature? It's so important, I'd rather be pragmatic over pure.

weaverryan:
Oh, and since you understand the problem well, I think you should give some good feedback on /webpack-contrib/copy-webpack-plugin#198 if it's the right approach (more than just 👍, mention why we need it & your issue - the maintainer is good about listening when somebody comes in with good technical knowledge).

deguif:
Just seing this conversation.

I found a way to bypass this issue a little time ago, having a look at the following comment can help with solving this problem: https://github.com/webpack-contrib/copy-webpack-plugin/issues/104#issuecomment-370174211

The bypass is not very clean but it allows to generate a manifest with proper key when using hash.

Lyrkan:
@deguif I saw your comment but sadly I don't think this is generic enough to add into this PR (it would not work for some filenames/configurations).

I'm also not sure that https://github.com/webpack-contrib/copy-webpack-plugin/pull/198 would be the right fix anymore... as some other people pointed out in that thread an event-based approach would probably be a better idea.

deguif:
Clearly I don't think webpack-contrib/copy-webpack-plugin#198 is a right fix, it would also be redundant with the use of webpack manifest plugin.

weaverryan:
I haven’t looked at the specifics, but if there is a workaround that would work for us, but not be exposed to our users (so if we change in the future, they won’t be affected), then we should check into it. It’s tough, this is such an important feature but one that has been continually blocked.

Lyrkan:
@weaverryan I looked again at this PR today and I don't think there is a viable workaround.

Imagine that you have the following config (which is more or less the one I used for the failing functional test):

```js
Encore.copyFiles([
  { from: 'images/symfony_logo.png', to: 'symfony_logo1.png' },
  { from: 'images/symfony_logo.png', to: 'symfony_logo2.png' },
  { from: 'images/symfony_logo.png', to: 'symfony_logo3.[hash:8].png' },
]);
```

What do you expect to get in the manifest file? Maybe something like this (which is what @deguif 's solution does if you only use the `[name].[hash].[ext]` pattern):

```js
{
  'build/symfony_logo.png': '/build/symfony_logo1.png',
  'build/symfony_logo2.png': '/build/symfony_logo2.png',
  'build/symfony_logo3.png': '/build/symfony_logo3.ea1ca6f7.png'
}
```

But no... the key should actually always be `build/symfony_logo.png` because the left part is supposed to be the original name of the file (you can't guess what you're supposed to get from the final filename). This is annoying because currently the generated manifest would look like this:

```js
{
  'build/symfony_logo1.png': '/build/symfony_logo1.png',
  'build/symfony_logo2.png': '/build/symfony_logo2.png',
  'build/symfony_logo3.ea1ca6f7.png': '/build/symfony_logo3.ea1ca6f7.png'
}
```

... which doesn't contain the original filename at all.

So, unless the CopyWebpackPlugin provides a way to get more info there isn't much we (or the WebpackManifestPlugin) can do... and then you'd probably still have some edge cases to handle.

I'm wondering if we couldn't do something simpler without the CopyWebpackPlugin... like creating a fake entry point that `require` the given files using the `file-loader`.

Lyrkan:
Closing that one since #409 has been merged :)

Issue #232, opened 2017-12-30
Add support for eslint


pinoniq:
Webpack encore adds built in support for a coupld of "wide spread" front end libraries. Being ReactJs, jQuery and SASS.

There is however one that is even more widespread: eslint. Having an .enableEslint() (or even enabling it by default in non production mode) would be a plus.

However, if we go along the "default should be good enough" path. This means taking a side in the many different code-styles that exist in the JavaScript world.

A small list:
* https://www.drupal.org/docs/develop/standards/javascript/javascript-coding-standards
* https://google.github.io/styleguide/jsguide.html
* https://github.com/airbnb/javascript
* https://github.com/rwaldron/idiomatic.js/
* https://github.com/standard/standard

Instead of choosing a side, symfony could use the default eslint, with the possibility to change the "extends" property.

```
.enableEslint({
  extends: "airbnb"
})
```

Combining this with ```.autoProvidejQuery()``` and ```.enableReactPreset()``` would basically resort to something like
```
.addLoader({
  test: /\.(js|jsx)$/,
  loader: 'eslint-loader',
  exclude: [/node_modules/],
  enforce: "pre",
  options: {
    "extends": "airbnb",
    "parser": "babel-eslint",
    "settings": {
      "import/resolver": {
        "webpack": {
          "config": "webpack.config.js"
        }
      }
    },
    "globals": {
      "$": false
    }
    emitWarning: true
  }
})
```

Is this something webpack encore is willing to add? If so, what path do we take?

weaverryan:
Hi there!

Interesting! So basically, this would output build warnings if eslint fails, right? I think it’s interesting; after all, it would be opt-in. But mostly, I like it including this in your Webpack build is considered a fairly common “best practice” (versus something that’s “possible”, but few people do it). Do you have an impression about that?

pinoniq:
At the company I work every webpack/gulp/grunt build task includes eslint. This to avoid having the developer to miss linting errors and only see it on post-commit instead of during his work.

Especially given the fact it adds very little overhead and allows you to see eslint errors "live" in the console running the encore.

How far this is a "best practice" in the world I don't really know. The eslint-loader we use has about 40k downloads a day: https://www.npmjs.com/package/eslint-loader . Most blogs even include it in the "default webpack/react setup" and I'm used to it. Haven't spoken about it to other people on conferences however.

And yes, it should be opt-in. Especially given the Symfony flex philosophy.

pinoniq:
[...]

PR #243, opened 2018-01-18
#232 Added support for the eslint-loader

pinoniq:
I wasn't sure about whether to include the eslint-loader in the devDependencies or dependencies.


index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
...
+     *     Encore.enableEslint('airbnb');
+     *
+     *     // You can also pass in an object of options
+     *     // that will be passed on to the eslint-loader
+     *     Encore.enableEslint({
+     *         extends: 'airbnb',
+               emitWarning: false
+     *     });
+     *
+     *     // For a mroe advanced usage you can pass in a callback
```
Lyrkan: mroe => more

index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
...
+     *     // https://github.com/MoOx/eslint-loader#options
+     *     Encore.enableEslint((options) => {
+     *          options.extends = 'airbnb';
+     *          options.emitWarning = fasle;
```
Lyrkan: fasle => false

lib/features.js
```diff
@@ -63,6 +63,12 @@ const features = {
         packages: ['vue', 'vue-loader', 'vue-template-compiler'],
         description: 'load VUE files'
     },
+    eslint: {
+        method: 'enableEslintLoader()',
+        // eslint is needed so the end-user can do things
+        packages: ['eslint', 'eslint-loader', 'babel-eslint', 'eslint-plugin-import'],
+        description: 'load VUE files'
```
Lyrkan: "Enable ESLint checks" ?

lib/loaders/eslint.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+module.exports = {
+    getOptions(webpackConfig) {
+        loaderFeatures.ensurePackagesExist('eslint');
+
+        const eslintLoaderOptions = {
+            parser: 'babel-eslint',
+            emitWarning: true,
+            rules: {
+                'linebreak-style': 'off'
```
Lyrkan: Should we really disable that rule by default?

index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
         return this;
     },
 
+    /**
+     * If enabled, the eslint-loader is enabled.
+     *
+     * https://github.com/MoOx/eslint-loader
+     *
+     *     // enables the eslint loaded using the default eslint configuration.
+     *     Encore.enableEslint();
```
Lyrkan: enableEslint => enableEslintLoader

index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
...
+     *
+     *     // Optionally, you can pass in the configuration eslint should extend.
+     *     Encore.enableEslint('airbnb');
```
Lyrkan: enableEslint => enableEslintLoader

index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
...
+     *
+     *     // You can also pass in an object of options
+     *     // that will be passed on to the eslint-loader
+     *     Encore.enableEslint({
```
Lyrkan: enableEslint => enableEslintLoader

index.js
```diff
@@ -629,6 +631,40 @@ const publicApi = {
...
+     *         extends: 'airbnb',
+               emitWarning: false
+     *     });
+     *
+     *     // For a mroe advanced usage you can pass in a callback
+     *     // https://github.com/MoOx/eslint-loader#options
+     *     Encore.enableEslint((options) => {
```
Lyrkan: enableEslint => enableEslintLoader

index.js
```diff
@@ -288,9 +288,11 @@ const publicApi = {
      * than the DefinePlugin:
      *
      *      const Encore = require('@symfony/webpack-encore');
-     *      const PluginPriorities = require('@symfony/webpack-encore/lib/plugins/plugin-priorities.js');
+     *      const PluginPriorities =
+     * require('@symfony/webpack-encore/lib/plugins/plugin-priorities.js');
```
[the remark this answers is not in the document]
pinoniq: it's Jetrbains :p I'll remove it

index.js
```diff
@@ -288,9 +288,11 @@ const publicApi = {
...
      *
-     *      Encore.addPlugin(new MyWebpackPlugin(), PluginPriorities.DefinePlugin);
+     *      Encore.addPlugin(new MyWebpackPlugin(),
+     * PluginPriorities.DefinePlugin);
```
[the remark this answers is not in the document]
pinoniq: again: it's Jetrbains :p I'll remove it

lib/features.js
pinoniq: obvious copy paste :)

lib/loaders/eslint.js
pinoniq: Given most people on windows have git doing a "checkout crlf, commit lf", I feel it's an option that should be default everywhere.

pinoniq:
@Lyrkan I think it's probably good to have a follow up ticket to have a common interface for all enableXXX() methods. This would allow us to abstract a part away into a helper method that "automaigcally" transforms the passed in parameter(s) into a optionsCallback.

lib/loaders/eslint.js
Lyrkan: I'm not sure that should be Git's job to do that kind of thing, especially since that's not its default behavior on Windows (you have to switch the `core.autocrlf` manually).

A cleaner way to handle that IMO is to use an `.editorconfig` file (which is supported by many editors) or change manually the config of your IDE so it doesn't use CRLF line breaks.

Note that this rule is also enabled in Airbnb and Google ESLint presets:

https://github.com/airbnb/javascript/blob/69a3554100935b79b9c444102504f98103834079/packages/eslint-config-airbnb-base/rules/style.js#L158-L160

https://github.com/google/eslint-config-google/blob/394bf3c9f858b83514fdcbf23d74492e253d611f/index.js#L214

pinoniq: Hmm, true. With default, I was referring to tools like the github desktop client and Atlassian sourcetree. I'll remove this config :)

pinoniq:
Hey, concerning the functional test. Not sure how to continue here since I would not be testing creation of files, but console output. If you point into the right direction I'm more than willing to look into it

Lyrkan:
@pinoniq You could do something like that: https://gist.github.com/Lyrkan/94aa818b11b35715857a52471d43e6de

The idea being to check the `stats` object returned by Webpack (we already do that in the "Vue.js error when using non-activated loader" test).

Some things you have to be careful though:

* you need to force eslint-loader to emit errors instead of warning (you could technically use `stats.toJson().warnings` but warnings would pollute the tests results since we display them [here](https://github.com/symfony/webpack-encore/blob/00e4051eb0ef18786a9d6abe2f9f2e73c887e8fa/lib/test/setup.js#L91-L93))
* be aware that the `.eslintrc.js` file at the root of Encore will be used and that if we change something in it later your test may break. To avoid that you can force the rules you'll use in your checks when calling `enableEslintLoader`

Another way to test that would be to emit a report (using the `outputReport` option) and check it once webpack is done processing everything.

pinoniq:
@Lyrkan I added the test (by using your gist). Concerning the warnings/errors kind of complicates a "proper" functional test. This because eslint allows you to define errors and warnings.

It is possible to easy test this, but it pollutes the test output (due to the warnings being displayed during test results like you pointed out). What do you prefer here? Only testing for the errors output, or also adding the test for warnings and errors? If the later, maybe look into adding a flag to runWebpack to not output warnings?
e.g. 
`function runWebpack(webpackConfig, callback, allowCompilationError = false, logWarnings = true) {`

Lyrkan:
👍 

@pinoniq I think the test using errors should be enough... we only need to check if ESlint is enabled and takes our conf into account, not how it behaves if we use other settings ([that's already part of the ESlint test suite](https://github.com/eslint/eslint/blob/9cbb487e8e84ff3268c05ab47b3b72c51b5f766e/tests/lib/cli.js#L410-L420))

pinoniq:
Ok. valid point. You want me to squash the commits into a single one? Or how do you guys do this here?

lib/loaders/eslint.js
```diff
@@ -0,0 +1,40 @@
...
+        loaderFeatures.ensurePackagesExist('eslint');
+
+        const eslintLoaderOptions = {
+            parser: 'babel-eslint',
+            emitWarning: true,
+            'import/resolver': {
+                webpack: {
+                    config: 'webpack.config.js'
+                }
+            }
```
weaverryan: Can you explain this config? What does this do exactly?

lib/loaders/eslint.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @param {WebpackConfig} webpackConfig
+ * @return {Object} of options to use for eslint-loader options.
+ */
+module.exports = {
+    getOptions(webpackConfig) {
+        loaderFeatures.ensurePackagesExist('eslint');
+
+        const eslintLoaderOptions = {
+            parser: 'babel-eslint',
```
weaverryan: According to their docs:

> You only need to use babel-eslint if you are using types (Flow) or experimental features not supported in ESLint itself yet. Otherwise try the default parser (you don't have to use it just because you are using Babel).

I don't think this applies to Encore users with "standard" config. If they're using experimental features, then they would need to configure some extra babel presets or plugins (which is something custom). So, I'm wondering if there's a good enough reason to use *this* parser.

lib/features.js
```diff
@@ -63,6 +63,12 @@ const features = {
         packages: ['vue', 'vue-loader', 'vue-template-compiler'],
         description: 'load VUE files'
     },
+    eslint: {
+        method: 'enableEslintLoader()',
+        // eslint is needed so the end-user can do things
+        packages: ['eslint', 'eslint-loader', 'babel-eslint', 'eslint-plugin-import'],
```
weaverryan: I don't quite understand what the `eslint-plugin-import` is supposed to do - obviously something related to parsing imports, etc :). Would it be possible to enhance the `fixtures/eslint.js` file to take advantage of a feature supplied by this library, so we can make sure it's configured correctly?

index.js
```diff
@@ -629,6 +629,40 @@ const publicApi = {
...
+     *
+     *     // Optionally, you can pass in the configuration eslint should extend.
+     *     Encore.enableEslintLoader('airbnb');
+     *
+     *     // You can also pass in an object of options
+     *     // that will be passed on to the eslint-loader
+     *     Encore.enableEslintLoader({
+     *         extends: 'airbnb',
+               emitWarning: false
+     *     });
```
weaverryan: Bah. So, this looks really great. But so far, we've avoided allowing users to pass in options like this... because of potential merging issues between their config and our config. I really can't think of an issue here - their config wins, should be pretty straightforward. But it breaks from our convention. @Lyrkan what do you think?

lib/config-generator.js
```diff
@@ -201,6 +202,16 @@ class ConfigGenerator {
             });
         }
 
+        if (this.webpackConfig.useEslintLoader) {
+            rules.push({
+                test: /\.jsx?$/,
+                loader: 'eslint-loader',
+                exclude: /node_modules/,
+                enforce: 'pre',
+                options: eslintLoaderUtil.getOptions(this.webpackConfig)
```
weaverryan: Should we use `cache: true`?

index.js
Lyrkan: Well, the main issue I have with it is that we don't do the same thing for other methods...
It isn't a really big issue since in this you can also use a callback but it could confuse people.

I'd say that we have to decide whether or not we'll implement the same kind of trick (and also always do `Object.assign(defaultConfig, userConfig)`) to other methods.

lib/features.js
pinoniq: given this is only useful if you actually use es2015+ import syntax, maybe even just remove it, no?

lib/loaders/eslint.js
pinoniq: because you can define additional alias in webpack, this import/Resolver allows eslint to verify that libraries/files exist.

lib/config-generator.js
pinoniq: probably

index.js
pinoniq: Like lyrkan said, we should probably aggree on a common interface to configure the different plugins

weaverryan: Ok, let's keep it. And then we can think about adding this on a case-by-case basis. I think sometimes it's *not* a good idea, because we'll need to merge the config. But in this case, even though we're merging the config, the config is flat and simple. There may be some cases where the merging could be more complex, and we could possibly not allow this then :)

lib/features.js
weaverryan: If we remove it, and someone *does* use `import`, what happens? What do we lose? Do we just lose some feature of eslint that validates that those paths are correct? Or is it something different?

lib/loaders/eslint.js
weaverryan: So, with `config: 'webpack.config.js'`, we are telling `eslint` that it should internally require our `webpack.config.js` file to learn about the aliases? If so, I think that would cause an error - as the Encore "runtime" config wouldn't be defined. Are we sure this works? If we can add a test case for it in `functional.js`, I would be SUPER happy :)

lib/loaders/eslint.js
pinoniq: No it doesnt. I should probably check here if enableBable or enableRract thingies are called and add by default.

lib/features.js
pinoniq: Just re-read the docs. if you don't use the import sytnac, but e.g. a commonJS or AMD syntax, then this will throw errors about this.
Personally, I prefer going for "enable by default" since this is my personal (and companies) use-case. However i'm not sure if we can simply enforce it. It should be easily disabled/enabled atleast.

Maybe look into having a way of advising users to install a package? or checking at run-time if it's installed or not?

index.js
pinoniq: I can also remove this "nifty" merging and create a follow up ticket where we can further discuss/address. This to avoid breaking backward capabilities.

lib/loaders/eslint.js
weaverryan: Hmm. Yea.... now that I process you reply, I *do* think it should be added always :p. The reason is that there is NOT an `enableBabel()` method - it's *always* enabled. And so, even though most users won't use any special syntax, we should try to make lives easier in case they do. If we *did* allow a user to disable babel in the future, *then* we could backup to the normal linter.

lib/features.js
weaverryan: Hmm. Let me see if I understand this correctly:

A) *Without* `eslint-plugin-import`, any `import` statements aren't validated, so in theory, you could be importing a non-existent path and would go no lint error. Are normal `require` paths linted?

B) *With( `eslint-plugin-import`, `import` statements *are* validated, but now you MUST use `import` in all situations: any calls to `require` will fail.

Is this correct? Actually, when I'm reading, I think (B) might be *incorrect*. I see the rule you're talking about - `eslint-plugin-import`. But under [Installation](https://www.npmjs.com/package/eslint-plugin-import#installation), it talks about how all the rules are disabled by default.

If we're not sure, then "less" and more "default" setup is better. The most important thing is that we're creating a really easy "system" for enabling eslint AND for adding your own plugins.

index.js
weaverryan: As long as @Lyrkan agrees with the idea of starting to add this config where it makes sense, I'm cool with keeping it here for the new method. It won't break BC - it's just a "new philosophy" for us

lib/loaders/eslint.js
weaverryan: Ping! This is my main concern / question about this PR. I simply don't understand what this config does, so I'm uncomfortable merging. Can you guide me @pinoniq?

lib/config-generator.js
weaverryan: Can you add the `cache: true`?

weaverryan:
ping @pinoniq! This is basically done (or maybe completely done), but I had a few pending questions/concerns that I need your help answering :)

lib/loaders/eslint.js
pinoniq: Ok, so I leave it in now?

lib/features.js
pinoniq: See my comment in the other track, I think it's best to remove this required dependency.

FYI, require still works. This plugin just analyses your import statements.

lib/config-generator.js
pinoniq: fixed

lib/features.js
weaverryan: I’ll take a closer look at your updates and comments soon, but yea, let’s remove this :).

Thanks for your work!

lib/loaders/eslint.js
weaverryan: Yes! I think this is exactly what we should do. So, once you've removed this config and the reference to the package, I think this is ready :).

lib/features.js
pinoniq: @weaverryan Anything left for me on this MR?

pinoniq:
@Lyrkan  @weaverryan I believe all issues have been corrected in the meantime. Are there still some blocking issues left?

you want me to resolve the conflicts for you? or will you be doing that doing the merge?

Lyrkan:
@pinoniq Sorry for the delay, I just rebased it and:

* fixed a line-endings issue in your last commit (it looks like your IDE switched to CRLF for that one)
* used the `apply-options-callback` method added in #300
* changed the `index.js` file a bit since we're now using a `class` (see #263)
* fixed the `cache: true` that wasn't added at the right place and caused the tests to fail

However I think [that comment](https://github.com/symfony/webpack-encore/pull/243/files#r181560041) from @weaverryan hasn't been taken into account yet (since there still is a reference to the `webpack.config.js` file), could you take a look at it?

pinoniq:
@Lyrkan Thx for the rebase. I also removed that reference of the import/resolver default config. Completly forgot about it. It should be ok now.

Lyrkan:
I fixed some test cases that were broken due to the removal of the `import/resolver` conf and changed the packages check related to enableEslintLoader() so it doesn't include `eslint-plugin-import` anymore.

LGTM now (ping @weaverryan)

index.js
stof: well, the callback API now support both editing the config to add more settings, and returning a new object. I'm not sure this automatic merging brings much value here

weaverryan:
Thank you for your hard work and patience @pinoniq! This one is now merged :)

lyrixx:
Hello. It would be nice to document this PR. I see in the config (by reading the code) that I could enable eslint, but I can not make it work.

```
bash-4.4$ yarn encore prod
yarn run v1.12.3
$ /home/app/app/offithings/node_modules/.bin/encore prod
Running webpack ...

 ERROR  Failed to compile with 1 errors                                                                                                                                             8:01:13 PM

Module build failed (from ./node_modules/eslint-loader/index.js):
Error: No ESLint configuration found.
    at Config.getLocalConfigHierarchy (/home/app/app/offithings/node_modules/eslint/lib/config.js:268:39)
    at Config.getConfigHierarchy (/home/app/app/offithings/node_modules/eslint/lib/config.js:192:43)
    at Config.getConfigVector (/home/app/app/offithings/node_modules/eslint/lib/config.js:299:21)
    at Config.getConfig (/home/app/app/offithings/node_modules/eslint/lib/config.js:342:29)
    at processText (/home/app/app/offithings/node_modules/eslint/lib/cli-engine.js:181:33)
    at CLIEngine.executeOnText (/home/app/app/offithings/node_modules/eslint/lib/cli-engine.js:690:40)
    at lint (/home/app/app/offithings/node_modules/eslint-loader/index.js:218:17)
    at transform (/home/app/app/offithings/node_modules/eslint-loader/index.js:201:18)
    at /home/app/app/offithings/node_modules/loader-fs-cache/index.js:127:18
    at ReadFileContext.callback (/home/app/app/offithings/node_modules/loader-fs-cache/index.js:31:14)
    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:238:13)

You may use special comments to disable some warnings.
Use // eslint-disable-next-line to ignore the next line.
Use /* eslint-disable */ to ignore all warnings in a file.
Entrypoint app = runtime.fa8f03f5.js app.5b7010d6.js
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Should I create a configuration ? I'm a bit confused 

Thanks

Lyrkan:
@lyrixx Yep, you have to define ESLint rules using one of the formats listed there: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

PR #248, opened 2018-01-25
feat(aurelia): Add an Aurelia plugin option and expose module resolve

tomtomau:
I thought Symfony needed some Aurelia love :)

I'm fairly new to Webpack so any feedback is welcomed

As Aurelia leverages dynamic modules, exposing the module resolve configuration let's the Aurelia Plugin work properly.

Aurelia also uses `.html` for templating as the standard option, so added in a rule for `.html` if the Aurelia plugin is enabled.

Will work on an example shortly :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -647,11 +647,11 @@ const publicApi = {
      *
      * https://github.com/aurelia/webpack-plugin/wiki
      *
-     * @param {function} aureliaLoaderOptionsCallback
+     * @param {function} aureliaPluginOptionsCallback
      * @returns {exports}
      */
-    enableAureliaPlugin(aureliaLoaderOptionsCallback = () => {}) {
-        webpackConfig.enableAureliaPlugin(aureliaLoaderOptionsCallback);
+    enableAureliaPlugin(aureliaPluginOptionsCallback = () => {}) {
+        webpackConfig.enableAureliaPlugin(aureliaPluginOptionsCallback);
 
         return this;
     },
--- a/index.js
+++ b/index.js
@@ -630,6 +630,18 @@ const publicApi = {
     },
 
     /**
+     * If enabled, the Aurelia plugin is enabled
+     *
+     * @param {function} aureliaLoaderOptionsCallback
+     * @returns {exports}
+     */
+    enableAurelia(aureliaLoaderOptionsCallback = () => {}) {
+        webpackConfig.enableAurelia(aureliaLoaderOptionsCallback);
+
+        return this;
+    },
+
+    /**
      * If enabled, display build notifications using
      * webpack-notifier.
      *
@@ -701,6 +713,12 @@ const publicApi = {
      */
     configureFilenames(filenames) {
         webpackConfig.configureFilenames(filenames);
+
+        return this;
+    },
+
+    configureResolveModules(directories) {
+        webpackConfig.configureResolveModules(directories);
 
         return this;
     },
--- a/index.js
+++ b/index.js
@@ -632,11 +632,26 @@ const publicApi = {
     /**
      * If enabled, the Aurelia plugin is enabled
      *
+     * https://github.com/aurelia/webpack-plugin
+     *
+     * Encore.enableAureliaPlugin();
+     *
+     * or with configuration options:
+     *
+     * Encore.enableAureliaPlugin(function(options) {
+     *      // Set the startup module to hint webpack to module tracing
+     *      options.aureliaApp = "main";
+     * });
+     *
+     * Read about more configuration options that are available:
+     *
+     * https://github.com/aurelia/webpack-plugin/wiki
+     *
      * @param {function} aureliaLoaderOptionsCallback
      * @returns {exports}
      */
-    enableAurelia(aureliaLoaderOptionsCallback = () => {}) {
-        webpackConfig.enableAurelia(aureliaLoaderOptionsCallback);
+    enableAureliaPlugin(aureliaLoaderOptionsCallback = () => {}) {
+        webpackConfig.enableAureliaPlugin(aureliaLoaderOptionsCallback);
 
         return this;
     },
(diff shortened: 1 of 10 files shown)


index.js
```diff
@@ -629,6 +629,18 @@ const publicApi = {
         return this;
     },
 
+    /**
+     * If enabled, the Aurelia plugin is enabled
```
weaverryan: Can you add a link to Aurelia here? And what are some valid callback options. We usually like to show an example :)

index.js
```diff
@@ -629,6 +629,18 @@ const publicApi = {
...
+     *
+     * @param {function} aureliaLoaderOptionsCallback
+     * @returns {exports}
+     */
+    enableAurelia(aureliaLoaderOptionsCallback = () => {}) {
```
weaverryan: If it's a plugin, we try to include that in the name to help be transparent. So, `enableAureliaPlugin`

package.json
```diff
@@ -26,6 +26,8 @@
   },
   "homepage": "https://github.com/symfony/webpack-encore",
   "dependencies": {
+    "aurelia-loader-webpack": "^2.1.0",
```
weaverryan: Is this needed? Is this a loader or a plugin?

Also, these should be in `devDependencies`. The `lib/features.js` should cause the user to see a nice error that they need to install these (instead of us shipping Encore with them included)

index.js
```diff
@@ -705,6 +717,12 @@ const publicApi = {
         return this;
     },
 
+    configureResolveModules(directories) {
```
weaverryan: Is this related to Aurelia? I mean, I know it's a webpack feature, but is it needed? I'm not familiar at all with Aurelia, so you'll have to help us out :)

weaverryan (changes requested):

lib/plugins/aurelia.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {Array} plugins
+ * @param {WebpackConfig} webpackConfig
+ * @return {void}
+ */
+module.exports = function(plugins, webpackConfig) {
+    if (!webpackConfig.useAurelia) return;
+
+    loaderFeatures.ensurePackagesExist('aurelia-webpack-plugin');
+
+    const AureliaPlugin = require.resolve('aurelia-webpack-plugin');
```
Lyrkan: Shouldn't that be `require('aurelia-webpack-plugin')` since a `require.resolve()` call would return a string and not a constructor?

package.json
```diff
@@ -49,6 +49,7 @@
     "yargs": "^8.0.1"
   },
   "devDependencies": {
+    "aurelia-webpack-plugin": "^2.0.0-rc.5",
```
Lyrkan: I see that's a RC version, is it stable enough (I don't know Aurelia at all)?

Lyrkan (changes requested):
Hi @tomtomau,

Here are some additional comments.
Some tests should also be added since in its current state I don't think using it as documented would work (because of the callback and `require.resolve` issues).

lib/plugins/aurelia.js
Lyrkan: To be more precise it should probably be something like:
```js
const { AureliaPlugin } = require('aurelia-webpack-plugin');
```

or without the destructuring assignment:
```js
const AureliaPlugin = require('aurelia-webpack-plugin').AureliaPlugin;
```

tomtomau:
Thank you for the comments, I will make the changes and update the PR when
I have done so :)

index.js
tomtomau: Done :)

index.js
tomtomau: I'll give my best understanding, again, not super familiar with Webpack.

This is something that's required for Aurelia. Aurelia uses dynamic module loading a fair bit, which means that Webpack doesn't understand how to resolve those dynamic modules. This option then tells Webpack to try resolving via a particular path.

For example, my Webpack config code:

```
Encore
    // ...
    .configureResolveModules([
        './assets/app',
        'node_modules'
    ]);
```

You can read more about it on the aurelia/webpack-plugin [Wiki page](https://github.com/aurelia/webpack-plugin/wiki/Getting-started#whats-the-problem-with-aurelia-anyway).

lib/WebpackConfig.js
```diff
@@ -403,6 +405,11 @@ class WebpackConfig {
         this.vueLoaderOptionsCallback = vueLoaderOptionsCallback;
     }
 
+    enableAureliaPlugin(aureliaPluginConfig = {}) {
```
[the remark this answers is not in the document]
tomtomau: Yep - my bad!

lib/config-generator.js
```diff
@@ -89,6 +91,10 @@ class ConfigGenerator {
             config.resolve.alias['react-dom'] = 'preact-compat';
         }
 
+        if (this.webpackConfig.resolveModules && Array.isArray(this.webpackConfig.resolveModules)) {
```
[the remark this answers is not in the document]
tomtomau: Done

package.json
tomtomau: IMO, yes. Aurelia has a solid track record of releasing RC's that rarely have many breaking changes. 2.0 of this plugin has not been released but 1.x has been abandoned. They tend to run RC's for a while to be certain that their true x.0 release is reliable after being battle tested a bit :)

tomtomau: When 2.0 is released, I'll do my best to remember to come back and update this though :)

tomtomau:
Hi @weaverryan and @Lyrkan!

I have updated my pull request per the feedback including adding test cases for both the configureResolveModules call and the enableAureliaPlugin call.

Travis seems to be failing on one of the builds because the `aurelia-webpack-plugin` dependency is "not published" but I'm not 100% in understanding how this works.

I'm also drafting a blog post walking through how to get Aurelia + Encore setup.

Cheers!
Tom

Kocal:
Closing this as Webpack Encore is now in low-maintenance mode, accepting bug fixes only. An Aurelia plugin is a new feature, so it falls outside that scope. You put real effort into addressing the review feedback back in 2018 though, so thanks for that, @tomtomau.

PR #255, opened 2018-02-02
fixed some comments in order to be coherent with docs

stoccc:
I fixed a typo in a comment (`/web/build` is not a relative path).

I fixed some comments in order to be coherent with symfony docs and with the output produced when setManifestKeyPrefix is missing but necessary (`Error: Cannot determine how to prefix the keys in manifest.json. Call Encore.setManifestKeyPrefix() to choose what path (e.g. build/) to use when building your manifes  t keys. This is happening because you passed an absolute URL to setPublicPath().`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,7 +39,7 @@ const publicApi = {
     /**
      * The directory where your files should be output.
      *
-     * If relative (e.g. /web/build), it will be set relative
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks @stoccc!

PR #264, opened 2018-02-13
These changes allow empty public path in webpack.config.js

stoccc:
This allows empty publicPath in webpack.config.js
Provided a functional test.
See https://github.com/symfony/webpack-encore/issues/26#issuecomment-362560685

I'm new to webpack/encore/nodejs/git, so I hope I didn't make any mistake :)

### DIFF
--- a/test/functional.js
+++ b/test/functional.js
@@ -246,8 +246,8 @@ describe('Functional tests using webpack', function() {
                     ],
                     (browser) => {
                         webpackAssert.assertResourcesLoadedCorrectly(browser, [
-                            'http://127.0.0.1:8080/public/build/0.js',
-                            'http://127.0.0.1:8080/public/build/main.js'
+                            'http://127.0.0.1:8080/build/0.js',
+                            'http://127.0.0.1:8080/build/main.js'
                         ]);
 
                         done();
--- a/test/functional.js
+++ b/test/functional.js
@@ -240,8 +240,7 @@ describe('Functional tests using webpack', function() {
                 );
 
                 testSetup.requestTestPage(
-                    // the webroot will not include the public/build part
-                    path.join(config.getContext(), ''),
+                    path.join(config.getContext(), 'public'),
                     [
                         convertToManifestPath('build/main.js', config)
                     ],
--- a/test/functional.js
+++ b/test/functional.js
@@ -227,6 +227,36 @@ describe('Functional tests using webpack', function() {
             });
         });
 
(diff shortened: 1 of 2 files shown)


stoccc:
@weaverryan I think I'll need some help with the functional test...

stoccc:
It seems that we have some problems using javascript `require.ensure`: the file path is not translated to the client according to the information stored in manifest.json.
Is there any config that we can edit so that the reference is generated properly?

test/functional.js
```diff
@@ -227,6 +227,35 @@ describe('Functional tests using webpack', function() {
             });
         });
 
+        it('Deploying to an unknown (at compile-time) subdirectory is no problem', (done) => {
+            const config = createWebpackConfig('public/build', 'dev');
+            config.addEntry('main', './js/code_splitting');
```
weaverryan: I think we should also include an entry that loads a CSS file normally (i.e. without code splitting) that includes a path to an image or a font. Then, after running webpack, we can manually check the contents of the final CSS file to make sure that the path to that image is correct. As I understand it, THAT is the biggest problem we're having with relative paths: the paths to images/fonts in CSS files was being generated incorrectly.

Unfortunately, the `assertResourcesLoadedCorrectly` method you're using below works for JS only - that's why we need to test the contents of the file.

weaverryan (changes requested):
Hey @stoccc!

Thanks for starting this! Ok, about the `require.ensure` call... this is Webpack's responsibility to resolve this path correctly: it does not read from our `manifest.json`. file. Actually, it basically just uses the final `publicPath` config key to figure out the path to any code split files. So... this is a real problem: I'm not sure if relative paths *can* work with code-splitting. Since Webpack uses `publicPath` to find these files, if you use a relative path, then it will try to request a different file based on the URL of your page - e.g. http://example.com/foo versus http://example.com/foo/bar.

I found this with some quick research: https://stackoverflow.com/questions/36675286/webpack-code-splitting-relative-url

If this is correct, it confirms my believe.... because the solution they're suggesting is to override the publicPath at runtime... which is the only time we know the true path.

Can you research this a bit more? It may be that relative paths will simply *not* work with code splitting (unless the user does this workaround at runtime (more on this: https://webpack.js.org/guides/public-path/#on-the-fly).

Kocal:
Closing this one. It never quite got finished, the functional test was still open when the PR stalled back in 2018, and the codebase has changed a lot since then. Thanks for giving it a shot, @stoccc.

PR #265, opened 2018-02-15
Fix 'plugin' typehint on Encore.addPlugin

Blackskyliner:
The `plugin` parameter is described as `{string}` but in fact we want to have instances of a Webpack plugin here. So I think we should use `object` here, which reflects the fact that all examples and known usages use an object there...


weaverryan:
Thank you Christoph!

Issue #288, opened 2018-03-28
Detect usage of npm or yarn in the project when asking to install more packages


stof:
Currently, Encore always assumes `yarn` is used to add packages (with inconsistencies when the error does not come from Encore itself, see #57).

This causes confusion for users using npm instead of yarn (see #287).

When starting to work on Encore, `yarn` was considered superior to `npm`, because we compared it to version 3. With the version 5 of `npm`, the difference between both is smaller, so we should try to respect choices of the user.

Here is my proposal:

- if the project contains a `yarn.lock` file, display the `yarn` command to the user
- if the project contains a `packages.lock` file, display the `npm` command to the user
- otherwise, display the `yarn` command (following our existing rule where we favor yarn)

The last point may be changed in the future once https://github.com/geowarin/friendly-errors-webpack-plugin/issues/33 gets resolved, to fix #57 (if they display the `yarn` version only when having a `yarn.lock` file, we may want to do the same, as running Yarn without lock file is not recommended anyway)

PR #291, opened 2018-03-30
Recommend a different command for missing loaders based on the presence of yarn/npm lock files

### LINKED ISSUE #57: friendly-errors-webpack-plugin suggests `npm install`
The friendly-errors-webpack-plugin has a built-in transformer/formatter for when a module is missing. To see the behavior, try (in any JavaScript file):

```js
require('foo'); // foo is not a module that is installed
```

The error is:

> This dependency was not found:
> * foo in ./app/Resources/assets/js/app.js
> To install it, you can run: npm install --save foo

The problem is the `npm install` part... which should be `yarn add` to avoid confusion with our users. There is an issue about this: https://github.com/geowarin/friendly-errors-webpack-plugin/issues/33

Lyrkan:
This PR implements #288 by checking which kind of lock files are in the current directory before recommending the user a command to add missing loaders, for instance:

* No lock file: `yarn add <packages> --dev`
* yarn.lock: `yarn add <packages> --dev`
* package-lock.json: `npm install <packages> --save-dev`
* yarn.lock and packages-lock.json: `yarn add <packages> --dev`

Note that it doesn't fix #57 since that error message is currently handled by `friendly-errors-webpack-plugin`.


stof:
AFAICT, it is not possible to disable core formatters in the plugin. So I suggest contributing to the plugin instead (there is already an issue discussing this in the plugin repo, and this is actually where I took the idea of checking lock files to perform the detection, but no PR is opened yet)

PR #296, opened 2018-04-11
Add Encore.configureUrlLoader() method to the public API

### LINKED ISSUE #295: Add support of the url-loader
Please add support of the https://github.com/webpack-contrib/url-loader

Lyrkan:
This PR adds a `configureUrlLoader()` method to the public API in order to allow the use of the `url-loader` instead of the `file-loader` for images and fonts (closes #295).

Usage:

```js
// Enable the url-loader for both images and fonts
Encore.configureUrlLoader({
    images: {
        limit: 8192,
        mimetype: 'image/png'
    },
    fonts: {
        limit: 4096
    }
});

// Enable the url-loader for images and implicitly
// disable it for fonts (by omitting the "fonts" key)
Encore.configureUrlLoader({
    images: {
        limit: 8192,
        mimetype: 'image/png'
    }
});

// Enable the url-loader for fonts and explicitly
// disable it for images (by using a falsy value
// for the "images" key).
Encore.configureUrlLoader({
    images: false,
    fonts: {
        limit: 4096
    }
});
```


lib/config-generator.js
```diff
@@ -149,13 +149,23 @@ class ConfigGenerator {
...
+            // calling Encore.configureUrlLoader({ images: {/* ... */}})
+            let loaderName = 'file-loader';
+            let loaderOptions = {
+                name: filename,
+                publicPath: this.webpackConfig.getRealPublicPath()
+            };
+
+            if (this.webpackConfig.urlLoaderOptions.images) {
+                loaderName = 'url-loader';
+                loaderOptions = this.webpackConfig.urlLoaderOptions.images;
```
stof: the file-loader options must be combined with the url-loader options. Otherwise, we won't have the right config in case the url-loader fallbacks to the file-loader for bigger files.

package.json
```diff
@@ -43,6 +43,7 @@
     "pretty-error": "^2.1.1",
     "resolve-url-loader": "^2.0.2",
     "style-loader": "^0.13.2",
+    "url-loader": "^1.0.1",
```
stof: should it actually be a requirement for everyone using Encore ? This is an opt-in feature.

lib/config-generator.js
Lyrkan: Definitely, should be good now!

package.json
Lyrkan: Yep, I hesitated a bit on that one but ended-up adding it as a main dependency without any good reason.

I moved it to dev dependencies and added the usual `ensurePackageExist` checks.

Koc:
@Lyrkan thank you so much for this PR

weaverryan:
Thank you @Lyrkan! I've wanted this feature for awhile :)

Lyrkan:
I'll try to add some doc for it next week!

Lyrkan:
@javiereguiluz These options are [the ones provided by the URL loader](https://github.com/webpack-contrib/url-loader#options), not something we added on our side:

* The mimetype is optional and isn't a filter. You can use it if you want to force the mimetype in the generated data URL (`data:<mimetype>;base64,<data>`). By default the URL loader will look at the file extension to guess it.

* The size limit is per require call (for instance an `url(...)` in a CSS file).

PR #300, opened 2018-04-12
Allow to replace options in configuration callbacks

### LINKED ISSUE #298: Allow replacing options in configuration callbacks
Currently, the different configuration callbacks we expose are expecting to alter the options object directly.
But this makes it much harder to use libraries exposing a function processing options and *returning* new ones

#294 shows such a use case. https://github.com/symfony/webpack-encore/issues/294#issuecomment-378988530 is showing the current code needed to handle it, while https://github.com/symfony/webpack-encore/issues/294#issuecomment-378988999 shows how this could look like with this proposal.

The idea would be simple: inspect the return value of the callback when calling it:
- if it is an object, use it as the new options object
- if it is undefined (nothing was returned), keep using the existing options object (which may have been altered inside the callback, as done today).

Lyrkan:
This PR closes #298 by using the return value of callback functions (for loaders/plugins) as options if it is an object.

For instance: 

```js
Encore.configureUglifyJsPlugin((options) => {
    return {
        ...options,
        beautify: true
    };
});
```

### DIFF
--- a/lib/plugins/manifest.js
+++ b/lib/plugins/manifest.js
@@ -11,6 +11,7 @@
 
 const ManifestPlugin = require('../webpack/webpack-manifest-plugin');
 const PluginPriorities = require('./plugin-priorities');
+const applyOptionsCallback = require('../utils/apply-options-callback');
 
 /**
  * @param {Array} plugins
@@ -33,17 +34,10 @@ module.exports = function(plugins, webpackConfig) {
         writeToFileEmit: true,
     };
 
-    const callbackResult = webpackConfig.manifestPluginOptionsCallback.apply(
-        manifestPluginOptions,
-        [manifestPluginOptions]
-    );
-
-    if (callbackResult instanceof Object) {
-        manifestPluginOptions = callbackResult;
-    }
-
     plugins.push({
-        plugin: new ManifestPlugin(manifestPluginOptions),
+        plugin: new ManifestPlugin(
+            applyOptionsCallback(webpackConfig.manifestPluginOptionsCallback, manifestPluginOptions)
+        ),
         priority: PluginPriorities.WebpackManifestPlugin
     });
 };
--- a/lib/plugins/manifest.js
+++ b/lib/plugins/manifest.js
@@ -26,7 +26,7 @@ module.exports = function(plugins, webpackConfig) {
         manifestPrefix = webpackConfig.publicPath.replace(/^\//, '');
     }
 
-    let manifestPluginOptions = {
+    const manifestPluginOptions = {
         basePath: manifestPrefix,
         // guarantee the value uses the public path (or CDN public path)
         publicPath: webpackConfig.getRealPublicPath(),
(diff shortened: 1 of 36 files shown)


Lyrkan:
> I'm not sure. It makes it a lot longer, and most users don't need this.

Yeah, that's why I didn't do it, but I don't mind adding it if you think that's needed.
However in this case maybe we should also warn the user that returning a brand new object can remove some default options that are set by Encore.

> Another idea would be to start requiring returning as the ONLY way (and deprecate returning nothing). It would simplify the example, and it would be obvious for expert users that they could, of course, return a totally different/custom object.

I'm not sure about that, you'll still have to use the "options" param since in many cases it contains some values that are set by Encore. Doing `return { ...options, ignoreOrder: true }` (and that's if you use a Node version that supports object spread...) may be less easy to understand than just `options.ignoreOrder = true`.

stof:
I suggest creating a small utility instead of duplicating the same logic everywhere:

```js

const processOptions = require('./util/process_options');

const options =  { ... };

loaders.push({
    loader: 'postcss-loader',
    options: processOptions(webpackConfig.postCssLoaderOptionsCallback, options)
};
```

```js
module.exports = function (configurationCallback, options) {
    const result = configurationCallback.apply(options, [options]); // could even use .call to avoid creating a temporary array)

    if (typeof result === 'object') {
        return result;
    }

    return options;
}
```


> > Another idea would be to start requiring returning as the ONLY way (and deprecate returning nothing). It would simplify the example, and it would be obvious for expert users that they could, of course, return a totally different/custom object.
> 
> I'm not sure about that, you'll still have to use the "options" param since in many cases it contains some values that are set by Encore. Doing return `{ ...options, ignoreOrder: true }` (and that's if you use a Node version that supports object spread...) may be less easy to understand than just `options.ignoreOrder = true`.

Thus, it becomes even harder in the case of nested options.

stof:
The great thing is that we have the same signature for all our configuration callbacks (and centralizing the feature would ensure it stays true in the future). So the doc could have one section describing the different supported ways, and have all other sections showing one of them (with a link to the full doc about config callbacks).

stof:
btw, returning the options could still be done with simple editions:

```js
Encore.configureExtractTextPlugin((options) => {
    options.ignoreOrder = true;

    return options;
});
```

this could allow dropping support for not returning in the future (and throwing an exception if the callback does not return anything), but this would require a deprecation period first anyway.

Lyrkan:
@stof I changed it in order to use your method (+ tweaked the vue-loader part a bit).

You're also right about the fact that you could still do `options.ignoreOrder = true` followed by a `return options` but I still think that's a bit more error-prone. We'll throw an error if the method doesn't return anything but won't notice if the user inadvertently creates a new object that doesn't contain the default options.

lib/plugins/manifest.js
```diff
@@ -25,21 +26,18 @@ module.exports = function(plugins, webpackConfig) {
         manifestPrefix = webpackConfig.publicPath.replace(/^\//, '');
     }
 
-    const manifestPluginOptions = {
+    let manifestPluginOptions = {
```
stof: should be reverted to `const`

Lyrkan: Yep, I missed that one

weaverryan:
Thank you @Lyrkan!

PR #301, opened 2018-04-14
Added support for the handlebars-loader

ogiammetta:
First PR on GitHub, any feedback is appreciated! :)


lib/loaders/handlebars.js
```diff
@@ -0,0 +1,34 @@
...
+ * @return {Array} of loaders to use for Handlebars
+ */
+module.exports = {
+    getLoaders(webpackConfig) {
+        loaderFeatures.ensurePackagesExist('handlebars');
+
+        let options = {};
+
+        // allow the options to be mutated via a callback
+        webpackConfig.handlebarsConfigurationCallback(options);
```
Lyrkan:
```diff
const loaderFeatures = require('../features');
+ const applyOptionsCallback = require('../utils/apply-options-callback');

// (...)

-       let options = {};
+       const options = {}

-       // allow the options to be mutated via a callback
-       webpackConfig.handlebarsConfigurationCallback(options);
-
        return [
            {
                loader: 'handlebars-loader',
-               options: options
+               options: applyOptionsCallback(webpackConfig.handlebarsConfigurationCallback, options)
            }
        ];
```

test/loaders/handlebars.js
```diff
@@ -0,0 +1,37 @@
...
+        const config = createConfig();
+        config.enableHandlebarsLoader(function(options) {
+            options.debug = true;
+        });
+
+        const actualLoaders = handlebarsLoader.getLoaders(config);
+        expect(actualLoaders).to.have.lengthOf(1);
+        // callback is used
+        expect(actualLoaders[0].options.debug).to.be.true;
+    });
```
Lyrkan: Could you also add a test there to check if returning a value from the callback overrides the previous options correctly?

For instance:

```js
config.enableHandlebarsLoader(function(options) {
  options.debug = true;

  // This should override the original config
  return { foo: true };
});

// (...)

expect(actualLoaders[0].options.debug).to.be.undefined;
expect(actualLoaders[0].options.foo).to.be.true;
```

weaverryan:
Wow, great work Olivier! And congrats on your first GitHub PR. You're welcome here anytime ;)

PR #307, opened 2018-04-22
[WIP] Generator to init setup & generate JS entries

weaverryan:
Hi guys!

This is an ugly, WIP PR that that @Lyrkan has been kind enough to help me on. I'd like to create the PR now, so we can hopefully push this forward.

Here's the idea:

A) A new `encore init` command that you'd run once to help setup your `webpack.config.js` file. This asks basic questions like `outputDir`, `publicPath`, what type of JS (multi or SPA) you want, and what type of CSS you want (e.g. to `enableSassLoader()`). The benefits of this are to create the `webpack.config.js` (nice, but not a HUGE deal), download any dependencies the user needs, add a consistent `scripts` section to the user's `package.json` file (e.g. `yarn run encore:dev`), enable PostCSS (which requires a config file), put `/node_modules` in `.gitignore`, and give you a first entry file.

B) I'd also like to add a second command, which is already half-finished: `encore add:entry`. This would generate you either a new mini "JS" app in your project, which would either be "Vanillia" JS, React or Vue. For React & Vue, we would use their starter projects, so that it's SUPER easy to start using React or Vue, and your setup is consistent with their docs.

The TODO's are:

* [ ] Make the `assets` dir not hardcoded
* [ ] Add explanation message at start of `init`
* [ ] lint the generated webpack.config.js file in the test
* [ ] Add vanilla js support
* [ ] Add vue.js support
* [ ] Validate the already-working `encore init` command - make sure it's how we want it, finish small details (like the encore executable docs)
* [ ] Add/finish the `add:entry` command. This shouldn't be too much work. The only tricky part would be if we tried to magically update the user's `webpack.config.js`. That would be awesome... but if it's a blocker, we can just print a message.

Less important todos:
* [ ] Think about what support we want to give for React's test with Jest
* [ ] Check into React service worker


weaverryan:
Closed in favor of #318

PR #323, opened 2018-05-16
Update the description of the webpack runtime chunk

stof:
What gets extracted in this chunk is not the manifest (which creates confusion with the webpack manifest which is a JSON file used by the backend code, i.e. the Symfony asset component); it is the webpack runtime.

This runtime indeed includes some information also present in the manifest (it contains the mapping for all dynamic chunks), but that's not all of it.

### DIFF
--- a/lib/plugins/commons-chunks.js
+++ b/lib/plugins/commons-chunks.js
@@ -29,8 +29,8 @@ module.exports = function(plugins, webpackConfig) {
             name: [
                 webpackConfig.sharedCommonsEntryName,
                 /*
-                * Always dump a 2nd file - manifest.json that
-                * will contain the webpack manifest information.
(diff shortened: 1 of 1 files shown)


stof:
Next question is whether we should rename the output file from manifest.js to runtime.js (matching the default name used by Webpack 4 when enabling such feature).

This would reduce the confusion between manifest.js (the runtime file) and manifest.json (the manifest). But that would be a BC break (as projects have to load this runtime file in their HTML.

stof:
yeah, delaying this until the Webpack 4 migration might be fine.

weaverryan:
It's so confusing, I even typo'ed it in my notes :p. And yes, I think we should rename this - runtime.js is much more clear. It's already done on the Webpack 4 branch.

Issue #250, opened 2018-01-25
Upgrade to Webpack 4


thekonz:
As you might have already seen: [Webpack 4 beta is out](https://medium.com/webpack/webpack-4-beta-try-it-today-6b1d27d7d7e2)

This beta release is pretty much stable and will be released as such in a month.

> As promised, we will wait a month from today before releasing webpack 4 stable

The new release offers a lot simpler configuration and better compression. This should reduce the amount of code in encore and yield better results in production.

Lyrkan:
> This should reduce the amount of code in encore

That would be the case if we drop Webpack 2 and 3 compatibility... which is probably ok for new projects but would prevent existing projects from updating Encore easily.

Making it work with both Webpack 2, 3 and 4 may not be easy though based on some of the changes listed on this page: https://github.com/webpack/webpack/issues/6357

seyfer:
It is released today
https://medium.com/webpack/webpack-4-released-today-6cdb994702d4
waiting for the update :fireworks:

weaverryan:
Yep! We just need someone to create a PR :). Also, some libs we depend on may not yet be compat with 4, but we’ll find out with a PR. Volunteers? It’ll get done faster that way ;)

enguerranws:
@weaverryan What do you need in that PR? I started upgrading Webpack version and resolving module dependencies errors, ran `yarn test` (no errors in the output), but what can I do more to help on this?

Really looking forward to use Encore with Webpack 4 👍

andrewtch:
Same question here @weaverryan  - what do you expect in the PR?

enguerranws:
@andrewtch For the moment, I fixed extract-text-plugin breaking by fixing its version to `"extract-text-webpack-plugin": "4.0.0-beta.0"`.

Lyrkan:
Hi @enguerranws,

I'd say that if you have something that works with Webpack 4 you can already create a PR for it.
That'll help to see if something more is needed :)

I'm really interested to see how you handled some of the other things that changed in this version (such as the new `mode` option and the removal of `CommonsChunkPlugin` and `UglifyJsPlugin`).

Does your version still works with Webpack 3 by the way ? (even though I'm not sure we should keep supporting it, WDYT @weaverryan ?)

stof:
Well, the new `mode` option maps directly to our concept of `Encore.isProduction()`, so this is probably the easiest part of the migration.

stof:
[...]

PR #324, opened 2018-05-16
Hello Webpack4 ⭐️

weaverryan:
Hi guys!

I've delayed long enough. Let's upgrade to Webpack 4. As discussed on #250, we will not support old versions of Webpack - the versions are too different, and most people won't notice/care anyways.

And actually, other than the first todo, this is almost finished. AND, if you ignore `yarn.lock`, this PR makes Encore smaller :).

TODOS:
* [x] Fix `createdSharedEntry()`. I think we should just keep the existing functionality initially, and explore other options in the future. See https://github.com/symfony/webpack-encore/issues/250#issuecomment-389621788

* [x] Test `createdSharedEntry()` with CSS

* [x] Verify that default cache groups still exist (are not completely overwritten) and warning if you create a shared entry called vendor.

* [x] Investigate the Webpack 4 output, and see if we can remove any of our special error handling or display supression.

* [x] remove `vue-unactivated-loader`

* [x] We should now support `browserlist` in `package.json`. Need to check into this (add a test) and update documentation as needed.

* [x] Check into `fork-ts-checker-webpack-plugin@0.4.1" has unmet peer dependency "tslint@^4.0.0`

* [x] update BIG jsdoc notes in versioning.js related to NamedModulesPlugin

* [x] Make sure CSS is being minified

* [x] Look into exposing the entry point chunks in a new JSON file

* [x] Can CSS chunks ever be split with cacheGroups?

* [x] Make the `import()` syntax work out-of-the-box

* [x] Add some "version checking" logic to, at least, our vue-loader code. This new version of Encore will only work with vue-loader 0.15.0 and higher. But, because this is an optional dep, nothing enforces that and the user will get strange errors if they use a lower version. There is a precedent set to check versions of optional deps at runtime.

* [x] When a CSS file references a non-existent image, the error is way off:

```
 ERROR  Failed to compile with 1 errors                                         8:32:32 PM

 error  in ./css/another_bg_image.css

Module build failed: ModuleNotFoundError: Module not found: Error: Can't resolve './../images/symfony_logo.png2' in '/Users/weaverryan/Sites/os/webpack-encore/tmp_project_playing/css'
(stacktrace)
```

A) I thought we cleaned up the stacktrace
B) the ModuleNotFound is listed several times
C) The path just ends in `/css`, not the filename, which is not that helpful.
D) The actual error is hidden in the ugly stack trace, not on top.

* [x] Create a solution to use the `entrypoints.json` file & deprecate `createdSharedEntry()`

## Documentation

* [ ] Document use of browserslist + update Encore recipe

## Later

* [ ] (Later, in another PR) change from webpack-dev-server to webpack-serve


lib/config-generator.js
```diff
@@ -314,6 +306,24 @@ class ConfigGenerator {
             .map((plugin) => plugin.plugin);
     }
 
+    buildOptimizationConfig() {
+        const optimization = {
+
```
stof: this needs to have `runtimeChunk: 'single'` to extract the runtime (and the list of BC breaks should mention the renaming from `manifest.js` to `runtime.js`)

weaverryan:
Just re-added the traditional `createSharedEntry()` functionality. As you hinted Stof, it was fairly straightforward - see sha: 183b146783c5fbefb934eb69a733594f09e7a9c9

About the default groups - `vendor` and `default`. First, I'm not sure if I've overwritten them, or if they still exist. Either way, they don't appear to be a problem, because they are only async. I need to look further, however, because I don't actually want to override them - they should still be available for async chunks.

stof:
yeah, as you enable splitting of the initial chunks only for your own cache group and not globally for all of them (as I was doing in my example in #250), they are totally fine (splitting the async chunks as much as webpack wants is totally fine)

lib/WebpackConfig.js
```diff
@@ -308,6 +309,14 @@ class WebpackConfig {
         this.babelConfigurationCallback = callback;
     }
 
+    configureSplitChunks(callback) {
+        if (typeof callback !== 'function') {
+            throw new Error('Argument 1 to configureBabel() must be a callback function.');
```
bgaleotti:
```javascript
throw new Error('Argument 1 to configureSplitChunks() must be a callback function.');
```

weaverryan:
Ok! This is ready for review!

The file diff is huge, so I fear it's near impossible to review fully. However, the `CHANGELOG` has the highlights, including the BC breaks.

Anyone reading this, if you could try the `webpack4` branch of this repo in your project, and report any issues, that would be a huge help :). You can do that by updating your `package.json` to have:

```
"@symfony/webpack-encore": "git://github.com/symfony/webpack-encore.git#webpack4"
```

And running `yarn upgrade`.

raupie:
I was having issues with running 'production' for a foundation-sites 6 install with the latest stable version of webpack-encore. I added this branch and now it's passing successfully! This just made my day, seriously. Win! Love webpack-encore!

stof:
@weaverryan how does the entrypoints.json file works regarding versionning ? does it contain versionned names, or non-versionned ones (to lookup in the manifest.json file) ?

lib/config-generator.js
```diff
@@ -325,6 +321,49 @@ class ConfigGenerator {
...
+                test: this.webpackConfig.sharedCommonsEntryName,
+                // seems to default the rest of the options (like minSize)
+                // to settings so that modules matched by "test" are
+                // *definitely* included.
+                enforce: true,
+            };
+
+            splitChunks.cacheGroups = cacheGroups;
+            // causes a runtime.js to be emitted with the Webpack runtime
+            optimization.runtimeChunk = 'single';
```
stof: shouldn't this be done by checking whether `optimization.splitChunks` is a non-empty object instead ? Now that you added support for a callback, the case of using `sharedCommonsEntryName` is not the only case triggering splitting.

test/functional.js
```diff
@@ -377,11 +421,13 @@ describe('Functional tests using webpack', function() {
...
-                        'h1.c84caea6.css',
-                        'bg.483832e4.css',
-                        'manifest.json'
+                        '0.8256b1ad.js', // chunks are also versioned
+                        '0.8ec31654.css',
+                        'main.ba427376.js',
+                        'h1.8ec31654.css',
+                        'bg.0ec2735b.css',
+                        'manifest.json',
+                        'entrypoints.json'
```
stof: I suggest asserting the content of `entrypoints.json` too in this test using versionning

test/functional.js
```diff
@@ -94,6 +95,47 @@ describe('Functional tests using webpack', function() {
...
+            config.setPublicPath('/build');
+            config.configureSplitChunks((splitChunks) => {
+                splitChunks.chunks = 'all';
+                splitChunks.minSize = 0;
+            });
+
+            testSetup.runWebpack(config, (webpackAssert) => {
+                webpackAssert.assertOutputJsonFileMatches('entrypoints.json', {
+                    main: {
+                        js: ['vendors~main~other.js', 'main~other.js', 'main.js'],
```
stof: should we have the runtime file in this list too ?

lib/plugins/plugin-priorities.js
```diff
@@ -10,20 +10,19 @@
 'use strict';
 
 module.exports = {
-    ExtractTextWebpackPlugin: 0,
+    MiniCssExtractPlugin: 0,
     DeleteUnusedEntriesJSPlugin: 0,
+    EntryFilesManifestPlugin: 0,
```
stof: should have a low priority, to be sure it runs after other plugins (btw, we should really change these priorities. Having them all at 0 makes the priority system a simple `prepend/append` one, which could then be simplified a lot)

lib/features.js
```diff
@@ -18,75 +18,106 @@ const packageHelper = require('./package-helper');
...
-        packages: ['coffeescript', 'coffee-loader'],
-        description: 'process CoffeeScript files'
-    },
     forkedtypecheck: {
         method: 'enableForkedTypeScriptTypesChecking()',
-        packages: ['typescript', 'ts-loader', 'fork-ts-checker-webpack-plugin'],
+        packages: [
+            { name: 'typescript' },
+            { name: 'ts-loader', version: 4 },
+            { name: 'fork-ts-checker-webpack-plugin', version: 0 },
```
stof: this will ask to install `fork-ts-checker-webpack-plugin@^0.0`, which won't work.

You should not handle `version` as a number IMO, but as a semver version constraint instead IMO (use the `semver` package to check that then).

index.js
```diff
@@ -123,25 +123,6 @@ class Encore {
...
-     * For example:
-     *
-     *      Encore.configureExtractTextPlugin((options) => {
-     *          options.ignoreOrder = true;
-     *      })
-     *
-     * @param {function} extractTextPluginOptionsCallback
-     * @returns {Encore}
-     */
-    configureExtractTextPlugin(extractTextPluginOptionsCallback = () => {}) {
```
Lyrkan: Maybe we could also do something in the `EncoreProxy` at the bottom of that file to detect when a user tries to call `configureExtractTextPlugin` or `enableCoffeeScriptLoader` and display some kind of "This method is not available anymore in Webpack Encore" error message instead of the default one?

lib/config-generator.js
```diff
@@ -115,7 +118,7 @@ class ConfigGenerator {
     }
 
     buildOutputConfig() {
-        // Default filename can be overriden using Encore.configureFilenames({ js: '...' })
+        // Default filename can be overridden using Encore.configureFilenames({ js: '...' })
```
Lyrkan: There is the same typo twice below (my bad!)

weaverryan:
@stof

> @weaverryan how does the entrypoints.json file works regarding versionning ? does it contain versionned names, or non-versionned ones (to lookup in the manifest.json file) ?

Actually, this is tricky. I believe it is currently the versioned file. But, I think that's a mistake: it should be the unversioned filename because we need to be able to loop up the value `manifest.json` (e.g. in case a CDN is used). In fact, I think the filenames in `main.js` need to be the same as the "keys" in `manifest.json`.

// entrypoints.json
```json
{
    "main": {
        "js": ["build/runtime.js", "build/shared.js", "build/main.js"],
        "css": ["build/shared.css"]
    }
}
```

// manifest.json
```json
{
    "build/runtime.js": "https://somecdn.com/build/runtime.abc123.js",
    "...": "..."
}
```

This would allow some future code like this in a template:

```twig
{% for jsFile in getEntrypointJavaScriptFiles('main') %}
        <script src="{{ asset(jsFile) }}"></script>
{% endfor %}
```

But I'm not sure yet if this entirely makes sense. Another option would create a simpler map:

// entrypoints.json
```json
{
    "main": {
        "js": ["runtime.js", "shared.js", "main.js"],
        "css": ["shared.css"]
    }
}
```

Then, do a bit more work in the template:

```twig
{% for jsFile in getEntrypointJavaScriptFiles('main') %}
        <script src="{{ asset('build/'~jsFile) }}"></script>
{% endfor %}
```

(or some shortcut version)

```twig
{{ webpackEntryScriptTags('main', 'build/') }}
```

(or even put the `build/` prefix in the config of that future bundle.

Thinking out loud - not sure about this exact detail yet.

weaverryan:
Was just working on it today :). All major parts are done - just finishing touches. I’ll have it soon!

weaverryan:
And here's the new bundle: https://github.com/knpuniversity/webpack-encore-bundle

lib/config-generator.js
```diff
@@ -365,6 +366,25 @@ class ConfigGenerator {
...
+                    .createHash('md4')
+                    .update(name)
+                    .digest('hex')
+                    .slice(0, 8);
+            };
+
+            splitChunks.name = function(module, chunks, cacheGroup) {
+                const names = chunks.map(c => c.name);
+
+                return cacheGroup + '~' + hashFilename(names.join('~'));
```
Lyrkan: I'm not sure `~` is a good separator choice, it's part of the "unsafe characters" listed in this RFC: http://www.ietf.org/rfc/rfc1738.txt

stof: that's the separator used by default in Webpack 4

weaverryan:
Woohoo!

I've just merged this! But, I'm not ready to create a tag: I need to make a PR for the documentation and discuss the plan for the WebpackEncoreBundle. But, it should now be a bit easier to test Webpack4 - just use the master branch - update your `package.json` file:

```
"@symfony/webpack-encore": "git://github.com/symfony/webpack-encore.git#master"
```

arendjantetteroo:
Nice work!

I've tried updating and got 2 errors 
1. style-loader was missing, not sure if this was a previous requirement? (run yarn add style-loader fixed it)
2. An error about a plugin/preset doing some things not allowed:

$ yarn run encore dev --watch
yarn run v1.7.0
warning package.json: No license field
$ /node_modules/.bin/encore dev --watch
/node_modules/@babel/core/lib/config/config-descriptors.js:138
    throw new Error(`Plugin/Preset files are not allowed to export objects, only functions. In ${filepath}`);
    ^

Error: Plugin/Preset files are not allowed to export objects, only functions. In /node_modules/babel-preset-react/lib/index.js

We use enableReactPreset. 
Am i missing a change to make?

stof:
maybe you need to update the preset to a newer version

arendjantetteroo:
Thanks, that was indeed it.

raupie:
I'm getting this error: `SyntaxError: unexpected token: ':'` which points to line one below.

```
/***/ "../../../tmp/tmp-92SW7NIosJEwEF.tmp":
/*!***********************************!*\
  !*** /tmp/tmp-92SW7NIosJEwEF.tmp ***!
  \***********************************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {
```

weaverryan:
@arendjantetteroo What version of react-preset *were* you using before? And did you see any warning that you needed to upgrade? After this PR, if you're using an optional library that is outside of our supported version constraints, you should see a warning. I'm worried there may be some issue with that - would love some more info from you on #359. Thanks!

> style-loader was missing, not sure if this was a previous requirement? (run yarn add style-loader fixed it)

This is mysterious, because it IS a under `dependencies`, so it should have been installed automatically. Basically, this should work *without* needing to add it. @arendjantetteroo if you `yarn remove style-loader` do things still work? 

@raupie Can you post your setup on #360? You are using `createSharedEntry()`, right? It indeed looks like a possible bug - but I'll need help reproducing.

arendjantetteroo:
@weaverryan Sorry, holiday came in between. 

We had  "babel-preset-react": "^6.24.1", in yarn. 
No style-loader in package.json 

Upgraded to "@babel/preset-react": "^7.0.0-beta.54" fixed the error (but caused a lot of other dependency to be needed to upgrade as well)

Removing the style-loader it now seems to work fine, might have been an issue with all the dependency updates or leftover webpack config.

Issue #334, opened 2018-06-10
Disable URL handling


XWB:
The webpack css-loader has the ability to disable url handling by setting the following option to false:

https://github.com/webpack-contrib/css-loader#url

As far as I can see, Encore has no ability to pass this setting through Webpack. I would be nice to have this ability.

PR #335, opened 2018-06-11
Add option to configure the CSS loader

XWB:
Disabling the URL handler can be handy for importing legacy CSS.

Fixes https://github.com/symfony/webpack-encore/issues/334

### DIFF
--- a/index.js
+++ b/index.js
@@ -582,6 +582,25 @@ class Encore {
     }
 
     /**
+     * Configure the css-loader.
+     *
+     * https://github.com/webpack-contrib/css-loader#options
+     *
+     * Encore.configureCssLoader(function(config) {
+     *      // change the config
+     *      // config.minimize = true;
+     * });
+     *
+     * @param {function} callback
+     * @returns {Encore}
+     */
+    configureCssLoader(callback) {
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -774,6 +774,18 @@ class Encore {
...
+    /**
+     * Call this if you wish to disable the CSS
+     * url handler.
+     *
+     * @returns {Encore}
+     */
+    disableUrlHandling() {
+        webpackConfig.disableUrlHandling();
+
+        return this;
```
weaverryan: Hmm, I think a better idea would be to allow the `css-loader` to be configured in general. So, to create a method more like `configureBabel()`, where you pass it a callback, and we pass the callback the final config (which you can then modify). That'll allow you to disable the url handling... but also to modify any options on the `css-loader`.

XWB:
@weaverryan I've made some changes, can you check it out?

XWB:
@Lyrkan @weaverryan All done.

weaverryan:
Thanks @XWB!

XWB:
@weaverryan Any idea when you will tag a new release that contains this feature?

PR #339, opened 2018-06-22
A hacky attempt at supporting createSharedEntry() in Webpack 4

weaverryan:
There is currently on big problem with the Webpack 4 support (#324): `createSharedEntry()` no longer works as expected.

Suppose this config:

```js
.createSharedEntry('layout', './assets/layout');
```

Where `assets/layout.js` looks like this:

```js
// assets/layout.js

// jquery will be in the "shared" entry and not repeated in other entries
const $ = require('jquery');

console.log('this code will not run');
```

In Webpack 3, the `console.log()` code WOULD run. But, in Webpack 4, the final `layout.js` file simply contains the code for all of the modules... but unless someone actually *requires* them, their code is not executed. As a result, unless some other entry explicitly had `require('./assets/layout.js')`, the `console.log()` would never actually execute.

This, unfortunately, is a big change in behavior, and it no longer allows you to use `createSharedEntry()` to identify a "layout" file whose code you want included (and executed) on every page.

There are 2 solutions:

A) A hacky solution to support `createSharedEntry()`, at least temporarily (this PR!)

B) Stop supporting `createSharedEntry()`, and "force" people to use Webpack 4's new code splitting, where each entry may ultimately require multiple `<script>` tags. We will need a new bundle to support this behavior anyways. But, in this scenario, this would be the MAIN way to use webpack, unless you have an SPA and don't care about this shared stuff.

I would love feedback - this has been blocking #324, and keeping me up at night ;). I also may be missing something: the new split chunks functionality is very powerful, but complex.

### DIFF
--- a/fixtures/js/shared_example.js
+++ b/fixtures/js/shared_example.js
@@ -1,0 +1,5 @@
+// used in a createdSharedEntry() test
+require('./no_require');
+require('./requires_arrow_function');
+require('./../css/h1_style.css');
+require('./print_to_app');
--- a/index.js
+++ b/index.js
@@ -392,14 +392,14 @@ class Encore {
     }
 
     /**
-     * Add a "commons" file that holds JS shared by multiple chunks.
+     * Add a "commons" file that holds JS shared by multiple chunks/files.
      *
      * @param {string} name The chunk name (e.g. vendor to create a vendor.js)
-     * @param {string|Array}  files Array of files to put in the vendor entry
+     * @param {string} file A file whose code & imports should be put into the shared file.
      * @returns {Encore}
      */
-    createSharedEntry(name, files) {
-        webpackConfig.createSharedEntry(name, files);
+    createSharedEntry(name, file) {
+        webpackConfig.createSharedEntry(name, file);
 
         return this;
     }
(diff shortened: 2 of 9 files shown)


stof:
the webpack SplitChunksPlugin probably does not expect the common chunks to be usable as an entry themselves (i.e. something executed on load). Or it might be a bug on the webpack side if this is meant to work

stof:
Also, I would add a test loading both the shared entry and the non-shared entry, and ensuring that the code is not executed twice.

weaverryan:
> Also, I would add a test loading both the shared entry and the non-shared entry, and ensuring that the code is not executed twice.

Great idea!

> the webpack SplitChunksPlugin probably does not expect the common chunks to be usable as an entry themselves (i.e. something executed on load). Or it might be a bug on the webpack side if this is meant to work

It indeed appears to be removed as an entry on purpose (https://github.com/webpack/webpack/blob/4fc03e17d1a6f115b3b773ac0b663a73942d9d22/lib/optimize/SplitChunksPlugin.js#L652-L656) - removing that code makes things work. There are a few issues around this (https://github.com/webpack/webpack/issues/7230) but nobody seems to know for sure.

PR #356, opened 2018-07-20
Fix shared entry file when source maps are enabled

### LINKED ISSUE #352: Source Maps Failing in Dev
When using `.enableSourceMaps(!Encore.isProduction())` I am getting a `Uncaught SyntaxError: Unexpected token :` error.

When I set source maps to false, it goes away. 

Missing break line?

Lyrkan:
This PR fixes #352.

When source maps are enabled, the original shared entry file ends with a commented line.

This is an issue when the "_tmp_shared" entry is appended to it since its first line then becomes part of that comment.


weaverryan:
Great job @Lyrkan! Thank you!

PR #364, opened 2018-08-06
Add badge telling current npm version

TheMaxoor:
It can be really handy to rapidly check if an upcoming version has been released or not.


Lyrkan:
Thank you @TheMaxoor

PR #380, opened 2018-08-28
Add alias prod for production command

vince83110:
Very simple PR, everytime I write `yarn run encore prod` instead of `yarn run encore production`, I guess it's because we use _dev_ for _development_.

I kept the `production` command of course.

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -60,7 +60,7 @@ if (runtimeConfig.useDevServer) {
 }
 
 function showUsageInstructions() {
-    const validCommands = ['dev', 'production', 'dev-server'];
+    const validCommands = ['dev', 'prod', 'production', 'dev-server'];
 
     console.log(`usage ${chalk.green('encore')} [${ validCommands.map(command => chalk.green(command)).join('|') }]`);
     console.log();
--- a/lib/config/parse-runtime.js
+++ b/lib/config/parse-runtime.js
@@ -33,8 +33,8 @@ module.exports = function(argv, cwd) {
             runtimeConfig.environment = 'dev';
             runtimeConfig.verbose = true;
             break;
-        case 'production': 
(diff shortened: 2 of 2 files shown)


lib/config/parse-runtime.js
```diff
@@ -33,7 +33,7 @@ module.exports = function(argv, cwd) {
             runtimeConfig.environment = 'dev';
             runtimeConfig.verbose = true;
             break;
-        case 'production':
+        case 'production': case 'prod':
```
stof: please put the second case on a separate line instead

bin/encore.js
```diff
@@ -60,7 +60,7 @@ if (runtimeConfig.useDevServer) {
 }
 
 function showUsageInstructions() {
-    const validCommands = ['dev', 'production', 'dev-server'];
+    const validCommands = ['dev', 'prod', 'production', 'dev-server'];
```
royklutman: Is it an idea to also add `development` for consistency? So we could have dev, development, prod and production? Or is that unnecessary?

vince83110: In my case I would remove "production", I don't think it's useful to add longer commands.

royklutman: Indeed, I think we share the same thoughts. Although I'd like to see short commands only, I don't think removing one is a good idea in terms of backwards compatibility issues.

vince83110: What about change documentation with short names (to make people use the new aliases), and keep long names in code for compatibility ? I will do same PR for Symfony environment variable then.

royklutman: I think that's not a bad idea at all, we might need to ping in @stof and @weaverryan for this matter, maybe they can share their opinions.

weaverryan:
Thanks @vince83110! If someone wants a dev alias, you could open a PR

PR #389, opened 2018-09-21
[Dependencies] Update postcss-loader to use latest v3 release

royklutman:


stof:
@Lyrkan if we don't change it, we don't test either what happens for people using v3, and anyone doing a `yarn add postcss-loader` today would get v3, not v2.

Lyrkan:
Yep, as I said it probably isn't a big issue for that one if it is included in the Webpack 4 release since it already has other BC breaks, my change request was mostly motivated by the missing `yarn.lock` update.

That being said, I just realized that Encore probably won't allow older version with the new `ensurePackagesExistAndAreCorrectVersion` check anyway. So, the only remaining question is: in the future should we only allow to change dev dependencies versions during major updates since it may cause BC breaks?

royklutman:
@Lyrkan I added the `yarn.lock` file, can you have a quick look to verify it looks okay?

Lyrkan:
Seems right :)

Lyrkan:
Thank you @royklutman

PR #394, opened 2018-09-30
Filter --keep-public-path from arguments (fixes #180)

### LINKED ISSUE #180: Fix dev-server --keep-public-path Unknown argument
See: https://github.com/symfony/webpack-encore/pull/96#issuecomment-334865939

x539:
The webpack-dev-server dose not know the --keep-public-path argument
and fails with an 'Unknown Argument' error if it is aplied.

So before webpack-encore hands over controle to webpack-cli/-dev-server it has to
remove arguments that only exist for webpack-encore.

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -23,6 +23,10 @@ context.runtimeConfig = runtimeConfig;
 // remove the command from the output
 process.argv.splice(2, 1);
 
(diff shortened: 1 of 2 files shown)


Lyrkan:
Thank you @x539

PR #396, opened 2018-10-01
Babel upgrade.

danielnv18:
2 things in this PR
* `babel` has now an stable release
* For the npm scripts, You don't need to use the full path to the local binaries inside the node_modules. 
From [npm's doc](https://docs.npmjs.com/cli/run-script)
>  Any binaries provided by locally-installed dependencies can be used without the node_modules/.bin prefix

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "./node_modules/.bin/mocha --reporter spec test --recursive",
-    "lint": "./node_modules/.bin/eslint lib test index.js",
+    "test": "mocha --reporter spec test --recursive",
+    "lint": "eslint lib test index.js",
     "travis:lint": "npm run lint"
   },
   "bin": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,10 +25,10 @@
   },
   "homepage": "https://github.com/symfony/webpack-encore",
   "dependencies": {
-    "@babel/core": "^7.0.0-beta.49",
-    "@babel/plugin-syntax-dynamic-import": "^7.0.0-beta.49",
-    "@babel/preset-env": "^7.0.0-beta.49",
-    "babel-loader": "^8.0.0-beta.2",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
```
Johann-S: I think you can update your babel packages because the last release is `7.1.2`

Lyrkan: Let's merge it like that, it doesn't really matter since the caret also matches 7.1.x.

Lyrkan:
Thank you @danielnv18

PR #401, opened 2018-10-05
Allow to configure the JS exclude rules through configureBabel

### LINKED ISSUE #342: Configuration setting to run babel on vendor files
Currently, Encore forces to exclude `node_modules` and `bower_components` from the babel-loader. However, publishing ES2016 code is becoming more common, so this causes issues for some users (see #341 for the latest report related to this, but there are many others in the issue tracker).

I think we should make this behavior configurable.

Lyrkan:
This PR closes #342 by allowing to configure the js/jsx loaders' exclude rule by using `configureBabel()`.

For instance:

```js
Encore.configureBabel(
    () => {},
    { exclude: /foo/ }
);
```

Note that it doesn't change the default behavior that excludes `node_modules` and `bower_components`.

**Edit:**

Also adds an `include_node_modules` option to use the default `exclude` rule but only include some Node modules (can't be used if the `exclude` option is also set):

```js
Encore.configureBabel(
    () => {},
    { include_node_modules: ['foo', 'bar', 'baz'] }
);
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -639,13 +639,23 @@ class Encore {
      *
      * Encore.configureBabel(function(babelConfig) {
      *      // change the babelConfig
+     * }, {
+     *      // set optional Encore-specific options
+     *      // exclude: /(node_modules|bower_components)/
      * });
      *
+     * Supported options:
+     *      * {Condition} exclude (default=/(node_modules|bower_components)/)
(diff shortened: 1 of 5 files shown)


weaverryan:
Thank you @Lyrkan! Great work as usual!

PR #403, opened 2018-10-07
Add new Travis jobs to test lowest/highest versions of all the dependencies

### LINKED ISSUE #39: Figure out the way to test different deps than the locked ones
As this package is used as a library by others, we have no guarantee that they will install dependencies being exactly the same than our locked deps.
So it would be great to have some CI job resolving deps separately from the lock.

We need to figure out what the Node.js best practices around that are, and apply them on our CI.

Lyrkan:
This PR adds two Travis jobs that allow to test the lowest and highest versions of all the dependencies (including the dev ones since the version ranges are used by the package helper).

I wasn't sure about the method to use for the lowest versions since we can't really guess them without calling the npm's registry API. So, if someone has a better idea... :)

Note that both jobs currently fail for the following reasons:

- Lowest versions: This isn't really an issue since it seems related to Webpack 4 for which we haven't released a version yet. We could increase the minimum version to match the one from the `yarn.lock` file before releasing.
- Highest versions: This is caused by the last version of the `mini-css-extract-plugin` (0.4.3) which doesn't seem to work well with the `webpack-manifest-plugin` anymore (see https://github.com/webpack-contrib/mini-css-extract-plugin/pull/177#issuecomment-422571408).

Closes #39.

### DIFF
--- a/scripts/force-lowest-dependencies.js
+++ b/scripts/force-lowest-dependencies.js
@@ -20,6 +20,7 @@ function getLowestVersion(dependency, range) {
             (error, stdout) => {
                 if (error) {
                     reject(`Could not retrieve versions list for "${dependency}@${range}"`);
+                    return;
                 }
 
                 const versions = stdout
@@ -28,6 +29,7 @@ function getLowestVersion(dependency, range) {
 
                 if (versions.length === 0) {
                     reject(`Could not find a lowest version for "${dependency}@${range}"`);
+                    return;
                 }
 
                 const parts = versions[0].split(' ');
@@ -36,6 +38,7 @@ function getLowestVersion(dependency, range) {
                 // is directly printed as the output of npm view.
                 if (parts.length === 1) {
                     resolve([dependency, parts[0]]);
+                    return;
                 }
 
                 // If multiple versions are available then it outputs
@@ -43,6 +46,7 @@ function getLowestVersion(dependency, range) {
                 // <package>@<version> '<version>'
                 if (parts.length === 2) {
                     resolve([dependency, parts[1].replace(/'/g, '')]);
+                    return;
                 }
 
                 reject(`Unexpected response for "${dependency}@${range}": ${versions[0]}`);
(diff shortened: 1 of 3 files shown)


scripts/force-lowest-dependencies.js
```diff
@@ -0,0 +1,128 @@
+/*
```
BackEndTea: Would it be worth to put this into a separate npm package? This sounds like something that more developers would run into when maintaining packages.

scripts/force-lowest-dependencies.js
```diff
@@ -0,0 +1,123 @@
...
+                if (parts.length === 1) {
+                    resolve([dependency, parts[0]]);
+                }
+
+                // If multiple versions are available then it outputs
+                // multiple lines matching the following format:
+                // <package>@<version> '<version>'
+                if (parts.length === 2) {
+                    resolve([dependency, parts[1].replace(/'/g, '')]);
+                }
```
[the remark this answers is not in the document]
Lyrkan: Yeah, I wasn't really happy with that part... I guess calling the npm.js API directly would have been a cleaner way to do it but then we would also have had to filter the versions ourselves (and since it happens before `npm install`/`yarn`, we don't have access to the `semver` package at that point).

Thank you for highlighting that part by the way, I noticed that I forgot some `return` in promises... not a big issue since a promise won't be able to change its state after being resolved/rejected but still something that had to be fixed :)

scripts/force-lowest-dependencies.js
stof: as this is meant to run before a `npm install`, it might be harder :smile:

BackEndTea: Maybe something like a global install? Or a phar like equivalent for js

weaverryan: I'm always interested in re-using existing libs or extracting Encore into new libs. But, with limited resources, we need to stay focused on Encore itself (i.e. things like this would need to be done by some community member).

weaverryan:
Thanks @Lyrkan! This is a big step forward imo!

Issue #404, opened 2018-10-07
Incompatibility between mini-css-extract-plugin 0.4.3 and the manifest plugin


Lyrkan:
While working on #403 I noticed that the build was failing when upgrading all of our dependencies to their latest allowed version.

It seems that this is caused by an incompatibility between `mini-css-extract-plugin@0.4.3` and `webpack-manifest-plugin`: https://github.com/webpack-contrib/mini-css-extract-plugin/pull/177#issuecomment-422571408

I opened an issue on the Webpack Manifest Plugin repository (https://github.com/danethurber/webpack-manifest-plugin/issues/167), but meanwhile maybe we should force `mini-css-extract-plugin` to use v0.4.2?

Lyrkan:
Duplicate of #415

PR #406, opened 2018-10-10
Fix the mini-css-extract-plugin 0.4.3 issue and increase min. allowed Webpack version

Lyrkan:
This PR fixes both of the issues detected in #403:

- Set the maximum allowed version for `mini-css-extract-plugin` to 0.4.2 (since 0.4.3 doesn't work with the current version of the webpack-manifest-plugin, see #404). This is a temporary solution, once it's fixed (either in mini-css-extract-plugin or in weback-manifest-plugin) we'll have to change it again.

- Set the minimum allowed version for `webpack` to 4.20.0. The previous one (4.0.0) seemed to break `mini-css-extract-plugin` (see https://travis-ci.org/Lyrkan/webpack-encore/jobs/439810002).


Lyrkan:
By the way, once we update `mini-css-extract-plugin` to 0.4.4+ we'll also have to change some of the tests related to versioning due to [this commit](https://github.com/webpack-contrib/mini-css-extract-plugin/commit/d7946d0abb777a86a8e0b1a70245550a4f170aa1) (hashes not being the same anymore).

weaverryan:
Looks like a bit of a mess with mini-css-extract. 0.4.4 is now out. Does it fix the issue, or does the issue remain? Another way of asking this is: will the issue be fixed in mini-css, or is their opinion that this is not a bug, and it should be fixed elsewhere?

Lyrkan:
> Does it fix the issue, or does the issue remain?

That's still an issue when using 0.4.4: https://travis-ci.org/symfony/webpack-encore/jobs/439825872#L1055

The only different thing with 0.4.4 is that more tests are failing because it changes how some hashes are generated.

> will the issue be fixed in mini-css, or is their opinion that this is not a bug, and it should be fixed elsewhere?

Based on the last comments [here](https://github.com/webpack-contrib/mini-css-extract-plugin/pull/177#issuecomment-422571408) I'm not sure that there is an issue with `mini-css-extract-plugin`.

weaverryan:
Hmm ok. Obviously, this is far from ideal :). I created #415 to track this. My guess is that we're going to need to go into the manifest plugin ourselves and submit a PR to fix this - I noticed one +1 on it, but no feedback :/ 

Anyways, thanks! This fixes our immediate problem - which is great.

PR #409, opened 2018-10-13
Add a copyFiles() method to the public API (using require.context)

Lyrkan:
This PR is an alternative solution to #221 since that one is still blocked because of the `CopyWebpackPlugin`/`WebpackManifestPlugin` issue.

Instead of using an external plugin it relies on a fake entry that basically does multiple `require.context()` calls and forces them to use the `file-loader`.

The API is a bit different, and may be a bit harder to use since globs are replaced by RegExps.

```javascript
// Copy the content of a whole directory and its subdirectories
Encore.copyFiles({ from: './images' });

// Only copy files matching a given pattern
Encore.copyFiles({ from: './images', pattern: /\.(png|jpg|jpeg)$/ })

// Set the path the files are copied to
Encore.copyFiles({
	from: './images',
	pattern: /\.(png|jpg|jpeg)$/,
	to: 'assets/images/[path][name].[ext]'
})

// Version files
Encore.copyFiles(
	from: './images',
	to: 'assets/images/[path][name].[hash:8].[ext]'
})

// Add multiple configs in a single call
Encore.copyFiles([
	{ from: './images' },
	{ from: './txt', pattern: /\.txt$/ },
]);
```


lib/WebpackConfig.js
```diff
@@ -365,6 +366,39 @@ class WebpackConfig {
...
+                    throw new Error(`Invalid config option "${configKey}" passed to copyFiles(). Valid keys are ${Object.keys(defaultConfig).join(', ')}`);
+                }
+            }
+
+            this.copyFilesConfigs.push(
+                Object.assign({}, defaultConfig, config)
+            );
+        }
+    }
+
```
weaverryan: 👍

lib/plugins/delete-unused-entries.js
```diff
@@ -11,19 +11,22 @@
...
  * @param {Array} plugins
  * @param {WebpackConfig} webpackConfig
  * @return {void}
  */
 module.exports = function(plugins, webpackConfig) {
+    const entries = [... webpackConfig.styleEntries.keys()];
+
+    if (webpackConfig.copyFilesConfigs.length > 0) {
+        entries.push(copyEntryTmpName);
+    }
```
weaverryan: Not a super important detail, but I wonder if this will play nice with `entrypoints.json` - i.e. not have this fake entry there after #410. I suppose our "shared entry" already isn't a problem? So we'd be good here

index.js
```diff
@@ -404,6 +404,64 @@ class Encore {
...
+     *          { from: './images' },
+     *          { from: './txt', pattern: /\.txt$/ },
+     *      ]);
+     *
+     * Notes:
+     *      * No transformation is applied to the copied files (for instance
+     *        copying a CSS file won't minify it)
+     *      * By default files won't be versioned even if enableVersioning()
+     *        has been called. In order to add a hash to your filenames you
+     *        must use the "[hash]" placeholder in the "to" option (see below)
```
weaverryan: This part surprised me. Though, I can see now (I think) why this is the case: we're parsing everything through file-loader, and the user is in control of their filename. 2 questions:

1) Without ever writing any code, I was expecting the implementation of this would be basically to loop over and `require` each file that you want copied. In that case, they *would* be versioned automatically. But I guess, the problem is that you don't have control over the final filename / path?

2) Could we (or is it a bad idea) hack the `[hash]` in when it's versioned? Or, do this if `to` is not specified, but make it the responsibility of the user if it *is* specified?

test/functional.js
```diff
@@ -1326,6 +1326,227 @@ module.exports = {
...
+                            'symfony_logo.f27119c2.png',
+                        ]);
+
+                    webpackAssert.assertManifestPath(
+                        'build/main.js',
+                        '/build/main.js'
+                    );
+
+                    webpackAssert.assertManifestPath(
+                        'build/images/symfony_logo.png',
```
weaverryan: Ok, so to use this stuff, people would:

```
<img src="{{ path('build/images/symfony_logo.png') }}">
```

Right? They would basically "know" that the copy function puts things into the `build/` directory, so then they reference that path. Gosh, and the `manifest.json` is even smart enough to create the key *without* the `[hash]` part of the filename? That's incredible!

lib/plugins/delete-unused-entries.js
Lyrkan: Yep, I started the PR before #410 was created.
A check will have to be added [there](https://github.com/symfony/webpack-encore/pull/410/files#diff-cfa2bee523e3ea72d0447f6181fa3d5aR18) to exclude the temp file.

test/functional.js
Lyrkan: Yep, it's not different than doing a `require(...)` inside of a real entry after all :)

It doesn't really remove the `[hash]` part but simply uses the original name of the file which is something that wasn't possible to do with the `CopyWebpackPlugin` since it didn't expose it.

index.js
Lyrkan: Actually the issue comes from the way the different calls to `Encore.copyFiles` are currently saved.

If you look at the `copyFiles` function in `WebpackConfig.js` you'll see that the `to` property of the `defaultConfig` object contains the default pattern (`[path][name].[ext]`). Unless the user explicitly set `to` when adding a config it'll be this pattern that will be used when generating the related `file-loader` rule.

Since we don't know at that point if the versioning is enabled  it's not possible to replace it by a versioned pattern.

That being said, if you want it to works that way I can probably do it quite easily by setting the `to` option to `null` in the default config. And then choose which pattern should be used only when generating the `require` calls.

Lyrkan:
@weaverryan There you go, it should now work as expected when the `to` option isn't set and versioning is enabled :)

weaverryan:
Thank you @Lyrkan! This is awesome! All this work to update to Webpack 4... and I'm pretty sure this will be people's favorite feature of the next release :p

piotr-cz:
the `copyFiles` expression lost ability to use globs in `from` paths as described in #582

Lyrkan:
@piotr-cz Not sure what you mean there, `copyFiles(...)` never accepted globs in `from` (unless you're talking about #221 which was not merged).

As explained in #582 you should be able to achieve more or less the same thing using `pattern` and `includeSubdirectories`.

piotr-cz:
Thanks @Lyrkan

Issue #408, opened 2018-10-12
Use ztoben/assets-webpack-plugin instead of creating our own entrypoints.json file


weaverryan:
In this spirit of re-using outside libs :).

https://github.com/ztoben/assets-webpack-plugin/pull/140

PR #410, opened 2018-10-14
Implement assets webpack plugin

codayblue:
This should solve issue #408 because it implements the plugin that was requested to be looked into. 

The changes that have happened by implementing this plugin is that the keys might not be in the file for a given entry. So lets say we have a main entry point that has JavaScript but no CSS, the CSS key will not be under that entry point. The other issue is that when a file type only has one file of a given file type it will not be in an array in the entrypoints.json file it will just be a string value linked to the property.


lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,41 @@
 
 'use strict';
 
-const EntryFilesManifestPlugin = require('../webpack/entry-files-manifest-plugin');
 const PluginPriorities = require('./plugin-priorities');
-const path = require('path');
-const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
-const manifestKeyPrefixHelper = require('../utils/manifest-key-prefix-helper');
```
Lyrkan: It seems a bit strange to remove that require (and therefore not use that helper anymore).
Does it mean that the keys won't be prefixed the same way than before?

.eslintrc.js
```diff
@@ -52,7 +52,11 @@ module.exports = {
             }
         ],
         "node/no-unpublished-bin": "error",
-        "node/no-unpublished-require": "error",
+        "node/no-unpublished-require": ["error", {
```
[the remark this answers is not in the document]
codayblue: Where should I look? This was failing on my local and in Travis. When I did some looking this was the suggested answer from google. I dug a little more but found no other path. Just wondering what suggestions you may have.

test/functional.js
```diff
@@ -103,17 +103,15 @@ describe('Functional tests using webpack', function() {
...
                     main: {
-                        js: ['build/runtime.js', 'build/main.js'],
-                        css: []
+                        js: ['build/runtime.js', 'build/main.js']
                     },
                     font: {
-                        // no runtime for style entries
-                        js: [],
-                        css: ['build/font.css']
+                        js: 'build/runtime.js',
```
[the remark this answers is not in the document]
codayblue: I was not sure how much we wanted to alter the what the plugin was generating. I will make it so we filter out runtime if it is alone.

.eslintrc.js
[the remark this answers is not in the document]
Lyrkan: That's because you added `assets-webpack-plugin` as a dev dependency.

Since people installing Encore in their project won't get our dev dependencies the linter is telling you that they also won't have the `assets-webpack-plugin`.

The dev dependencies in Encore are used for optional things only (for instance if you use it with vue.js you'll have to add the loader manually), in this case we check that the dependency is present before requiring it (see `notifier.js` for instance).

For this plugin it makes more sense to add it to normal dependencies, so people will automatically get it when they install Encore.

lib/plugins/entry-files-manifest.js
codayblue: They come out already prefixed based on the webpack settings. I will dig more into seeing if this plugin that was picked out will allow us to pass in another prefix.

.eslintrc.js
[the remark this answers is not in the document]
codayblue: Okay I will move it. Thank you!

lib/plugins/entry-files-manifest.js
codayblue: This comment makes it look like I removed sharedEntryTmpName. I have put it back after realizing I had removed a module that should not have been removed and it is in the current commit.

test/functional.js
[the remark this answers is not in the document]
Lyrkan: Yeah, altering the output of the plugin is definitely a bit annoying and should be avoided as much as possible but I'm afraid that having `runtime.js` there would also cause the Webpack Encore Bundle to add it to the page multiple times.

Maybe wait for @weaverryan's to come by before making that change (same thing for the public path thing from another comment of mine).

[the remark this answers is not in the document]
weaverryan: I think I'm ok with it. If you add a style entry (which you shouldn't anyways - it's a "hack" feature that's meant to help with legacy projects), then you just need to be smart enough not to *try* to render the `script` tags for that entry. If you do, yep, you may get an extra script tag for `runtime.js`. But no big deal :)

lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,42 @@
...
 
-const EntryFilesManifestPlugin = require('../webpack/entry-files-manifest-plugin');
 const PluginPriorities = require('./plugin-priorities');
-const path = require('path');
 const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
-const manifestKeyPrefixHelper = require('../utils/manifest-key-prefix-helper');
+const AssetsPlugin = require('assets-webpack-plugin');
+
+function cleanFilePaths(assets) {
+    const tmpAssets = assets;
```
stof: this does not create a copy of the data structure. It only create a separate reference to the same object. So AFAICT, this is useless here, and you could as well use `assets` all the time (the other solution is that it is broken)

lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,42 @@
...
 'use strict';
 
-const EntryFilesManifestPlugin = require('../webpack/entry-files-manifest-plugin');
 const PluginPriorities = require('./plugin-priorities');
-const path = require('path');
 const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
-const manifestKeyPrefixHelper = require('../utils/manifest-key-prefix-helper');
+const AssetsPlugin = require('assets-webpack-plugin');
+
+function cleanFilePaths(assets) {
```
stof: the function name looks weird, as it is not only cleaning paths, but also stringifying the output to JSON.

test/functional.js
```diff
@@ -1362,7 +1360,7 @@ module.exports = {
                 const config = createWebpackConfig('web/build', 'dev');
                 config.addEntry('main', ['./css/roboto_font.css', './js/no_require', 'vue']);
                 config.addEntry('other', ['./css/roboto_font.css', 'vue']);
-                config.setPublicPath('http://localhost:8080/build');
+                config.setPublicPath('http://localhost:8080/custom_prefix');
```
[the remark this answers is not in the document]
weaverryan: Hmm. This is a tough one. In my original version, I apparently designed it (I say "apparently" because this is what the test tells us, but it was something I was "debating" by myself for a bit) so that when the entrypoints.json file is dumped, if the user is using an absolute publicPath (with http://), that is NOT included in the entrypoints values. Instead, the user would use the value from `entrypoints.json` to make a second lookup in `manifest.json`.

I'm not sure that was the best idea. And so, the NEW implementation is better. Specifically: it looks like the new implementation always makes the value in `entrypoints.json` the final value - so there is no `manifest.json` lookup necessary. Do you guys see any problems with this? It would mean that nobody could use `entrypoint.json` to *simply* get the "local" paths of the needed CSS or JS files - it would always be the absolute path (with the `http://` if they have one set). I'm trying to think if that's a problem...

Anyways, if we assume that the new implementation is better, then there is just one change to make here: change the publicPath BACK to `http://localhost:8080/build` because it's useful to set this path to something different than the manifestKeyPrefix (it's just a better test). Then, update the test below to look for `http://localhost:8080/build` in the values. Also, the description of this test would need to be updated, or 2 tests would need to be created. The new description would be something like:

> Custom public path is used in entrypoints.json, but does not affect manifest.json

lib/plugins/entry-files-manifest.js
Lyrkan: Maybe `processOutput(assets)` instead?
It's a bit vague but it would match the name of the related option of the `assets-webpack-plugin`.

lib/plugins/entry-files-manifest.js
Lyrkan: Maybe something like this instead? (I didn't test it so there may be some errors)

```js
const processedAssets = assets
    .filter((asset, name) => name !== sharedEntryTmpName);
    .map(asset => {
        return asset.map(typeAssets => {
            if (!Array.isArray(typeAssets )) {
                typeAssets = [typeAssets];
            }

            return typeAssets.map(buildPath => buildPath.replace(/^\//g, '');
        });
    });

return JSON.stringify(processedAssets);
```

It would also put everything in an array like the previous code, even if there is only one file.

codayblue: I tried to do the fancy snippet you had but it came out with filter is not a function. assets is a json object when it comes in. since I am doing dynamic keys that is why it looks like I am using an array when it is a json object. I checked the object prototype and I dont see any really helpful methods. I can try implementing them my self in some helper module. I dont know how fancy we want to go with it.

test/functional.js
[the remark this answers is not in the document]
codayblue: @weaverryan I updated the test to what I think you meant? not quite sure. Let me know if it needs more changes.

lib/plugins/entry-files-manifest.js
Lyrkan: Yeah, nevermind, you'd have to filter assets.keys() first and then rebuild the object... not sure that's worth it if directly mutating `assets` isn't an issue.

lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,41 @@
...
 const PluginPriorities = require('./plugin-priorities');
-const path = require('path');
 const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
-const manifestKeyPrefixHelper = require('../utils/manifest-key-prefix-helper');
+const AssetsPlugin = require('assets-webpack-plugin');
+
+function processOutput(assets) {
+    for (const asset in assets) {
+        if (sharedEntryTmpName === asset) {
+            delete assets[asset];
```
Lyrkan: Couldn't this be an unconditionnal `delete assets[sharedEntryTmpName];` placed outside of the loop instead?

test/functional.js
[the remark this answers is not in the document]
Lyrkan: @weaverryan It would also execute the content of that file multiple times, couldn't that be an issue?

lib/plugins/entry-files-manifest.js
stof: I think mutating it is fine if things work. But then, we should not create a `tmpAssets` variable, which make people reading the code *think* we have 2 separate data structures (while we only have a second name for the same one)

test/functional.js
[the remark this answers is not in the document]
codayblue: I have been thinking about this issue. I think the bundle could do something like array_merge and it would dedup the array of the same file. The reason I think it should be done in the bundle is because what if the user only requests one entry point and it was one of the ones that does not have runtime.js. They would not get all the files needed. Although in this case I see it only popped up because we pulled in css. So in that case we could then still do it here if we check that js count is 1 and is the runtime.js file and css count 1 or more then remove it. I will do which ever approach we want to take.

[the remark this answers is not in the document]
Lyrkan: Hm, I'm not sure your example would be an issue.
Removing `runtime.js` would only be done when it's the only element for the `js` key, which would also mean that it's not needed for the related entry anyway (no other JS file).

That being said, maybe you're right and it could be handled by the bundle instead...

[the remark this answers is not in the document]
codayblue: Ah I see your point now a little more clearly. Yeah I will filter it out later today and get this PR updated asap. So we can get this merged and running.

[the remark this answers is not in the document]
weaverryan: Just to chime in here: the bundle already prevents the same file from being rendered on the same page load. So, if you have an `app` entry that (of course) includes `runtime.js` and you also render another entry - `checkout` - there will be no `script` tag for `runtime.js` for checkout. This is needed in general, because when files are split, they may be "required" for multiple entry points, but we don't want to render multiple, duplicate `script` tags on the same page.

So, `runtime.js` will always be included 0 or 1 times. And, most likely, for a style entry, it will already be included in the page due to some other entry. And even if it is *not* (meaning, there is literally no JS on this page). it's harmless: just contains bootstrap code.

Does that alleviate concerns @Lyrkan?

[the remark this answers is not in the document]
Lyrkan: It does :)

codayblue:
@Lyrkan Is there anything else that needs to be taken care of in this PR?

Lyrkan:
@codayblue I think everything is fine now :)

codayblue:
Is there an way to have appveyor run again? I dont understand why it failed.

Lyrkan:
Don't worry about AppVeyor, it happens to fail for no real reason sometimes (mostly because of timeouts during functional tests), which seems to be the case here.

lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,40 @@
...
-const EntryFilesManifestPlugin = require('../webpack/entry-files-manifest-plugin');
 const PluginPriorities = require('./plugin-priorities');
-const path = require('path');
 const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
-const manifestKeyPrefixHelper = require('../utils/manifest-key-prefix-helper');
+const AssetsPlugin = require('assets-webpack-plugin');
+
+function processOutput(assets) {
+    if (assets.hasOwnProperty(sharedEntryTmpName)) {
+        delete assets[sharedEntryTmpName];
```
weaverryan: We now need to delete one other "magic: entry that may exist :) https://github.com/symfony/webpack-encore/pull/409#discussion_r226689309

lib/plugins/entry-files-manifest.js
```diff
@@ -9,26 +9,40 @@
...
+    }
+    for (const asset in assets) {
```
weaverryan: Can we add a little comment about what this is doing / why?

codayblue: true I never thought of that. Yes we could turn it into an array. I will add that to this PR.

lib/plugins/entry-files-manifest.js
codayblue: I will get this done.

codayblue: @weaverryan Is the any other files we need to delete from the output?

lib/plugins/entry-files-manifest.js
codayblue: @weaverryan Is there anything else we need to do to the output?

lib/plugins/entry-files-manifest.js
```diff
@@ -9,27 +9,46 @@
...
-            webpackConfig.styleEntries
-        ),
+        plugin: new AssetsPlugin({
+            path: webpackConfig.outputPath,
+            filename: 'entrypoints.json',
+            includeAllFileTypes: true,
+            entrypoints: true,
+            processOutput: processOutput
+        }),
         priority: PluginPriorities.EntryFilesManifestPlugin
```
weaverryan: Can you change this to be a new AssetsPlugin?

weaverryan:
Thank you Cody! And welcome to being an Encore contributor :D

Issue #418, opened 2018-10-23
Include JavaScript file hashes in entrypoints.json for hash & nonce attributes (CSP)


weaverryan:
Hi guys!

This is an idea that I've just been talking to a few people about: in the `entrypoints.json`, what if we output the file hashes for all of the JavaScript files? For example:

```json
{
  "app": {
    "js": [
      "build/runtime.js",
      "build/vendors~a33029de.js",
      "build/app.js"
    ],
    "css": [
      "build/vendors~ea1103fa.css",
      "build/app.css"
    ]
  },
  "checkout": {
    "js": [
      "build/runtime.js",
      "build/vendors~a33029de.js",
      "build/checkout.js"
    ],
    "css": [
      "build/checkout.css"
    ]
  },
  "_hashes": {
    "build/runtime.js": "oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC",
    "build/vendors~a33029de.js": "oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC",
    "build/app.js": "oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC",
    "build/checkout.js: "oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC"
  }
}
```

I am far from a security expert, but as I understand it, this would give us 2 possible security-related possibilities - especially when used with the new Twig function in WebpackEncore bundle that outputs your `script` tags automatically (https://github.com/symfony/webpack-encore-bundle):

1) A `integrity=` attribute for Subresource integrity (https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)

2) A `nonce=` attribute (set to the same hash), which would allow you to use a CSP that requires nonces for *all* script tags (a very strict policy). The server could also read these hash values for the CSP header.

Questions:

A) Good idea? Bad idea?
B) Is the `integrity=` something that's valuable for "self" script/link tags?

stof:
`nonce` should not use a hash of the file as the value (as the same value would be used multiple times).

frankdejonge:
[...]

PR #424, opened 2018-10-30
Put current data under an "entrypoints" key in the entrypoints.json file

Lyrkan:
Following my comment in #418 this PR adds an `entrypoints` key in the `entrypoints.json` file instead of putting everything at the root. It also simplifies a bit the code of the `processOutput` function to avoid duplicated code.

For now it won't really matter (since it'll be the only key) but it will avoid using "magic" keys later on if we want to add additional data (such as the hashes of the referenced files).

I'll also open a PR in the `webpack-encore-bundle` repository.


weaverryan:
Thank you @Lyrkan!

PR #440, opened 2018-11-09
updating css-loader dep and ts-loader dev dep

### LINKED ISSUE #431: Any way to use ts-loader ^5.1.1 or newer?
It's nice to be able to periodically run `yarn upgrade --latest` and have your package.json update to the latest versions of everything, but ts-loader on encore 0.21 seems to be locked at ^4.3.0 and instead I need to manually type out every dependency to update minus ts-loader to avoid getting the newer version. This was also an issue with encore 0.20 which was locked into an even older version of ts-loader and I was surprised to see the same issue on the latest encore.

Is it possible to bump the dependency requirement to ^5.1.1 which is the oldest 5.x version that contains a bug fix?

### LINKED ISSUE #432: CSS files not minimized in production build?
Trying to figure out why our CSS files are not minimized when use `encore production`. JS files are minimized without any problems but CSS files not, any ideas why?

### LINKED ISSUE #438: Upgrade css-loader to 1.0
The version of css-loader used by encore currently relies on a buggy version of cssnano to optimize the CSS. Upgrading to the new version should be done: https://github.com/symfony/webpack-encore/issues/436#issuecomment-437263637

weaverryan:
Fixes #438 and fixes #431 and I believe fixes #432 

This PR is more interesting than I expected! The latest version of `css-loader` (1.0) does not have the `minimize` option. Even though (until now) the older version was locked on our `package.json`, it may have been possible for users to be using a new version (I believe that should not be possible, but node deps can do strange things).


package.json
```diff
@@ -82,7 +82,7 @@
     "sinon": "^2.3.4",
     "stylus": "^0.54.5",
     "stylus-loader": "^3.0.2",
-    "ts-loader": "^4.3.0",
+    "ts-loader": "^5.3.0",
```
weaverryan: This is in `devDependencies`, so it really just means that we *test* against this version - the user can install something lower, but they will get a warning. We could also set this to `^5.0.0`, which would mean we would test against `5.0.0` in one our test suites. When the user needs to install this, they would be give the command `yarn add ts-loader ^5.0.0`, but that should install the latest version anyways.

weaverryan:
This has one minor failing test, but is otherwise ready for review

lib/webpack/delete-unused-entries-js-plugin.js
```diff
@@ -22,15 +22,17 @@ DeleteUnusedEntriesJSPlugin.prototype.apply = function(compiler) {
...
-                chunk.files.forEach((filename) => {
+                // loop in reverse, to avoid issues as we remove items from chunk.files
+                for (let i = chunk.files.length - 1; i >= 0; --i) {
+                    let filename = chunk.files[i];
                     if (/\.js(\.map)?(\?[^.]*)?$/.test(filename)) {
                         fileDeleteCount++;
                         // remove the output file
                         delete compilation.assets[filename];
                         // remove the file, so that it does not dump in the manifest
                         chunk.files.splice(chunk.files.indexOf(filename), 1);
```
stof: `chunk.files.indexOf(filename)` could be `i` instead

Issue #429, opened 2018-11-06
Add support for preload/prefetch and Symfony's WebLink component


weaverryan:
The WebLink component allows you to preload assets - https://symfony.com/doc/current/weblink.html

In the Webpack world, this is ability can be written into your Webpack config via https://github.com/GoogleChromeLabs/preload-webpack-plugin

We should be able to bring this nice functionality to Encore without too much trouble :)

1) Allow the user to configure which entrypoints should be preloaded. We should consider what the preload-webpack-plugin does for this. However, it may also make sense to add this as an option to `addEntry()`. This is something to think about.

2) Based on the above config, we would likely render new metadata in `entrypoints.json` about this.

3) WebpackEncoreBundle could then read this metadata and communicate with the WebLink component so that the proper headers are set.

Symfony already supports these new HTTP features. Let's make them easy to use in Encore!

nicolas-grekas:
How does caching come into play?
E.g. should we push files that are shared on all pages? since they would be cached most of the time, this may cause extra traffic by forcing the browser to send a frame discarding the push?
Should we only push the page-specific assets?
@dunglas what's your experience here?

dunglas:
> E.g. should we push files that are shared on all pages? since they would be cached most of the time, this may cause extra traffic by forcing the browser to send a frame discarding the push?

It's usually not a problem because browsers can cancel a push they already have the data in cache. However some browsers have (had?) bugs, and may not cancel the push, or not leverage it (especially Safari IIRC).

There was a very interesting article about this topic, but it seems to be offline right now: https://www.shimmercat.com/en/blog/articles/whats-push/

dunglas:
This article is a bit outdated, but provides a nice overview of how the different cache work together: https://jakearchibald.com/2017/h2-push-tougher-than-i-thought/

rakelley:
Without this feature `splitEntryChunks` is a mixed blessing for us as well.  Our existing shared entry setup utilizes `defer`ed script tags with preloading to ensure shortest time to interactivity.

With split chunks and `encore_entry_script_tags`'s non-deferred non-preloaded tags, the total build size is ~5% smaller, time to `DOMContentLoaded` is slightly faster on slow connections and slightly slower on fast connections, but full load always takes much longer.

stof:
@rakelley `encore_entry_script_tags` renders non-deferred tags, but you can use `encore_entry_js_files` instead and render the script tags yourselves, giving you full flexibility over making them deferred

[...]

PR #455, opened 2018-11-16
[WIP] Adding support for preload/prefetch & async/defer

weaverryan:
Hi guys!

This is start to supporting preload/prefetch out of the box - #429. I'd like feedback based on real-world requirements.

The usage is entirely centered around each *entry*:

```js
.addEntry('entry1', './js/entry1', {
    preload: true,
    async: true,
});
.addEntry('entry2', './js/entry2', {
    prefetch: true,
    defer: true,
});
```

This will output a new `metadata` key in `entrypoints.json`:

```json
{
    "entrypoints": { "all the normal stuff" },
    "metadata": {
        "entrypoints": {
            "entry1": {
                "hint": "preload",
                "attributes": {
                    "async": true
                }
            },
            "entry2": {
                "hint": "prefetch",
                "attributes": {
                    "defer": true
                }
            },
        }
    }
}
```

Then, WebpackEncoreBundle would be updated to read this new metadata, and output the correct attributes or communicate with the WebLink component to add the prefetch/preload headers.

## Features / Limitations

This initial attempt centers everything around the "entry" file. If you use [splitChunks()](https://symfony.com/doc/current/frontend/encore/split-chunks.html), the `entry1` file may actually render, for example, 3 script tags. And so, all 3 script tags would have the `async` attribute and have a `preload` header set. There is no file-by-file control. The question is: is this the best approach? Or, should this be entirely *filename* based (more like https://github.com/numical/script-ext-html-webpack-plugin)? But does that even make sense with `splitChunks`, where you don't know what the filenames will be.

There is also currently no support for configuring preload/prefetch for async chunks, but we could certainly add that. We *could* also add file-based support as an "override" - e.g. you configure the entries, but can define a pattern or callback to "override" the behavior for one specific file.

Also, if a file - e.g. `runtime.js` - is required by two entries, and both are included on the same page, what should the behavior be in the bundle? If one entry has `prefetch` and one has `preload`, the stronger `preload` should probably win? Or should the "first included entry on the page" win?

Or, we could entirely implement this in the bundle - via a new Twig function or some new options to indicate that you want prefetch/preload behavior.

Any real-world feedback is greatly appreciated!

### DIFF
--- a/index.js
+++ b/index.js
@@ -257,14 +257,30 @@ class Encore {
      * If the JavaScript file imports/requires CSS/Sass/LESS files,
      * then a CSS file (e.g. main.css) will also be output.
      *
+     * To "hint" that you would like to use defer/ajax or preload/prefetch,
+     * use the options argument. This will output to your entrypoints.json file
+     * which can be used when rendering (WebpackEncoreBundle does this):
+     *
+     *      // final output file will be main.js in the output directory
+     *      Encore.addEntry('main', './path/to/some_file.js', {
+     *          // async or defer
+     *          //async: true
+     *          //defer: true
+     *
+     *          // preload or prefetch
+     *          //preload: true
+     *          //prefetch: true
+     *      });
+     *
      * @param {string} name       The name (without extension) that will be used
      *                            as the output filename (e.g. app will become app.js)
      *                            in the output directory.
      * @param {string|Array} src  The path to the source file (or files)
+     * @param {object} options    An object of options related to the entry
      * @returns {Encore}
      */
-    addEntry(name, src) {
-        webpackConfig.addEntry(name, src);
+    addEntry(name, src, options) {
+        webpackConfig.addEntry(name, src, options);
 
         return this;
     }
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -34,6 +34,7 @@ class WebpackConfig {
 
         this.runtimeConfig = runtimeConfig;
         this.entries = new Map();
+        this.entryOptions = new Map();
         this.styleEntries = new Map();
         this.plugins = [];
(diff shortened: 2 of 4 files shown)


rakelley:
I'll try to briefly describe our use as a real world case study, but the tldr is that tying this to the entry definition may not work ideally for cases like ours.

Our site uses traditional server views as much as possible with React via `limenius/reactbundle` for elements that are significantly asynchronous or with complicated interactivity.  Pages with such functionality have a page-specific entry point and preload all entries if the elements are above the fold.  Most pages have 1 or 2 auxiliary entries for generic site elements, which are not preloaded if they're the only entries on the page.  We use defer for all scripts primarily because of not yet finding a perfect solution for supplying Google Maps to multiple components/entries in undefined order, other than a good old deferred script tag before all entries.

Because ReactOnRails (which ReactBundle uses) does not render until all deferred scripts have loaded and executed, having a page with mixed preloaded/not-preloaded entries is faster than with nothing preloaded but still limited by the last entry to load.  Page A with both `important-entry-page-a` and `entry-site-foo` wants to preload both, while Page B with only `unimportant-entry-page-b` and `entry-site-foo` does not need to preload either.  Having preload controlled at the entry definition level rather than the template use level prevents this, though admittedly centralizing configuration is always appealing and I'm not sure there's a meaningful impact to preloading these small auxiliary site entries everywhere.  I'll look into it on Monday.

File-by-file control is not necessary, as for us either an entire entry needs to be preloaded or none of it.  We aren't currently using any asynchronous chunking, but I can't see a situation in which you would want them preloaded as they're by-definition things you split off because you aren't sure you're going to use them.

weaverryan:
Closing - I think this should be handled in the bundle

Issue #445, opened 2018-11-13
TypeError: Cannot read property 'startsWith' of undefined...


Rikijs:
Hello!

I embarked on the journey to upgrade to `Encore v0.21.0`. I followed instructions (at https://symfony.com/blog/encore-0-21-0-webpack-4-support-copy-files-support-webpackencorebundle) closely. Yet i get this error and i can not figure out the problem.

Actually, this is second project that i am updating to `Encore v0.21.0` and the first one works without a  hitch. I compared configurations and can not see what could be wrong.

I tried deleting `node_modules` and rerunning `yarn install` but the problem persists.

What may the source of the problem be?
Please advise.
Thank you!

> package.json

    {
        "devDependencies": {
            "@symfony/webpack-encore": "^0.21.0",
            "node-sass": "^4.10.0",
            "sass-loader": "^7.0.1"
        },
        "license": "UNLICENSED",
        "private": true,
        "scripts": {
            "dev-server": "encore dev-server",
            "dev": "encore dev",
            "watch": "encore dev --watch",
            "build": "encore production"
        },
        "dependencies": {
            "bootstrap": "^4.1.3",
            "cookieconsent": "^3.0.6",
            "countdown": "^2.6.0",
            "jquery": "3.3.1",
            "linearicons": "^1.0.1",
            "magnific-popup": "^1.1.0",
            "moment": "^2.21.0",
            "moment-countdown": "^0.0.3",
            "popper.js": "^1.14.5"
        }
    }

> webpack.config.js

    // webpack.config.js
    var Encore = require('@symfony/webpack-encore');

    Encore
    // directory where all compiled assets will be stored
        .setOutputPath('public/build/')

        // what's the public path to this directory (relative to your project's document root dir)
        .setPublicPath('/build')

        .copyFiles({
            from: './assets/static',

            // optional target path, relative to the output dir
            to: 'static/[path][name].[ext]'

            // only copy files matching this pattern
            //pattern: /\.(png|jpg|jpeg)$/
        })

        // empty the outputPath dir before each build
        .cleanupOutputBeforeBuild()

        // will output as public/build/app.js
        .addEntry('app', './assets/js/app.js')
        .addEntry('gallery', './assets/js/gallery.js')
        .addEntry('contact_us', './assets/js/contact_us.js')

        // will output as public/build/global.css
        .addStyleEntry('main', './assets/css/main.scss')
        .addStyleEntry('admin', './assets/css/admin.scss')

[...]

PR #456, opened 2018-11-16
Display a warning when calling "copyFiles" with an invalid path

Lyrkan:
This PR adds a warning when `copyFiles()` is called with a `from` value that isn't an existing directory.

This is kind of related to #445 but even if the friendly-errors-webpack-plugin worked properly in this case the origin of the error would probably not be obvious (it'd most likely contain a reference to the temporary entry).

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -148,11 +148,25 @@ class ConfigGenerator {
             entry[sharedEntryTmpName] = tmpFilename;
         }
 
-        if (this.webpackConfig.copyFilesConfigs.length > 0) {
+        const copyFilesConfigs = this.webpackConfig.copyFilesConfigs.filter(entry => {
+            const copyFrom = path.resolve(
+                this.webpackConfig.getContext(),
+                entry.from
+            );
+
+            if (!fs.existsSync(copyFrom) || !fs.lstatSync(copyFrom).isDirectory()) {
+                logger.warning(`The "from" option of copyFiles() should be set to an existing directory but "${entry.from}" does not seem to be one. No file will be copied for this entry.`);
+                return false;
+            }
+
+            return true;
+        });
+
+        if (copyFilesConfigs.length > 0) {
             const tmpFileObject = tmp.fileSync();
             fs.writeFileSync(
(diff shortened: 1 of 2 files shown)


lib/config-generator.js
```diff
@@ -148,11 +148,25 @@ class ConfigGenerator {
...
 
-        if (this.webpackConfig.copyFilesConfigs.length > 0) {
+        const copyFilesConfigs = this.webpackConfig.copyFilesConfigs.filter(entry => {
+            const copyFrom = path.resolve(
+                this.webpackConfig.getContext(),
+                entry.from
+            );
+
+            if (!fs.existsSync(copyFrom) || !fs.lstatSync(copyFrom).isDirectory()) {
+                logger.warning(`The "from" option of copyFiles() should be set to an existing directory but "${entry.from}" does not seem to be one. No file will be copied for this entry.`);
```
weaverryan: "entry" might not be the best word here... because it means something else

lib/config-generator.js
```diff
@@ -148,11 +148,25 @@ class ConfigGenerator {
...
         }
 
-        if (this.webpackConfig.copyFilesConfigs.length > 0) {
+        const copyFilesConfigs = this.webpackConfig.copyFilesConfigs.filter(entry => {
+            const copyFrom = path.resolve(
+                this.webpackConfig.getContext(),
+                entry.from
+            );
+
+            if (!fs.existsSync(copyFrom) || !fs.lstatSync(copyFrom).isDirectory()) {
```
weaverryan: I guess we could break these into 2 separate pieces: 1 error is that the path doesn't exist. Another error is that they've pointed directly to a single file, not a directory, right?

lib/config-generator.js
Lyrkan: Hmmm, maybe... by "entry" I meant "config entry", would it be better to put that?

lib/config-generator.js
Lyrkan: 👍

weaverryan:
Thanks @Lyrkan!

PR #462, opened 2018-11-27
Always used named chunks AND prevent split chunks from changing too often.

### LINKED ISSUE #461: Chunk ids can change in production causing updated file contents
Hi guys!

I just finished hunting down a deep bug in Webpack / Encore. Here are the details:

* A) When `mode: production`, Webpack sets the `optimization.namedChunks` option to `false`. This is an internal key that's used in the generated files and when this is set to false, it uses incrementing integers.

* B) When you make certain changes (e.g. add a new entry), it may cause the chunk name for a specific chunk to change - e.g. 19 becomes 20. But this means that any files that depend on this module will change - a little `19` inside these files will be changed to `20`, even if nothing else in that file changed.

* C) To make matters worse, this updated file contents does NOT cause the file's hash to change with versioning. The result is that the same filename is built, but with different contents.

The bug is described in Webpack here: https://github.com/webpack/webpack/issues/8354 - and will likely be fixed in Webpack 5 - https://github.com/webpack/webpack/pull/8374

I believe that this is only a problem if you use `.enableSingleRuntimeChunk()` (which is the default in a new app) because this causes each file to be wrapped in code like this:

```js
(window.webpackJsonp=window.webpackJsonp||[]).push([[6]
```

Where that `[6]` is the chunk id that may change. If you do not use the single runtime chunk, the files are not wrapped with this.

I'll open a PR shortly to fix this.

weaverryan:
Fixes #461 

But, this does not expose the entry names internally in the code (as these are used instead of auto-incremet ids). I don't think there's a way around that.

Also, I'm still tracking down a somewhat related but that causes a related issue (meaningless id's changing in generated code) under very specific conditions. In fact, you can see it in this test case just by changing all the 3 of the entries to an array - e.g. 

```js
config.addEntry('main1', ['./js/code_splitting']);
if (includeExtraEntry) {
    config.addEntry('main2', ['./js/eslint']);
}
config.addEntry('main3', ['./js/no_require']);
```

That doesn't need to block this PR, but I need to figure out the issue (it's Webpack behavior) and see if it affects versioning (i.e. if the contents change but the file hashes do NOT change, this is a BIG problem, otherwise, it's more of a minor, performance problem as filenames would just be changing too often).

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -412,22 +412,12 @@ class ConfigGenerator {
             chunks: this.webpackConfig.shouldSplitEntryChunks ? 'all' : 'async'
         };
 
-        // hash the split filenames in production to protect exposing entry names
+        // This causes the split filenames (& internal names) to be,
+        // for example, 0.js. This is needed so that the filename
+        // doesn't change suddenly when another entry needs that same
+        // shared code (e.g. vendor~entry1~entry2.js).
         if (this.webpackConfig.shouldSplitEntryChunks && this.webpackConfig.isProduction()) {
-            // taken from SplitChunksPlugin
-            const hashFilename = name => {
-                return crypto
-                    .createHash('md4')
-                    .update(name)
-                    .digest('hex')
-                    .slice(0, 8);
-            };
-
-            splitChunks.name = function(module, chunks, cacheGroup) {
-                const names = chunks.map(c => c.name);
-
-                return cacheGroup + '~' + hashFilename(names.join('~'));
-            };
+            splitChunks.name = false;
         }
 
         if (this.webpackConfig.sharedCommonsEntryName) {
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -403,6 +403,10 @@ class ConfigGenerator {
             // which can be useful for debugging, especially with HMR
             optimization.namedModules = true;
         }
+        // https://github.com/webpack/webpack/issues/8354
+        // likely can be removed in Webpack 5
+        // https://github.com/webpack/webpack/pull/8374
+        optimization.chunkIds = 'named';
 
         let splitChunks = {
             chunks: this.webpackConfig.shouldSplitEntryChunks ? 'all' : 'async'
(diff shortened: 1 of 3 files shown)


test/functional.js
```diff
@@ -318,6 +318,22 @@ describe('Functional tests using webpack', function() {
...
+            testSetup.runWebpack(config, (webpackAssert) => {
+                // check that main.js has the correct contents
+                webpackAssert.assertOutputFileContains(
+                    'main.js',
+                    'Hello World!'
+                );
+
+                done();
+            });
+        });
```
weaverryan: Test just moved

weaverryan:
I just added an unrelated, second bug fix to this, because it uses much of the same code.

It is described here: https://github.com/webpack/webpack/issues/8426

Basically, when using `splitEntryChunks()`, the split filenames would change whenever a new entry required that split content - even if NOTHING else changed. The fix is simple: to set `splitChunks.name = false`, which causes the split chunk names to look like ajax split chunk names - e.g. `0.js`.

lib/config-generator.js
```diff
@@ -403,27 +403,21 @@ class ConfigGenerator {
...
 
         let splitChunks = {
             chunks: this.webpackConfig.shouldSplitEntryChunks ? 'all' : 'async'
         };
 
-        // hash the split filenames in production to protect exposing entry names
+        // This causes the split filenames (& internal names) to be,
+        // for example, 0.js. This is needed so that the filename
+        // doesn't change suddenly when another entry needs that same
+        // shared code (e.g. vendor~entry1~entry2.js).
```
stof: should this comment also include a link to the webpack issue where they suggested using `splitChunks.name = false;` ?

stof:
@weaverryan are the split chunks named `0.js` or something like `0-[hash].js` ? If there is no hash, it breaks long-term caching

Lyrkan:
@stof I did a quick check and it seems to be fine.

If the option that defines how chunks are named (`output.chunkFilename`) is not set [it takes the value of `output.filename`](https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/TemplatedPathPlugin.js#L133-L135)... which now contains `[name].[contenthash:8].js` when versioning is enabled (#463).

For instance, when adding `Encore.enableVersioning()` to the "Use splitChunks in production mode" test, it creates the following files:

* /build/runtime.6cf710cd.js
* /build/0.241ffa77.js
* /build/1.b31df8bf.js
* /build/main.a3f15efa.js
* /build/1.a60ee57d.css

stof:
OK, then things are fine.

PR #463, opened 2018-11-28
Using contenthash instead of chunkhash

weaverryan:
Since Webpack 4.3.0, a `[contenthash]` has existed, which is the [recommended way](https://webpack.js.org/guides/caching/#output-filenames) of versioning your filenames. Previously, we used a special plugin that gave us a `chunkhash`, which worked around issues with hash changing unnecessarily. However, `chunkhash` actually created a bug - in #461, some chunk id's change unnecessarily. That's bad for caching, but the *real* problem was that `[chunkhash]` did not change, and so the filenames did not change. The filenames DO change with `[contenthash]`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -993,7 +993,7 @@ class Encore {
      * file is generated.
      *
      *     Encore.configureFilenames({
-     *         js: '[name].[chunkhash].js',
+     *         js: '[name].[contenthash].js',
      *         css: '[name].[contenthash].css',
      *         images: 'images/[name].[hash:8].[ext]',
      *         fonts: 'fonts/[name].[hash:8].[ext]'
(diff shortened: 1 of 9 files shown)


weaverryan:
Good catch! I do need to change those!

> Wouldn't it be a good idea to add a warning if we detect [chunkhash] to inform users that it would be better to use [contenthash]?

Good idea... which means that, for now, we'll need to continue adding the plugin that makes this possible.

weaverryan:
Test failure is due to a change to the hash calculation in Webpack 4.22.0. I'm going to ignore, so we can release these bug fixes as a patch release. Then, we can bump the minimum Webpack version from `^4.20.0` to `^4.22.0` in another PR for a minor release.

PR #472, opened 2018-12-13
Store externals in an array. This fixes #471

### LINKED ISSUE #471: Encore.addExternals breaks Webpack config with functional external
The externals property of the webpack config states that externals may be one of the following: string, object, function, or regex. Encore.addExternals only permits externals of type object. Attempting to pass anything but an object results in an build exception.

E.g. the following will fail during build:

```javascript
const nodeExternals = require('webpack-node-externals');

Encore
   ...
   .addExternals(nodeExternals());
   // Note: .addExternals([nodeExternals()]) fails, but should also be accepted

```

However the following works just fine:

```javascript
const nodeExternals = require('webpack-node-externals');

Encore
    ...

config.externals = nodeExternals();
// config.externals = [nodeExternals()]; also works

module.exports = config;
```

deAtog:


lib/config-generator.js
```diff
@@ -107,7 +107,7 @@ class ConfigGenerator {
             config.resolve.alias['react-dom'] = 'preact-compat';
         }
 
-        config.externals = Object.assign({}, this.webpackConfig.externals);
+        config.externals = this.webpackConfig.externals;
```
Lyrkan: It would probably be better to copy the array here to avoid it from being modified between two `getWebpackConfig()` calls:

```suggestion
        config.externals = [...this.webpackConfig.externals];
```

It won't prevent deep changes (maybe we'll have to do something about this later) but will handle obvious "adding/removing an external" situations.

test/WebpackConfig.js
```diff
@@ -940,24 +940,15 @@ describe('WebpackConfig object', () => {
...
             config.addExternals({ 'lodash': 'lodash' });
 
-            expect(config.externals).to.deep.equals({
-                'jquery': 'jQuery',
-                'react': 'react',
-                'lodash': 'lodash'
-            });
-        });
-
-        it('Calling it with an invalid argument', () => {
```
Lyrkan: Shouldn't we keep that test and check for the types that are actually accepted by `config.externals` in Webpack?

deAtog: My thought here was to just let Webpack report an error with any invalid externals. From what I can tell though, if the external is not a string, object, regex, or function webpack just ignores it.

lib/config-generator.js
deAtog: I considered this, but figured since users can only add externals there would be no problem directly referencing the internal array. The only other indirect manipulation of this array is via the reset() function when it creates a whole new webpackConfig.

Lyrkan: That's a bit of an edge case but the issue is that the generated config can be manipulated at will after being generated.

For instance with a multi-config export:

```js
const config1 = Encore.getWebpackConfig();
const config2 = Encore.getWebpackConfig();

config1.externals.push('onlyConf1');

module.exports = [config1, config2];
```

Each call to `getWebpackConfig()` should generate its own config object... but if the array is not copied the `onlyConf1` external will be present in both of them.

As I said above, that's probably an edge case since most of the time people won't call `getWebpackConfig()` multiple times without using `reset()`... but it doesn't cost much to prevent it (well... at least the most obvious cases with a shallow copy).

test/WebpackConfig.js
Lyrkan: Yeah maybe that's not needed... since it accepts nearly everything it's unlikely someone will try to pass it something invalid.

deAtog:
I modified the addExternals interface slightly to allow multiple externals to be added without having to explicitly provide an array and updated the documentation accordingly. Overall, I feel as though this is a slightly better approach as it reduces the burden on the user.

lib/WebpackConfig.js
```diff
@@ -298,12 +298,8 @@ class WebpackConfig {
         Object.assign(this.aliases, aliases);
     }
 
-    addExternals(externals = {}) {
-        if (typeof externals !== 'object') {
-            throw new Error('Argument 1 to addExternals() must be an object.');
-        }
-
-        Object.assign(this.externals, externals);
+    addExternals(...externals) {
```
Lyrkan: Nitpicking but I'm not sure that rest parameters are needed there since this is not part of the exposed API.
It adds a bit of unneeded complexity to the call in `index.js` (you have to use `apply` instead of just doing `webpackConfig.addExternals(externals)`).

Lyrkan: Do you mean Babel? If that's the case it's not relevant here since our code don't go through Babel.

Having `...externals`may be a bit easier to understand (and Node probably optimize it... and even if that's not the case it won't be a big overhead anyway :)) but I still don't really see the benefits of using rest parameters here instead of simply passing the array directly.

deAtog: The reason for using rest parameters was to simplify the API and the function itself. By using rest parameters or the `arguments` array, there's no need to check if the parameter is an array in order to perform the concatenation. Users can still pass an array if they want, it will just add the array to the externals list instead of appending each element. Both mechanisms work.

I meant Babel, but seeing as how it is not used, it's irrelevant. Nonetheless I would hope Node internally just converts the ES6 syntax to an `apply` call. I will change the code to the ES6 syntax to be consistent with the new API.

test/WebpackConfig.js
```diff
@@ -940,24 +940,15 @@ describe('WebpackConfig object', () => {
         it('Adds new externals', () => {
             const config = createConfig();
 
-            expect(config.externals).to.deep.equals({});
+            expect(config.externals).to.deep.equals([]);
 
             config.addExternals({ 'jquery': 'jQuery', 'react': 'react' });
             config.addExternals({ 'lodash': 'lodash' });
```
weaverryan: Can you update this test by adding, for example, another `addExternal()` that passes a regex?

lib/WebpackConfig.js
weaverryan: Hmm, I agree with @Lyrkan. Look at the "mixed" syntax: https://webpack.js.org/configuration/externals/#combining-syntaxes

Basically, when you call `addExternal()`, there should only be 3 valid things that you can pass to it:

```js
Encore
     // an object, though the value of each entry can be a string, array or object - but we don't care
     .addExternal({jquery: 'jQuery'})

    // a function
    .addExternal(function(context, request, callback) { /*... */ })

    // regex
    .addExternal(/^(jquery|\$)$/i)
```

The current implementation allows you to pass these are multiple arguments for multiple externals. That's nice - but I'd rather just accept 1 argument and check that it's one of these 3 types.

@deAtog Could you make that change? If not - just let us know. Sorry for the delay :)

deAtog:
Closing this pull request as it has been superseded by a more recent version.

PR #484, opened 2019-01-04
Allow to import .vue files when using TypeScript

### LINKED ISSUE #467: Import Components from vue does not work
Hi everyone,

I'm sorry in advance because i'm not sure that this issue has it place here.

I'm using vue in my Symfony project using webpack-encore's config. My webpack config is really basic and it works but, the problem is in my main.ts, i would like to import a component (Header.vue that is in ./test/Header.vue).

- When i'm using the import statement -> import Header from "./test/Header.vue", it doesn't work and return that cannot find the module.
- When i'm using require statement -> const Header = require('./test/Header.vue').default it works

It seems like webpack-encore babel's configuration does not read ES2015 statement ( and i really want to use it );

My question is, how can i configure babel (if the problem comes from it) to make the import statement works ?

Thanks !
```
// main.ts
import Vue from 'vue'; // Here it work
import store from './store'; // Here too

// import Header from "./test/Header.vue"; // But here not :'(
// const Header = require('./test/Header.vue').default;

 // Vue.component("test", Header);

new Vue({
    store,
    delimiters: ["%%", "%%"],
}).$mount('#app');

require('./styles/main.scss');
```

Lyrkan:
With the current default config importing a `.vue` file from a `.ts` file results in an error because the `ts-loader` don't know how to handle Vue files.

This PR resolves that issue by setting the `appendTsSuffixTo` option of the `ts-loader` to `[/\.vue$/]` (as described [here](https://vue-loader.vuejs.org/guide/pre-processors.html#typescript)) when the `vue-loader` is enabled (fixes #467).


weaverryan:
Excellent job @Lyrkan! Love the test!

Lyrkan:
The test is basically the one we already had for Vue.js but converted to TypeScript :p

PR #485, opened 2019-01-04
Fix small mistakes in shared-entry-concat-plugin.js and validator.js

Lyrkan:
Just two small issues I noticed while playing with the TypeScript compiler :)

### DIFF
--- a/lib/webpack/shared-entry-concat-plugin.js
+++ b/lib/webpack/shared-entry-concat-plugin.js
@@ -41,7 +41,7 @@ function getChunkFilename(compilation, chunkName) {
 function getAssetSource(asset) {
     let content = asset.source();
     if (Buffer.isBuffer(content)) {
-        content = Buffer.toString('utf-8');
+        content = content.toString('utf-8');
     }
 
     return content;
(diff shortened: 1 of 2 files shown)


lib/webpack/shared-entry-concat-plugin.js
```diff
@@ -41,7 +41,7 @@ function getChunkFilename(compilation, chunkName) {
 function getAssetSource(asset) {
     let content = asset.source();
     if (Buffer.isBuffer(content)) {
-        content = Buffer.toString('utf-8');
+        content = content.toString('utf-8');
```
weaverryan: This worked before somehow? Or, maybe being a Buffer is some edge-case?

weaverryan:
Thanks @Lyrkan!

lib/webpack/shared-entry-concat-plugin.js
Lyrkan: Yeah, most of the time it probably doesn't go into that `if` (not sure what triggers it).

However if it did the build would not fail (since `Buffer.toString('utf-8')` *does* return a JS string) but the generated file wouldn't work either... I don't think I've seen someone post an issue that could be related to that though.

Issue #126, opened 2017-08-01
dev-server usage within vagrant


scuben:
I have troubles getting `dev-server` command working. While `production` and `dev` arguments work like expected and the assets being build correctly I do have troubles with `dev-server`.

My setup is a vagrant machine with virtualbox. I can access the vm by ip 172.16.23.10 or hostname (e.g. website.company.dev). So far, so good! Problem arises when I try to use `dev-server` within my vm (not on my local machine!). Here are my steps I already tried to make it working:

1. run `node_modules/.bin/encore dev-server`. The output is:
Running webpack-dev-server ...
Content not from webpack is served from /vagrant
404s will fallback to /index.html
DONE  Compiled successfully in 12873ms

Due tu the fact that the server within my vm is binding itself to `127.0.0.1:8080` I cant access it from my local machine. next try:

2. run `node_modules/.bin/encore dev-server --host 0.0.0.0`. Same output as in my first try. But now the server within my vm is bind to `0.0.0.0:8080`.

Now I can access the server from my local machine using my browser to request `http://website.company.dev:8080`. Now I get a `Invalid Host header` Response. next try:

3. run `node_modules/.bin/encore dev-server --host 0.0.0.0 --public website.company.dev`. Same output as in my first try. I could als have used `--disable-host-check` option but I like to just set the hostname right. Browsing the site with `http://website.company.dev:8080` now gives me the error I cant get around:

> Cannot GET /

According to my nginx access logfile it doesn't even hit the underlying webserver at all to serve my site (not the assets). Accessing `http://website.company.dev:8080/webpack-dev-server` is showing me by build files:

```
fonts
glyphicons-halflings-regular.eot
glyphicons-halflings-regular.ttf
glyphicons-halflings-regular.woff
glyphicons-halflings-regular.woff2
images
glyphicons-halflings-regular.svg
app.js
app (magic html for app.js) (webpack-dev-server)
main.css
manifest.json
```
What else ist there to say? I dont know why its not working. I'll post my config and dependency files to give you an overview on how I use encore. Any help or hints are much appreciated!

**package.json**
```
{
  "devDependencies": {
    "@symfony/webpack-encore": "^0.12.0",
    "bootstrap-sass": "^3.3.7",
    "jquery": "^3.2.1",
    "node-sass": "^4.5.3",
    "postcss-loader": "^2.0.6",
    "sass-loader": "^6.0.6"
  },
  "browserslist": [
    "last 2 version"
  ],
  "scripts": {
[...]

Issue #35, opened 2017-06-19
Watch config


julienj:
Hello,
watch option doesn't work with docker machine and nfs.
cf : http://andrewhfarmer.com/webpack-watch-in-vagrant-docker/

How can I add this config 
 
`
watchOptions: {
  poll: true
}
`
Thank

stof:
As Encore is a builder for the webpack config, you can easily alter the final config before exporting it:

```js
// webpack.config.js

var Encore = require('@symfony/webpack-encore');

// all Encore config here

var config = Encore.getWebpackConfig();

config.watchOptions = { poll: true };

module.exports = config;
```

weaverryan:
We should doc this as the official approach of adding "extra" config. I don't think there's any benefit of adding another method on Encore to set things like this. I've opened and issue: symfony/symfony-docs#8067

Thanks!

julienj:
Thank you
Is a good news

PR #486, opened 2019-01-07
feat: configure watching options

Kocal:
This PR adds support for _properly_ configuring watching options.
References: #35, #126

Before, we had to do this:
```js
const config = Encore.getWebpackConfig();

if (config.devServer) {
  config.devServer.watchOptions.poll = 250;
}
```

Now, we can do this:
```js
Encore.configureWatchOptions(watchOptions => {
    watchOptions.poll = 250;
});
```

Also, I think it would be better to rename it to `configureWatchOptions()`, but I don't know how encore apply watching options to webpack when running dev or prod environment with `encore [prod|dev] --watch`. :thinking: 

Thanks!

---

**EDIT**: Ah, I found it! 
selection_052

I will see how can I modify `watchOptions` in dev/prod modes.

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -527,6 +527,11 @@ class ConfigGenerator {
 
     buildDevServerConfig() {
         const contentBase = pathUtil.getContentBase(this.webpackConfig);
+        const watchOptions = {
+            ignored: /node_modules/
+        };
+
+        this.webpackConfig.devServerWatchOptionsConfigurationCallback(watchOptions);
 
         return {
             contentBase: contentBase,
@@ -539,9 +544,7 @@ class ConfigGenerator {
             quiet: true,
             compress: true,
             historyApiFallback: true,
-            watchOptions: {
-                ignored: /node_modules/
-            },
+            watchOptions: watchOptions,
             https: this.webpackConfig.useDevServerInHttps()
         };
     }
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -531,9 +531,10 @@ class ConfigGenerator {
             ignored: /node_modules/
         };
 
-        this.webpackConfig.watchOptionsConfigurationCallback(watchOptions);
-
-        return watchOptions;
+        return applyOptionsCallback(
+            this.webpackConfig.watchOptionsConfigurationCallback,
+            watchOptions
+        );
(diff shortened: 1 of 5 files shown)


Lyrkan:
Hi @Kocal,

`Encore.configureDevServerWatchOptions()` feels a bit long to me, `configureWatchOptions()` would probably be better imo.

That being said... I'm wondering if we shouldn't have a `configureDevServer` instead?  Is `watchOptions` really *that* special to justify giving it its own method?

Kocal:
Watch mode can be used without a dev-server, so I think it's better to have a `.configureWatchOptions()` methods that applies both for [webpack](https://webpack.js.org/configuration/watch/) and [webpack-dev-server](https://webpack.js.org/configuration/dev-server/#devserver-watchoptions-).

lib/config-generator.js
```diff
@@ -525,6 +526,16 @@ class ConfigGenerator {
         return stats;
     }
 
+    buildWatchOptionsConfig() {
+        const watchOptions = {
+            ignored: /node_modules/
+        };
+
+        this.webpackConfig.watchOptionsConfigurationCallback(watchOptions);
```
stof: this should use `applyOptionsCallback` to support the different patterns for the callback (either mutating the argument, or returning a new object), as we do elsewhere.

Kocal: Ah yes you're right, updated.

Kocal:
What should I do with failing tests? It's working fine on local :confused: 
capture d ecran de 2019-01-07 20-22-01

Lyrkan:
Don't worry about the failing tests, they are not related to your PR :)

You don't have them locally because they are in a Travis job that tests the highest possible version of the dependencies (it removes the lock file before downloading them) and sometimes Webpack changes something that causes the hardcoded hashes from our functional tests to become outdated.

Kocal:
Oh okay, thanks for the explanation :+1:

Kocal:
Who is ready for a new review? :smiley_cat:

Lyrkan:
Thank you @Kocal
As you can see in https://github.com/symfony/symfony-docs/pull/10923 this was definitely needed :)

Kocal:
Yep! :)

PR #495, opened 2019-01-15
Fix support for Webpack externals. This fixes #471

### LINKED ISSUE #471: Encore.addExternals breaks Webpack config with functional external

deAtog:
This is an alternative to https://github.com/symfony/webpack-encore/pull/472 which does not modify the public Encore API and requires the user to pass an array if they want to add multiple externals at a time.

### DIFF
--- a/index.js
+++ b/index.js
@@ -397,7 +397,9 @@ class Encore {
      *      const nodeExternals = require('webpack-node-externals');
      *
      *      Encore.addExternals(
-     *          nodeExternals()
+     *          // add any valid externals you have
+     *          nodeExternals(),
+     *          /^(jquery|\$)$/i
      *      );
      *
      * @param {*} externals
--- a/index.js
+++ b/index.js
@@ -387,15 +387,18 @@ class Encore {
      *
      * For example:
      *
+     *      Encore.addExternals({
+     *          jquery: 'jQuery',
+     *          react: 'react'
+     *      });
+     *
+     * Or:
+     *
      *      const nodeExternals = require('webpack-node-externals');
      *
-     *      Encore.addExternals([
-     *          {
-     *             jquery: 'jQuery',
-     *             react: 'react'
-     *          },
+     *      Encore.addExternals(
      *          nodeExternals()
-     *      ]);
+     *      );
      *
      * @param {*} externals
      *
--- a/index.js
+++ b/index.js
@@ -397,7 +397,7 @@ class Encore {
      *          nodeExternals()
      *      ]);
      *
-     * @param {object} externals
+     * @param {*} externals
      *
      * @returns {Encore}
      */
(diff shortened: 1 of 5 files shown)


deAtog:
@Lyrkan fixed the linting issues.

index.js
```diff
@@ -387,10 +387,15 @@ class Encore {
      *
      * For example:
      *
-     *      Encore.addExternals({
-     *          jquery: 'jQuery',
-     *          react: 'react'
-     *      })
+     *      const nodeExternals = require('webpack-node-externals');
```
Lyrkan: I'm wondering if that's a good idea to put this specific example. I'm a bit scared that some people will think this is a requirement in order to add externals...

index.js
```diff
@@ -387,10 +387,15 @@ class Encore {
...
+     *
+     *      Encore.addExternals([
+     *          {
+     *             jquery: 'jQuery',
+     *             react: 'react'
+     *          },
+     *          nodeExternals()
+     *      ]);
      *
      * @param {object} externals
```
Lyrkan: Maybe `{*}` instead of `{object}`?

index.js
deAtog: Feel free to suggest an alternate example, but webpack-node-externals is probably the most commonly used functional usage here. Maybe spitting it into two examples would be better? One that only add jquery and react, and one that uses webpack-node-externals?

index.js
deAtog: Yeah, {*} would indeed be better.

index.js
deAtog: I think splitting is better since I really can't see someone defining 'jquery' and 'react' as externals and then using webpack-node-externals in the same config.

index.js
```diff
@@ -397,7 +397,9 @@ class Encore {
      *      const nodeExternals = require('webpack-node-externals');
      *
      *      Encore.addExternals(
-     *          nodeExternals()
+     *          // add any valid externals you have
+     *          nodeExternals(),
+     *          /^(jquery|\$)$/i
```
Lyrkan: That PR is the one that doesn't use rest parameters (as opposed to #472), so you'd need to put these two lines into an array.

weaverryan:
Thank you @deAtog!

PR #497, opened 2019-01-18
feat(terser): enable cache and parallelism

Kocal:
Recently Laravel Mix have sped up build time by configuring Terser plugin   (https://github.com/JeffreyWay/laravel-mix/pull/1925).

I though `cache` and `parallelism` options was already set on encore, but nope.

Here a comparaison, before:
capture d ecran de 2019-01-18 06-49-45

and after:
capture d ecran de 2019-01-18 06-51-05

From 16-18 seconds to 14-8 seconds on my machine. 
And even if my project does not have a tons of JavaScript, the building time have been reduced.

### DIFF
--- a/lib/plugins/terser.js
+++ b/lib/plugins/terser.js
@@ -18,7 +18,9 @@ const applyOptionsCallback = require('../utils/apply-options-callback');
  */
 module.exports = function(webpackConfig) {
     const terserPluginOptions = {
-        sourceMap: webpackConfig.useSourceMaps
+        sourceMap: webpackConfig.useSourceMaps,
+        cache: true,
+        parallel: true,
     };
 
     return new TerserPlugin(


lib/plugins/terser.js
```diff
@@ -18,7 +18,9 @@ const applyOptionsCallback = require('../utils/apply-options-callback');
  */
 module.exports = function(webpackConfig) {
     const terserPluginOptions = {
-        sourceMap: webpackConfig.useSourceMaps
+        sourceMap: webpackConfig.useSourceMaps,
+        cache: true,
+        parallel: true,
```
Seb33300: Is it normal to end the last line with a comma?

Kocal: Yes this is allowed in JavaScript (like in TypeScript, Python, PHP...).

However I'm not sure there is a specifing style/linting rule on Encore.


See https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Virgules_finales for JavaScript

PR #499, opened 2019-01-20
chore: rename AppVeyor file

Kocal:
Since we have `.travis.yml` for Travis, I guess it would be better to have `.appveyor.yml` for AppVeyor.

Doc: https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming


Lyrkan:
Thank you @Kocal, I guess it can't hurt to have consistent file names.
Especially since `.appveyor.yml` is also used in symfony/symfony (I think you were the one that changed it there too :))

Kocal:
Yep that was me :stuck_out_tongue_closed_eyes:

Issue #236, opened 2018-01-05
PHPStorm integration


plfort:
PHPStorm can interpret standart Webpack config file but not the Encore webpack.config.js : 

> Can't analyse webpack.config.js: coding assistance will ignore module resolution rules in this file.
			Possible reasons: this file is not a valid webpack configuration file or its format is not currently supported by the IDE

I made this workaround to handle this : https://gist.github.com/plfort/93da0ee1465f1d7f929fe5d9e6307532

Do you have a better solution ?

weaverryan:
Hey there!

I’m aware of this issue... but I think it’s not an issue (but it *is* confusing people). As far as I understand it, phpstorm has support to analyze you mr wepback.config.js file, which can help if you’re writing it by hand. But Encore generates this confgi for you, so I don’t think there’s any benefit to the phpstorm analysis anyways when using Encore. In other words, the warning should just be ignored.

Does the Phpstorm analysis give you any useful functionality with your workaround? Or is it just a way to make the earning disappear?

plfort:
PHPStorm can resolve modules with my workaround : https://www.jetbrains.com/help/phpstorm/webpack.html#webpack_module_resolution
This is really helpful when working with aliases

weaverryan:
Ah, so it uses the roots and aliases. Hmm, interesting. I don’t know if we can fix this or not, but it’s a legit issue.

anurbol:
If you want to get rid of this problem you can remove the notifications & log for the Webpack from Settings->Apperance&Behavior->Notifications->Webpack

https://stackoverflow.com/a/48562153/9259778

florentdestremau:
I don't have the autocomplete for the Encore library. Not very damaging but a bit annoying, i need to Ctrl + click on the `require('@symfony/webpack-encore');` to get a list of all the methods and explore the plugin by hand. 

Is there a link with this issue ?

Lyrkan:
@florentdestremau Nop, that's not related to this issue but to #151.

Aerendir:
This is definitely an annoying issue.

The trick suggested by @plfort works great (after changing the default `webpack.config.js` to `fakewebpackconfig.js` in `Preferences > Languages & Framework > JavaScript > Webpack`.

Maybe `Encore` can create the file using a method like `.createFakeConfigForPHPStorm` but I understand its name is really coupled with the PHPStorm editor and that this trick maybe useful in other editors too...

Lyrkan:
[...]

PR #500, opened 2019-01-23
Add a public method to know whether the runtime is configured

stof:
Refs #236

### DIFF
--- a/index.js
+++ b/index.js
@@ -1157,6 +1157,10 @@ class Encore {
         return this;
     }
 
+    isRuntimeEnvironmentConfigured() {
+        return runtimeConfig !== null;
+    }
+
     /**
      * Clear the runtime environment.
      *
@@ -1213,6 +1217,7 @@ const EncoreProxy = new Proxy(new Encore(), {
             const safeMethods = [
                 'configureRuntimeEnvironment',
                 'clearRuntimeEnvironment',
+                'isRuntimeEnvironmentConfigured',
             ];
 
             if (!webpackConfig && (safeMethods.indexOf(prop) === -1)) {
--- a/test/index.js
+++ b/test/index.js
@@ -414,6 +414,22 @@ describe('Public API', () => {
 
     });
 
+    describe('isRuntimeEnvironmentConfigured', () => {
+
+        it('should return true if the runtime environment has been configured', () => {
+            const returnedValue = api.isRuntimeEnvironmentConfigured();
+            expect(returnedValue).to.be.true;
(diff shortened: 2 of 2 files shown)


stof:
Failure in the job `Highest versions of the dependencies` is unrelated to that change. It looks like the hash of the runtime is different for the new version of webpack (they might have altered their runtime a bit).

stof:
Once this is merged and released, we can update the recipe to use it by default.

stof:
and documentation too.

Kocal:
Thanks for the feature, it will really helps with PhpStorm webpack integration. :+1: 

Also I've just seen some linting issues. In the whole file, it use `it(` instead of `it (`, and there is always a blank line between `});` (from `it()`) and `});` (from `describe()`).
I think it would be nice to correct them to keep something consistent. :)

test/index.js
```diff
@@ -414,6 +414,21 @@ describe('Public API', () => {
...
     });
 
+    describe('isRuntimeConfigured', () => {
+
+        it ('should return true if the runtime environment has been configured', () => {
+            const returnedValue = api.isRuntimeConfigured();
+            expect(returnedValue).to.be.true;
+        });
+
+        it ('should return false if the runtime environment has been configured', () => {
```
javiereguiluz: `has been configured` -> `has not been configured` ?

stof: indeed, bad copy paste

stof:
> Also I've just seen some linting issues.

these don't seem to be part of the defined coding standards, as the linting is green on CI. But they are now fixed.

stof:
should it be marked as an allowed failure then if you don't care about the fact it fail ? That would make the github status more useful.

Lyrkan:
I don't know if allowing the job to fail is the right solution... it could be a temporary one though.

What I meant in my other comment is that it's not important if it fails for that specific reason but it could be for another one.

In my opinion the proper way to do it would be to ignore some checks (such as the ones that are based on a hash that we can't control) only when running that job, but I haven't had the time to look into that.

index.js
```diff
@@ -1157,6 +1157,10 @@ class Encore {
         return this;
     }
 
+    isRuntimeConfigured() {
```
stof: Question here: should it be named `isRuntimeConfigured` or `isRuntimeEnvironmentConfigured` ? Other methods are taking about `RuntimeEnvironment`.

Lyrkan:
Let's merge it that way, thanks @stof

PR #501, opened 2019-01-24
Fix various issues related to functional tests

Lyrkan:
This PR fixes various things that currently make the tests result indicators quite unreliable:

* Set a `DISABLE_UNSTABLE_CHECKS` environment variable and use it to disable some checks related to filenames with hashes that are subject to frequent changes (for instance `runtime.xxxxxxxx.js` when updating Webpack). This only applies to "Lowest versions of the dependencies" and "Highest versions of the dependencies" jobs.

* Fix the Vue.js + Typescript test that was silently failing in the "Lowest versions of the dependencies" job (no idea why it worked for the other ones...)

* Increase default timeouts of functional tests from 8s to 10s (since they were frequently failing because of that)

* Restore the standard output at the beginning of the compiler callback in `runWebpack()` (doing it later could end up hiding the rest of the job's output if an exception was thrown)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,9 @@ matrix:
     - name: 'Lowest versions of the dependencies'
       os: linux
       node_js: "10"
-      env: JOB_PART=test
+      env:
+        - JOB_PART=test
+        - DISABLE_UNSTABLE_CHECKS=1
       install:
         - rm yarn.lock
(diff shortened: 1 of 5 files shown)


weaverryan:
What? The tests are green! Very sensible, and impressive updates.T Thank you @Lyrkan! We really needed to get this cleaned up!

PR #503, opened 2019-01-29
Remove debug logging in some tests

stof:
These logging calls are written in the output while running tests, which look weird in the test output.
I assume they were added to debug stuff while writing tests, but they should not be necessary anymore.


Lyrkan:
Thanks!

Issue #473, opened 2018-12-20
Possible incompatibility with `enableEslintLoader()` and eslint-vue-plugin


Kocal:
Hi, I am the [frontend dev that was blaming `webpack-encore` one year ago](https://github.com/symfony/webpack-encore/issues/277#issue-302738146) :eyes: 

First, let me congratulate you @weaverryan and the community, `webpack-encore` seems has become very mature over time. It's nice to have webpack 4, Babel 7 and ESLint working out of the box. 

Even if I still prefer using [`vue-cli`](https://github.com/vuejs/vue-cli) for a SPA (with Symfony as API) or a raw simple webpack config for smaller project, `webpack-encore` is really a must have for _more oriented_ Symfony projects. 

---

So I'm facing an issue with using the ESLint loader with [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue) for linting `.vue` files.

This is how I configure the `eslint-loader`:

```js
Encore.enableEslintLoader(config => {
  config.configFile = __dirname + '/.eslintrc';
});
```

This is my `.eslintrc`:

```json
{
  "extends": [
    "airbnb-base",
    "plugin:vue/recommended",
    "@yproximite/base"
  ],
  "rules": {
      ...
   }
}
```

And this is my ESLint dependencies:
```
❱Development❰ app@condictio.vm:/srv/app|⇒  yarn list --pattern eslint  
yarn list v1.12.3
├─ @yproximite/eslint-config-base@0.2.0
│  ├─ babel-eslint@10.0.1
│  └─ eslint-scope@3.7.1
├─ babel-eslint@8.2.6
│  └─ eslint-scope@3.7.1
├─ eslint-config-airbnb-base@13.1.0
├─ eslint-config-prettier@3.3.0
├─ eslint-import-resolver-node@0.3.2
├─ eslint-loader@1.9.0
├─ eslint-module-utils@2.2.0
├─ eslint-plugin-import@2.14.0
├─ eslint-plugin-prettier@3.0.0
├─ eslint-plugin-vue@5.0.0
├─ eslint-restricted-globals@0.1.1
├─ eslint-scope@4.0.0
├─ eslint-utils@1.3.1
├─ eslint-visitor-keys@1.0.0
├─ eslint@5.9.0
└─ vue-eslint-parser@4.0.3

```

And this is what happens when running `webpack-encore`:

[...]

PR #504, opened 2019-01-29
enable eslinting of .vue files

elmariachi111:
addresses #473 :  if the eslintloader is activated aside the vue loader it also must load .vue files. This proposal might not be complete: for that case we could also add dependency notes so users are hinted to install `vue-eslint-parser`.

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -318,9 +318,9 @@ class ConfigGenerator {
         }
 
         if (this.webpackConfig.useEslintLoader) {
-            const test = this.webpackConfig.useVueLoader 
-            ? /\.(jsx?|vue)$/
-            : /\.jsx?$/
+            const test = this.webpackConfig.useVueLoader
+                ? /\.(jsx?|vue)$/
+                : /\.jsx?$/;
                         
             rules.push({
                 test,
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -318,8 +318,12 @@ class ConfigGenerator {
         }
 
         if (this.webpackConfig.useEslintLoader) {
+            const test = this.webpackConfig.useVueLoader 
+            ? /\.(jsx?|vue)$/
+            : /\.jsx?$/
+                        
             rules.push({
-                test: /\.jsx?$/,
+                test,
                 loader: 'eslint-loader',
                 exclude: /node_modules/,
                 enforce: 'pre',
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -321,7 +321,7 @@ class ConfigGenerator {
             const test = this.webpackConfig.useVueLoader
                 ? /\.(jsx?|vue)$/
                 : /\.jsx?$/;
-                        
+
             rules.push({
                 test,
                 loader: 'eslint-loader',


Kocal:
Aaaah so it was coming from this...

I didn't even think about it, thanks!

lib/config-generator.js
```diff
@@ -318,8 +318,12 @@ class ConfigGenerator {
         }
 
         if (this.webpackConfig.useEslintLoader) {
+            const test = this.webpackConfig.useVueLoader
```
Lyrkan: I'm wondering if that test is needed... would it be a bad thing if we always included `.vue` files? If the vue loader isn't enabled (and not added manually) the build would fail anyway right?

If we always include `.vue` file:
* it makes our code a bit simpler
* if for some reason someone chooses not to call `enableVueLoader` but set-up its own loader manually they can still use our default ESLint config

Kocal: > If the vue loader isn't enabled (and not added manually) the build would fail anyway right?

Yes, I think ESLint will throw an error like this if the user didn't have configurer eslint-plugin-vue: 
> Unexpected token &lt;template&gt; ....


A safer way is to let the user configure ESLint loader:

```js
Encore
  .enableVueLoader()
  .enableEslintLoader()
  .configureEslintLoader(loader => {
    loader.test = /.(jsx?|vue)$/
  })
```

It's a bit more verbose, but it will prevent some issues.

Kocal: Or maybe a more generic method, if we don't want to add one method for each loader: 

```js
Encore
  .configureLoader('eslint', loader => {
    loader.test = /.(jsx?|vue)$/
  });

Comment (Lyrkan) [reply lib/config-generator.js]: > Yes, I think ESLint will throw an error like this if the user didn't have configurer eslint-plugin-vue

Hm, yes but that's only if you try to import a `.vue` file right?

What I meant is that if you don't enable the `vue-loader` the build will fail because Webpack won't know how to process those files anyway.

And if you don't use `.vue` files, well, that's not an issue because you'll never get that error.

This raises a point though... should we force people to use the Eslint vue plugin by default if they have `.vue` files?

Comment (Kocal) [reply lib/config-generator.js]: > Hm, yes but that's only if you try to import a .vue file right?

Yep

> What I meant is that if you don't enable the vue-loader the build will fail because Webpack won't know how to process those files anyway.
>
> And if you don't use .vue files, well, that's not an issue because you'll never get that error.

Hum you're right, I didn't think this way.

> This raises a point though... should we force people to use the Eslint vue plugin by default if they have .vue files?

I don't know. I'm probably sure that some people will complain about that. :thinking: 

But if we do, we should not hardcode eslint-plugin-vue configuration inside Encore.

**Comment by Kocal** [comment]
I guess this one can be closed since #509 has been merged.

**Comment by Lyrkan** [comment]
@Kocal I'm not sure, due to how it works `configureLoaderRule()` should only be used as a last recourse in my opinion. Linting vue file seems like a really common use case to me and we could probably make it work...

Maybe not automatically but through an `enableEslintLoader(...)` option?

**Comment by Kocal** [comment]
Yes it's common to linting Vue files, but I don't think it's a good idea to force the user.

The good behavior for me is, if the used enabled ESLint loader and VueLoader, then: 
  - if the dependency `eslint-plugin-vue` is installed, we enable linting on Vue files
  - if not, we display a warning like `If you want to lint Vue files, then install the dependency "eslint-plugin-vue" [...]`

Also we should probably tell the user he can configure ESLint+Vue with the following code:
```js
Encore.configureEslintLoader({
  extends: [
    // add more generic rulesets here, such as:
    // 'eslint:recommended',
    'plugin:vue/recommended'
  ]
});
```
(In fact I'm not even sure that we can do this :sweat_smile:)

What do you think?

masi:
> What do you think?

As a user I like this proposal.

Lyrkan:
I don't really like the idea of automatically enabling features based on which packages are installed (especially since the notion of "installed" isn't as obvious as it seems because of dependency hoisting).

I think we could probably just tweak `Encore.enableEslintLoader()` (which currently only accepts 1 argument) to automatically enable the plugin + `vue/recommended`:

```js
Encore.enableEslintLoader(options => {
    // This is applied on the final options object
    // options.extends = [ /* ... */ ];
}, {
    // Make the eslint-loader also process `.vue`
    // files and add `plugin:vue/recommended` to
    // extended rulesets.
    useVuePlugin: true,
});
```

masi:
> I think we could probably just tweak `Encore.enableEslintLoader()` (which currently only accepts 1 argument) to automatically enable the plugin + `vue/recommended`:

Ok. Though Encore is opinioned in other areas and enables features you have even to look up the sourcecode to know they are turned on (eg Terser or Define).

BTW, if I have enables TypeScript I want to linit it too. Now that ESLint has taken over from TSLint you want to probably have something like this as well:

```
Encore.enableEslintLoader(options => {
    // This is applied on the final options object
}, {
    // Make the eslint-loader also process `.ts` files
    // and add `plugin:@typescript-eslint/recommended` to
    // extended rulesets.
    useTypescriptPlugin: true,
});
```

Lyrkan:
> Now that ESLint has taken over from TSLint

Is that already the case? As far as I know TSLint will only be deprecated once ESLint checks are on the same level than the ones from TSLint... which is far from being the case [based on the following roadmap](https://github.com/typescript-eslint/typescript-eslint/blob/9b7e6920aecd2f0ebb5d19c0952c1dc654705b7f/packages/eslint-plugin/ROADMAP.md).

Also, since a TSLint => ESLint compat package is planned (cf. [this post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9)), it doesn't really make sense to start using ESLint for TypeScript files.

masi:
> Also, since a TSLint => ESLint compat package is planned (cf. [this post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9)), it doesn't really make sense to start using ESLint for TypeScript files.
As I read it the compat package is for ESLint, so former TSLint users can use their old TSLint rules with ESLint. Besides the fact that ESLint isnt there at 100% Webpack has made to move: https://github.com/webpack/webpack-cli/pull/834
But I digress this this is about .vue support though for my use I will have to get either TSLint or ESLing workin with .vue files.

Lyrkan:
> As I read it the compat package is for ESLint, so former TSLint users can use their old TSLint rules with ESLint.

That's what I understood as well.

What I meant is that since a compatibility package will be available, using TSLint isn't a big issue. It still receives updates and has some useful checks that are not available in ESLint yet.

That being said I'm not against your proposal of also adding an `useTypescriptPlugin` option :)

Kocal:
Hum, I'm thinking about something when reading this:

> if the dependency eslint-plugin-vue is installed, **we enable linting on Vue files**

> I think we could probably just tweak Encore.enableEslintLoader() (which currently only accepts 1 argument) to automatically **enable the plugin + vue/recommended** ...

In fact, how the user will be able to choose between every _sets_ of rules? (`base`, `essential`, `strongly-recommended`, and `recommended`)? What if the user does not want to use any preset but just some rules? :thinking: 

To be honest, I really don't think we can't/should do more than enable `eslint-loader` to be ran on `.vue` files (`test: /.(jsx?|vue)$/`). :confused:

masi:
> To be honest, I really don't think we can't/should do more than enable `eslint-loader` to be ran on `.vue` files (`test: /.(jsx?|vue)$/`). 😕

Does the Encore configure anything else for ESLint? AFAIC the only thing Encore does is to enable the eslint loader. So there isn't much to do than provide a nice interface to enable ESLint for other file types like .vue, .ts or whatever ESLint can handle. Though one could argue that adding the recommended rules for all supported files is the way to go. After all you can reconfigure it manually anyway.

Should be enough for .vue files. To lint typescript you need not only have the plugins @typescript-eslint but also the parser @typescript-eslint/parser enabled. Which makes it a bit more interesting to setup when you have both .js and .ts files in your sources. Unless @typescript-eslint/parser can handle plain JS without any issues. I have tried tried it out.

Kocal:
This can be closed since #574 has been merged.

PR #505, opened 2019-01-30
Migrate from yargs to yargs-parser for the CLI

stof:
The encore CLI is not implemented using the yargs features, as it is a wrapper around the webpack and webpack-dev-server. Only the parsing from yargs was used. And this is implemented by yargs-parser. Depending on that package directly avoids bring the full (outdated) yargs.


Lyrkan:
Hm... actually the method is called for all the tests of [`test/index.js`](https://github.com/symfony/webpack-encore/blob/801fc9194eadcf9eafe03d99ae6b8ac4b535aaf0/test/index.js#L17)... they probably pass because `yargs` is required by another dependency.

Kocal:
`sass-graph`, `uglify-js`, `webpack-cli` and `webpack-dev-server` depends of `yargs` (search for `yargs "` in `yarn.lock`). 

I don't think we can do something about it.

stof:
I fixed the place I missed.

Regarding tests, there is indeed nothing we can do here, due to the way yarn and npm are flattening the `node_modules` folder, making extra packages available.

Yarn's PnP mode may allow detecting such things (as it builds `no_modules` in a different way), but Encore is not compatible with it, due to the usage of optional peer dependencies (all these optional deps that we ask the main package to require), which are not compatible with PnP by design.

stof:
> `sass-graph`, `uglify-js`, `webpack-cli` and `webpack-dev-server` depends of `yargs` (search for `yargs "` in `yarn.lock`).

And all these were requiring *different* versions of yargs, so we were not sharing it with them previously.

stof:
Note that when using snyk.io on my project, it reports it being affected by https://app.snyk.io/vuln/npm:mem:20180117 due to 2 things:

```
Detailed paths and remediation

    Introduced through: incenteev@0.0.0 › @symfony/webpack-encore@0.22.4 › yargs@8.0.2 › os-locale@2.1.0 › mem@1.1.0
    Remediation: No remediation path available.

    Introduced through: incenteev@0.0.0 › @symfony/webpack-encore@0.22.4 › @babel/preset-env@7.3.1 › @babel/plugin-transform-named-capturing-groups-regex@7.3.0 › regexp-tree@0.1.0 › yargs@10.1.2 › os-locale@2.1.0 › mem@1.1.0
    Remediation: Your dependencies are out of date, otherwise you would be using a newer mem than mem@1.1.0. Try reinstalling your dependencies. If the problem persists, one of your dependencies may be bundling outdated modules.
```

This PR solves this `No remediation path available.` thing where webpack-encore would force using an old version of `mem` (by migrating to `yargs-parser`, we remove the transitive dependency on `mem` entirely).
Note that the second path also comes from webpack-encore deps, but this is already solved as `regexp-tree` released a 0.1.1 version a few hours ago which upgrades to yargs 12 (which uses `os-locale` 3.x which itself uses `mem` 4.0).

stof:
yeah, I know it is not a big issue (otherwise, I would have sent the PR weeks ago when I saw it for the first time)

PR #508, opened 2019-02-01
Support CSS modules in Vue.js projects by default

### LINKED ISSUE #460: CSS Modules no longer work
Hey guys!

Just updated encore to latest version and noticed that VueLoader css modules no longer are being generated (raw classnames are being generated instead and `$style.className` no longer returns appropriate class)...

Was there something changed that broke CSS modules or is this an error?

Thanks!

Lyrkan:
With the current version of Encore you are able to either:

* only use standard CSS (default behavior)
* only use CSS modules (by calling `Encore.configureCssLoader()`)

This PR should help detecting when `<style module>` is used and enabling css modules accordingly (fixes #460).

**Reference:** https://vue-loader.vuejs.org/guide/css-modules.html#opt-in-usage

### DIFF
--- a/fixtures/vuejs-css-modules/App.vue
+++ b/fixtures/vuejs-css-modules/App.vue
@@ -1,0 +1,15 @@
+<template>
+  <div id="app" class="red" :class="$style.italic"></div>
+</template>
+
+<style>
+  .red {
+    color: red;
+  }
+</style>
+
+<style module>
+  .italic {
+    font-style: italic;
+  }
+</style>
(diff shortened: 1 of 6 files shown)


weaverryan:
Thank you @Lyrkan!

Chesskingt:
Hi! Unfortunately this does not work with a language (scss) set as it used to... Can this support SCSS lang modules too?

Chesskingt:
Also, local ident name no longer can be configured manually...

Lyrkan:
Hmm...we probably have to use the same trick for all languages since they re-include the css loader. I'll se what I can do, unless I'm missing something it shouldn't be too hard to implement.

Chesskingt:
Thank you so much!

PR #517, opened 2019-02-10
Allow to use Dart Sass instead of Node Sass

### LINKED ISSUE #422: Dart Sass? (Sass implementation customization)
[`sass-loader`](https://github.com/webpack-contrib/sass-loader) supports to specify Sass implementation to use: `node-sass`/`dart-sass`/`sass`

```js
// ...
    {
        loader: "sass-loader",
        options: {
            implementation: require("dart-sass")
        }
    }
// ...
```

[If I understand things right](https://symfony.com/doc/current/frontend/encore/css-preprocessors.html), I can customize `sass-loader` config like this:

```js
Encore.enableSassLoader(options => {
	options.implementation = require('sass')
})
```

But if I uninstall `node-sass` and install [`sass`](https://www.npmjs.com/package/sass) instead, I got an error because `Encore.enableSassLoader()` and `Encore.addStyleEntry` are limited to `node-sass` only

Lyrkan:
This PR closes #422 by allowing `sass` (Dart Sass) as an alternative choice to `node-sass`.

---

The `sass-loader` allows to set the implementation that should be used to process Sass files. It currently accepts either `sass` or `node-sass` (default):

In theory this can already be used by setting the related option when calling `Encore.enableSassLoader()`, but Encore then throws an error asking you to install `node-sass`, even if you already have `sass` installed.

One way to prevent that could be to detect when `options.implementation` is set but in its next version the `sass-loader` will [automatically detect](https://github.com/webpack-contrib/sass-loader/commit/ff90dd62cdfc51b4ca6cd755a65d3e2adc51657a) which implementation is installed and use it accordingly.

So, instead of doing that, this PR allows to define alternative packages in our `feature.js` file, for instance:

```js
const features = {
    sass: {
        method: 'enableSassLoader()',
        packages: [
            { name: 'sass-loader', enforce_version: true },
            [{ name: 'node-sass' }, { name: 'sass' }] // Will allow both `node-sass` and `sass`
        ],
        description: 'load Sass files'
    },
    // ...
}
```

If neither `node-sass` or `sass` is available, the error message will display both choices, and recommend the first one in the `yarn add` advice:

image

Note that for now using Dart Sass still requires the `implementation` option to be set:

```js
Encore.enableSassLoader(options => {
    options.implementation = require('sass');
});
```

This shouldn't be a big issue since the default recommendation is still to use `node-sass`.


weaverryan:
Wow! That's a really excellent solution. Thanks for the research and nice feature - love it!

PR #518, opened 2019-02-10
Allow to set a custom context in copyFiles

### LINKED ISSUE #490: copyFiles doesn't works for files required in js
I got a weird behaviour with copyFiles.

I know in the docs, the copyFiles is used  fore images in templates, but, in my case, and I guess in many cases, I have images in CSS and in templates.

For example with this webpack config
```
.copyFiles({
        from: './assets/img',
        // optional target path, relative to the output dir
        //to: 'images/[path][name].[ext]',

        // if versioning is enabled, add the file hash too
        to: 'img/[path][name].[hash:8].[ext]',

        // only copy files matching this pattern
        pattern: /\.(png|jpg|jpeg|gif|ico|svg|webp)$/
    })
```

In this case, the images are copied into 2 folders.
The images from templates follow the `copyFiles` rules and end up in `public/build/img/[path]`
The images from CSS/JS are copied into `public/build/images/*.jpg|png|etc.`

and if I have the same images used in template and in CSS/JS, the images will be duplicated, one on each build sub folder.

It could be good if all images followed the same rules, no ?
Or I missed something in the docs ?

Edit : Sorry I misinterpreted something.
It seems copyFiles copy all files, not only those used in templates.
But for some reasons I also have images copied into an `images` folder and I don't understand what is causing this.
The images copied into the `images` folder are only images used in CSS (used in my CSS "framework", `fomantic-ui`).

Lyrkan:
The current behavior of `copyFiles()` is to use the value of the `from` option as a context when copying files.

This means that when doing the following:

```js
Encore.copyFiles({
  from: './assets/images',
  to: '[path][name].[ext]',
  includeSubdirectories: false,
});
```

Will end-up putting all the files from `./assets/images` directly at the root of the output directory.

Sometimes this is not the desired behavior (see https://github.com/symfony/webpack-encore/issues/490#issuecomment-453982520) which is why this PRs adds a `context` option to `copyFiles()` entries (closes #490).

For instance:

```js
Encore.copyFiles({
  from: './assets/images',
  to: '[path][name].[ext]',
  includeSubdirectories: false,
  context: './',
});
```

Will put all the copied files inside of an `./assets/images` folder in the output directory.

### DIFF
--- a/index.js
+++ b/index.js
@@ -493,7 +493,8 @@ class Encore {
      *          { from: './txt', pattern: /\.txt$/ },
      *      ]);
      *
-     *      // Set the context path
+     *      // Set the context path: files will be copied
+     *      // into an images/ directory in the output dir
      *      Encore.copyFiles({
      *          from: './assets/images',
      *          to: '[path][name].[hash:8].[ext]',
--- a/index.js
+++ b/index.js
@@ -493,6 +493,13 @@ class Encore {
      *          { from: './txt', pattern: /\.txt$/ },
      *      ]);
      *
+     *      // Set the context path
+     *      Encore.copyFiles({
+     *          from: './assets/images',
+     *          to: '[path][name].[hash:8].[ext]',
+     *          context: './assets'
+     *      });
+     *
      * Notes:
      *      * No transformation is applied to the copied files (for instance
      *        copying a CSS file won't minify it)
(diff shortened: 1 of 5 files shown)


weaverryan:
LOVE this - great improvement!

PR #520, opened 2019-02-11
Improve 'Install x to use y' and 'Unrecognized method' error messages.

### LINKED ISSUE #420: Shorter error when a package needs to be installed
If you try to enable a feature that requires some dependencies, you get a really nice error that tells you the command you need to run to get them. But... woh! The stack trace makes this pretty tough to see - you need to scroll up a bunch.

When we have *this* specific error, we should probably not print the stacktrace at all.

screen shot 2018-10-26 at 3 28 37 pm

Lyrkan:
This PR:

- Removes the stacktrace from the "_Install x to use y_" error message by throwing a string instead of an `Error` (fixes #420)
- Only keeps the second line of the stacktrace for the "_Unrecognized method_" error message (since this is most likely the one related to the `webpack.config.js` file)
- Fixes the "_Did you mean xxxx_" advice that was broken since the `Encore` object became a `class`

---

**Install x to use y**

Before: 

After:

---

**Unrecognized method**

Before:

After:

### DIFF
--- a/test/bin/encore.js
+++ b/test/bin/encore.js
@@ -193,4 +193,41 @@ module.exports = Encore.getWebpackConfig();
             done();
         });
     });
+
+    it('Display an error when calling an unknown method', (done) => {
+        testSetup.emptyTmpDir();
+        const testDir = testSetup.createTestAppDir();
+
+        fs.writeFileSync(
+            path.join(testDir, 'package.json'),
+            `{
+                "devDependencies": {
+                    "@symfony/webpack-encore": "*"
+                }
+            }`
+        );
+
+        fs.writeFileSync(
+            path.join(testDir, 'webpack.config.js'),
+            `
+const Encore = require('../../index.js');
+Encore
+    .setOutputPath('build/')
+    .setPublicPath('/build')
+    .enableSingleRuntimeChuck()
+    .addEntry('main', './js/no_require')
+;
+
+module.exports = Encore.getWebpackConfig();
+            `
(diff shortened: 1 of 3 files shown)


test/bin/encore.js
```diff
@@ -193,4 +193,41 @@ module.exports = Encore.getWebpackConfig();
...
+        );
+
+        const binPath = path.resolve(__dirname, '../', '../', 'bin', 'encore.js');
+        exec(`node ${binPath} dev --context=${testDir}`, { cwd: testDir }, (err, stdout, stderr) => {
+            expect(err).not.to.be.null;
+            expect(stdout).to.contain('is not a recognized property or method');
+            expect(stdout).to.contain('Did you mean');
+            done();
+        });
+    });
```
weaverryan: ❤️ for this test

Kocal:
> image

Given your screenshot, actually we don't need `vue` but `vue-loader`, no?

weaverryan:
Oh, you fixed that stacktrace that was bothering me for the longest time. And more stuff. Thank you @Lyrkan!

Lyrkan:
@Kocal When you call `enableVueLoader()` you actually need 3 dependencies:

https://github.com/symfony/webpack-encore/blob/677c434e2e6d9d798b90a8b3d3bb789593155486/lib/features.js#L81-L89

For my screenshot I only removed `vue` from an existing project, so the error message only displayed that one.

PR #521, opened 2019-02-12
Allow to set @babel/preset-env's useBuiltIns option

### LINKED ISSUE #444: Feature idea: enablePolyfills() to add usage-based polyfills
With the following simple code, it is possible to modify Encore's babel config to provide automatic usage-based polyfills for stuff like `Promise, String.prototype.includes, Map, Set` through [@babel/polyfill](https://babeljs.io/docs/en/babel-polyfill) and core-js:
```bash
yarn add --dev @babel/polyfill
```
```javascript
// webpack.config.js
.configureBabel(function (babelConfig) {
    const preset = babelConfig.presets.find(([name]) => name === "@babel/preset-env");
    if (preset !== undefined) {
        preset[1].useBuiltIns = "usage";
    }
})
```
Polyfills are added based on 1. detected usage in your code and 2. your browserslist query in package.json. I think many users would find this convenient through an encore config option such as `enablePolyfills()` or whatever name seems more suitable.

Lyrkan:
This PR adds an `useBuiltIns` option to the second parameter of `Encore.configureBabel()` that sets the same option on the`@babel/preset-env` preset.

That option allows to change how the preset handles polyfills (fixes #444), for instance:

```js
Encore.configureBabel(() => {}, {
    // automatically import individual polyfills
    // where they are needed
    useBuiltIns: 'usage';
});
```

The PR also depreciates the `include_node_modules` option and replaces it by `includeNodeModules` since other methods (for instance `Encore.enableSassLoader()`) use camelCased names.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -81,7 +81,8 @@ class WebpackConfig {
             fonts: false
         };
         this.babelOptions = {
-            exclude: /(node_modules|bower_components)/
+            exclude: /(node_modules|bower_components)/,
+            useBuiltIns: 'entry',
         };
 
         // Features/Loaders options callbacks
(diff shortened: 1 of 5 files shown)


lib/WebpackConfig.js
```diff
@@ -326,14 +327,20 @@ class WebpackConfig {
 
         this.babelConfigurationCallback = callback;
 
-        for (const optionKey of Object.keys(options)) {
+        for (let optionKey of Object.keys(options)) {
```
stof: this could stay a const. You never re-assign it.

Lyrkan: Yep, I was re-assigning it in a previous version, forgot to change it back.

weaverryan:
Excellent compromise on making this feature more obvious. Thanks @Lyrkan!

PR #522, opened 2019-02-13
Add an enableIntegrityHashes() method to the public API

### LINKED ISSUE #418: Include JavaScript file hashes in entrypoints.json for hash & nonce attributes (CSP)

Lyrkan:
This PR adds an `Encore.enableIntegrityHashes()` method that allows to compute integrity hashes of all the files referenced in the `entrypoints.json` (closes #418).

For instance:

```js
// Enable it for all builds with the
// default hash algorithm (sha384)
Encore.enableIntegrityHashes();

// Enable it only in production with
// a custom hash algorithm
Encore.enableIntegrityHashes(
    Encore.isProduction(),
    'sha512'
);
```

And here is the resulting `entrypoints.json` file:
```json
{
  "entrypoints": {
    "main": {
      "css": [
        "/main.3d1dcb7e.css"
      ],
      "js": [
        "/main.7c6b7c81.js"
      ]
    }
  },
  "integrity": {
    "/main.3d1dcb7e.css": "sha384-ce7d1nV3CFoSIfinwm53befb9CMHNAAlPEb61deOf3zBvpXK9lct44/U2ieSOKt4",
    "/main.7c6b7c81.js": "sha384-kHFhNTJlbSuDijSTimOHZGTxKzlLYCWc03AmmRSAE3b173SMlGiQG6uasAzl29+0"
  }
}
```

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -11,6 +11,7 @@
 
 const path = require('path');
 const fs = require('fs');
+const crypto = require('crypto');
 const logger = require('./logger');
 
 /**
@@ -48,6 +49,7 @@ class WebpackConfig {
         this.configuredFilenames = {};
         this.aliases = {};
         this.externals = [];
+        this.integrityAlgorithms = [];
 
         // Features/Loaders flags
         this.useVersioning = false;
@@ -748,6 +750,25 @@ class WebpackConfig {
         }
 
         this.loaderConfigurationCallbacks[name] = callback;
+    }
+
+    enableIntegrityHashes(enabled = true, algorithms = ['sha384']) {
+        if (!Array.isArray(algorithms)) {
+            algorithms = [algorithms];
+        }
+
+        const availableHashes = crypto.getHashes();
+        for (const algorithm of algorithms) {
+            if (typeof algorithm !== 'string') {
+                throw new Error('Argument 2 to enableIntegrityHashes() must be a string or an array of strings.');
+            }
+
+            if (!availableHashes.includes(algorithm)) {
+                throw new Error(`Invalid hash algorithm "${algorithm}" passed to enableIntegrityHashes().`);
+            }
+        }
+
+        this.integrityAlgorithms = enabled ? algorithms : [];
     }
 
     useDevServer() {
--- a/lib/plugins/entry-files-manifest.js
+++ b/lib/plugins/entry-files-manifest.js
@@ -13,32 +13,71 @@ const PluginPriorities = require('./plugin-priorities');
 const sharedEntryTmpName = require('../utils/sharedEntryTmpName');
 const copyEntryTmpName = require('../utils/copyEntryTmpName');
 const AssetsPlugin = require('assets-webpack-plugin');
+const fs = require('fs');
+const path = require('path');
+const crypto = require('crypto');
 
-function processOutput(assets) {
-    for (const entry of [copyEntryTmpName, sharedEntryTmpName]) {
-        delete assets[entry];
-    }
+function processOutput(webpackConfig) {
+    return (assets) => {
+        for (const entry of [copyEntryTmpName, sharedEntryTmpName]) {
+            delete assets[entry];
+        }
 
-    // with --watch or dev-server, subsequent calls will include
-    // the original assets (so, assets.entrypoints) + the new
-    // assets (which will have their original structure). We
-    // delete the entrypoints key, and then process the new assets
-    // like normal below
-    delete assets.entrypoints;
+        // with --watch or dev-server, subsequent calls will include
+        // the original assets (so, assets.entrypoints) + the new
+        // assets (which will have their original structure). We
+        // delete the entrypoints key, and then process the new assets
+        // like normal below
+        delete assets.entrypoints;
 
-    // This will iterate over all the entry points and convert the
-    // one file entries into an array of one entry since that was how the entry point file was before this change.
-    for (const asset in assets) {
-        for (const fileType in assets[asset]) {
-            if (!Array.isArray(assets[asset][fileType])) {
-                assets[asset][fileType] = [assets[asset][fileType]];
+        // This will iterate over all the entry points and convert the
+        // one file entries into an array of one entry since that was how the entry point file was before this change.
+        const integrity = {};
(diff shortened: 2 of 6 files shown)


Lyrkan:
Slightly changed how the `integrity` part of the `entrypoints.json` looks, from:

```json
{
    "integrity": {
        "file1" : { "algorithm": "<alg>", "hash": "<hash>" },
        "file2" : { "algorithm": "<alg>", "hash": "<hash>" }
    }
}
```

to:

```json
{
    "integrity": {
        "algorithm": "<alg>",
        "hashes": {
            "file1" : "<hash>",
            "file2" : "<hash>"
        }
    }
}
```

I thought the first version could be useful (for instance if we added a way to use a different algorithm for each file) but after thinking about it for a while I'm not sure that's really needed.

lib/plugins/entry-files-manifest.js
```diff
@@ -13,32 +13,66 @@ const PluginPriorities = require('./plugin-priorities');
...
+
+        for (const asset in assets) {
+            for (const fileType in assets[asset]) {
+                if (!Array.isArray(assets[asset][fileType])) {
+                    assets[asset][fileType] = [assets[asset][fileType]];
+                }
+
+                if (webpackConfig.integrityAlgorithm) {
+                    for (const file of assets[asset][fileType]) {
+                        if (!(file in integrity.hashes)) {
```
weaverryan: Is file sometimes already in `integrity.hashes`? Like, maybe a shared file that's used by multiple entries?

Also, we can flip this to `if (file in integrity.hashes) { continue; }

lib/plugins/entry-files-manifest.js
```diff
@@ -13,32 +13,66 @@ const PluginPriorities = require('./plugin-priorities');
...
             }
         }
-    }
 
-    return JSON.stringify({
-        entrypoints: assets
-    }, null, 2);
+        return JSON.stringify({
+            entrypoints: assets,
+            integrity
```
weaverryan: Should we include the `integrity` key at all if the feature is disabled?

lib/plugins/entry-files-manifest.js
Lyrkan: > Is file sometimes already in `integrity.hashes`? Like, maybe a shared file that's used by multiple entries?

Yep, the `runtime.js` file for instance :)

> Also, we can flip this to `if (file in integrity.hashes) { continue; }

Will be part of the next changeset!

lib/plugins/entry-files-manifest.js
Lyrkan: The idea was to be able to detect if Encore supported that feature to make the bundle display a relevant error if someone tried to use it with an old version of Encore.

But since I also agree about not making it optional in the bundle anymore (by detecting if `integrity` is present in the manifest) that's not needed anymore :)

Lyrkan:
Ok so, changed the content again to something like:

```js
{
    "integrity": {
        "file1" : "<alg>-<hash1>",
        "file2" : "<alg>-<hash2>"
    }
}
```

The reason being that it now allows to generate multiple hashes for each file (which is a feature supported by the `integrity` attribute):

```js
{
    "integrity": {
        "file1" : "<alg1>-<hash1> <alg2>-<hash2>",
        "file2" : "<alg1>-<hash3> <alg2>-<hash4>"
    }
}
```

Which also means that the `Encore.enableIntegrityHashes()` can now take an array of strings as its second parameter, for instance:

```js
Encore.enableIntegrityHashes(
    true,
    ['sha256', 'sha384', 'sha512']
);
```

weaverryan:
Needs a rebase now :)

weaverryan:
This is going to be awesome! Thank you @Lyrkan!

Phobetor:
Epic! This will be really useful. Thanks @Lyrkan!

lib/WebpackConfig.js
```diff
@@ -750,6 +752,25 @@ class WebpackConfig {
         this.loaderConfigurationCallbacks[name] = callback;
     }
 
+    enableIntegrityHashes(enabled = true, algorithms = ['sha384']) {
```
stof: should we also forbid passing `enabled = true` with an empty array as second argument ? That would actually disable it.

PR #523, opened 2019-02-15
Ignore hot-update files in the SharedEntryConcatPlugin

### LINKED ISSUE #493: dev-server hot: createSharedEntry with sass fails if enableSourceMaps is enabled
When running **dev-server --hot** and having **enableSourceMaps** enabled it fails whenever I update a .scss file. It looks like it is stuck at `[WAIT] Compiling`. After I changed webpack `devServer.quiet = false`, I saw that I fails with an error:

`Error: Invalid number of files for chunk vendor - got vendor.js, vendor.469d3207de4d53a8a23b.hot-update.js`

in node_modules/@symfony/webpack-encore/lib/webpack/shared-entry-concat-plugin.js line 37

At first compile of dev-server --hot everything seems fine, when I only change something in a .js file as well.

Only when I make a change in any .scss file, the var **jsFiles** (in shared-entry-concat-plugin.js line 26) seems to contain two files. A **vendor.js** and a **vendor.xxxxx.hot-update.js**. When I disable **enableSourceMaps** in Encore the latter file isn't created and the error disappears (but so does the source maps functionality)

```
// webpack.config.js

Encore
    .setPublicPath('/build')
    .setManifestKeyPrefix('build/')

    .createSharedEntry('vendor', './assets/js/webpack.shared_entry.js')

    .enableSingleRuntimeChunk()

    .enableSassLoader()
;

```

```
// assets/js/webpack.shared_entry.js

require('../css/app.scss');
```

_@symfony/webpack-encore 0.22.4
symfony/webpack-encore-bundle 1.0.0_

Lyrkan:
This PR fixes #493.

When using `dev-server --hot` an additional `.hot-update.js` file can be [added to the list of files associated to a chunk](https://github.com/webpack/webpack/blob/7076c05353bae6d788247cfcaa94704f34cca292/lib/HotModuleReplacementPlugin.js#L287). If that list contains another JS file the current version of Encore throws an Error that causes the compilation to fail.

The good news is that the `.hot-update.js` file is also [added to the `additionalChunkAssets`](https://github.com/webpack/webpack/blob/7076c05353bae6d788247cfcaa94704f34cca292/lib/HotModuleReplacementPlugin.js#L284) property of the `compilation` object. This allows us to ignore it and only keep the JS file that we need.

### DIFF
--- a/lib/webpack/shared-entry-concat-plugin.js
+++ b/lib/webpack/shared-entry-concat-plugin.js
@@ -18,13 +18,14 @@ function SharedEntryConcatPlugin(sharedEntryName) {
 
 function getChunkFilename(compilation, chunkName) {
     const chunk = compilation.namedChunks.get(chunkName);
+    const additionalChunkAssets = compilation.additionalChunkAssets || [];
 
     if (!chunk) {
         throw new Error(`Cannot find chunk ${chunkName}`);
     }
 
     const jsFiles = chunk.files.filter(filename => {
-        return /\.js$/.test(filename);
+        return /\.js$/.test(filename) && !additionalChunkAssets.includes(filename);
     });
 
     if (jsFiles.length !== 1) {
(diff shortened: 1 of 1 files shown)


weaverryan:
Thank you @Lyrkan!

PR #525, opened 2019-02-16
Allow setOutputPath to create nested directories

### LINKED ISSUE #189: Create the output directory if it does not exist
Currently, we do create the output directory if it does not exist... but we only go one level deep (if `web/` exists, we will create `web/build`, but if `web` does not exist, we throw an error): https://github.com/symfony/webpack-encore/blob/799720602aab2069f827704e4a6d672d03d38d76/lib/WebpackConfig.js#L107

We do that for safety, but in this (rare) situation, it's kind of annoying. My friend Beau hit it recently with Encore on Sculpin :).

Lyrkan:
This PR changes the check that is currently made when calling `setOutputPath` in order to allow the creation of nested directories (closes #189).

Note that it doesn't remove it entirely since it still prevents the creation of more than one level when the output path is located outside of the context path.

For instance, given the context is `/home/foo/bar/`:

```js
// Will work
Encore.setOutputPath('build');

// Will work, even if "/home/foo/bar/public" does not exist
Encore.setOutputPath('public/build/output');

// Will work, even if "/home/foo/public" does not exist
Encore.setOutputPath('../public');

// Will work if only "/home/foo/public/build" does not exist
// ... but **won't** work if "/home/foo/public" does not exist
Encore.setOutputPath('../public/build');
```

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -129,11 +129,23 @@ class WebpackConfig {
         }
 
         if (!fs.existsSync(outputPath)) {
-            // for safety, we won't recursively create directories
-            // this might be a sign that the user has specified
-            // an incorrect path
-            if (!fs.existsSync(path.dirname(outputPath))) {
-                throw new Error(`outputPath directory does not exist: ${outputPath}. Please check the path you're passing to setOutputPath() or create this directory`);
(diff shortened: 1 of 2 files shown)


weaverryan:
Good balance here +1

PR #529, opened 2019-02-24
Honor --public flag given to 'yarn run dev-server ...'

lyrixx:
I'm using docker (one container for the front, another one for webpack)
and I had to do this PR to make it work.

For reference, this is the command I (docker actually) executed
```
yarn run dev-server --host 0.0.0.0 --port 9999 --hot --public https://example.com:443 --disable-host-check
```

### DIFF
--- a/lib/config/parse-runtime.js
+++ b/lib/config/parse-runtime.js
@@ -49,9 +49,19 @@ module.exports = function(argv, cwd) {
             runtimeConfig.useHotModuleReplacement = argv.hot || false;
             runtimeConfig.devServerKeepPublicPath = argv.keepPublicPath || false;
 
-            var host = argv.host ? argv.host : 'localhost';
-            var port = argv.port ? argv.port : '8080';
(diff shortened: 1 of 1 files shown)


lyrixx:
Thanks for the review. I updated the PR. Is it better?

lib/config/parse-runtime.js
```diff
@@ -49,9 +49,19 @@ module.exports = function(argv, cwd) {
             runtimeConfig.useHotModuleReplacement = argv.hot || false;
             runtimeConfig.devServerKeepPublicPath = argv.keepPublicPath || false;
 
-            var host = argv.host ? argv.host : 'localhost';
-            var port = argv.port ? argv.port : '8080';
-            runtimeConfig.devServerUrl = `http${runtimeConfig.devServerHttps ? 's' : ''}://${host}:${port}/`;
+            if (typeof argv.public !== 'undefined') {
```
Lyrkan:
```suggestion
            if (typeof argv.public === 'string') {
```

weaverryan:
Nice one @lyrixx! Thank you!

PR #530, opened 2019-02-28
Add some checks to the 'pattern' option of Encore.copyFiles()

Lyrkan:
While reviewing https://github.com/FriendsOfSymfony/FOSCKEditorBundle/pull/174 I was a bit surprised that calling `copyFiles` with a string as a `pattern` (instead of a `RegExp`) worked fine.

It turns out that's not an issue since the `copyFiles` method causes the creation of a temporary `.js` file in which the `RegExp` is casted to a string anyway.

However, what *is* an issue is that using a string that does not contain a valid regular expression can break that temporary file.

For instance:

```js
Encore.copyFiles({
  from: './assets',
  pattern: 'foo;'
});
```

Will generate something like:

```js
const context_0 = require.context(
    '!/* ... */',
    true,
    foo; // <= Syntax error here
);

context_0.keys().forEach(context_0);
```

This PR doesn't prevent using strings for that option because I don't think that would be the right choice (the unquoted `/regexp/` syntax may not be obvious for JS beginners) but instead officially allow them with some extra checks.

It also casts `includeSubdirectories` (2nd parameter of the `require.context` above) to a boolean when creating the JS file to avoid the same kind of syntax error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -500,8 +500,9 @@ class Encore {
      * Supported options:
      *      * {string} from (mandatory)
      *              The path of the source directory (mandatory)
-     *      * {RegExp} pattern (default: all files)
-     *              A pattern that the filenames must match in order to be copied
+     *      * {RegExp|string} pattern (default: all files)
+     *              A regular expression (or a string containing one) that
+     *              the filenames must match in order to be copied
      *      * {string} to (default: [path][name].[ext])
      *              Where the files must be copied to. You can add all the
      *              placeholders supported by the file-loader.
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -451,6 +451,20 @@ class WebpackConfig {
                 }
             }
 
+            if (typeof config.pattern !== 'undefined' && !(config.pattern instanceof RegExp)) {
+                let validPattern = false;
+                if (typeof config.pattern === 'string') {
+                    const regexPattern = /^\/(.*)\/([a-z]*)?$/;
+                    if (regexPattern.test(config.pattern)) {
+                        validPattern = true;
+                    }
+                }
+
+                if (!validPattern) {
+                    throw new Error(`Invalid pattern "${config.pattern}" passed to copyFiles(). Make sure it contains a valid regular expression.`);
+                }
+            }
+
             this.copyFilesConfigs.push(
                 Object.assign({}, defaultConfig, config)
             );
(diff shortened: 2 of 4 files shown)


Lyrkan:
Seems like the failing job is caused by the latest version of the `assets-webpack-plugin`, and is not related to this PR.

Nothing is shown because we [mute Webpack's output](https://github.com/symfony/webpack-encore/blob/853e0f97a20ee664143152078fd285f0cfd27bc1/test/helpers/setup.js#L67-L70) during functional tests, and in some cases `compiler.run(handler)` fails without calling `handler` (which is supposed to unmut stdout)... maybe we could add a timeout that resets stdout after a certain amount of time if `handler` hasn't been called.

Anyway, the real stacktrace is:

```
Uncaught TypeError: Cannot read property 'integrity' of undefined
  at node_modules/assets-webpack-plugin/dist/index.js:97:55
  at Array.reduce (<anonymous>)
  at node_modules/assets-webpack-plugin/dist/index.js:87:30
  at Array.reduce (<anonymous>)
  at afterEmit (node_modules/assets-webpack-plugin/dist/index.js:74:27)
  at AsyncSeriesHook.eval [as callAsync] (eval at create (node_modules/tapable/lib/HookCodeFactory.js:32:10), <anonymous>:7:1)
  at AsyncSeriesHook.lazyCompileHook (node_modules/tapable/lib/Hook.js:154:20)
  at asyncLib.forEachLimit.err (node_modules/webpack/lib/Compiler.js:432:27)
  at node_modules/neo-async/async.js:2813:7
  at done (node_modules/neo-async/async.js:3521:9)
  at node_modules/graceful-fs/graceful-fs.js:45:10
  at FSReqWrap.oncomplete (fs.js:141:20)
```

So it looks like `assets-webpack-plugin` tried to [add its own subresource-integrity support](https://github.com/ztoben/assets-webpack-plugin/commit/d69ac3ca027511f4228e0ef3a3984b86a038cddf) using [`webpack-subresource-integrity`](https://github.com/waysact/webpack-subresource-integrity) but did something that isn't compatible with what we do in the "createdSharedEntry() creates commons files" test case.

Kocal:
Maybe we should let V8 itself check if the regex is valid no?
With a `try/catch` around a `new Regex(thePattern)`?

Lyrkan:
@Kocal Thought about doing that... but it sadly won't work since `/regexp/` does not mean the same thing than `new RegExp('/regexp/')`.

The `RegExp` constructor expects the first parameter to be the content of the regular expression (i.e. without the beginning and trailing `/`) and the flags as a second parameter.

Kocal:
Ah yes, I forgot about that "rule". 😞

weaverryan:
Thank you @Lyrkan!

PR #536, opened 2019-03-08
Fix JSON files support in Encore.copyFiles()

### LINKED ISSUE #535: Encore.copyFiles() doesn't work for JSON files
As reported in https://github.com/FriendsOfSymfony/FOSCKEditorBundle/pull/177#issuecomment-470941465, calling `Encore.copyFiles()` with a path/pattern that includes JSON files triggers the following error:

```
Module parse failed: Unexpected token m in JSON at position 0 while parsing near 'module.exports = __w...'
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token m in JSON at position 0 while parsing near 'module.exports = __w...'
```

This is caused by how JSON imports are now handled in Webpack 4, which removed the need of using the `json-loader` in most common cases but also prevents from using custom inline loaders for the `.json` extension.

Since `Encore.copyFiles()` internally uses `!file-loader?<params>` in `require.context()` calls, the output of the `file-loader` (which contains JS) is passed to that part of Webpack that processes JSON (which then throws the error).

From what I've read the only way to currently bypass that behavior is to set `rule.type`... which isn't possible from an inline call.

References:
* https://github.com/webpack/webpack/issues/6572
* https://github.com/webpack-contrib/file-loader/issues/259
* https://github.com/webpack/webpack/issues/7057
* https://github.com/webpack/webpack/issues/6586

Lyrkan:
This PR fixes #535.

As explained in the issue, the current version doesn't allow `Encore.copyFiles()` to process JSON files because of how Webpack 4 handles them.

Basically, from what I understand, for most of the files it works that way:

```
Imported file (module.type = javascript/auto)
  => file-loader (takes anything, returns JS content)
  => Webpack JS parser
```

But for `.json` files the JS parser is replaced by a JSON parser:

```
Imported file (module.type = json)
  => file-loader (takes anything, returns JS content)
  => Webpack JSON parser
```

Since that "Webpack JSON parser" part expects JSON data (and not something that contains JS), the compilation fails at that point.

There is no way to disable that behavior when you specify loaders the "inline" way... which is what we do for `Encore.copyFiles()`, you can only do it by defining a config rule that forces the `javascript/auto` type.

What the PR does: it adds another loader in front of the file-loader that returns the input it received but after setting its `_module.type` (and the generator/parser that go with it) to `javascript/auto`.

So in the end we get:

```
Imported file (module.type = json)
  => file-loader (takes anything, returns JS content)
  => copy-files-loader (sets type to javascript/auto and returns the same content)
  => Webpack JS parser
```

### DIFF
--- a/fixtures/copy/foo.css
+++ b/fixtures/copy/foo.css
@@ -1,0 +1,1 @@
+This is an invalid content to check that the file is still copied
--- a/fixtures/copy/foo.js
+++ b/fixtures/copy/foo.js
@@ -1,0 +1,1 @@
+This is an invalid content to check that the file is still copied
--- a/fixtures/copy/foo.json
+++ b/fixtures/copy/foo.json
@@ -1,0 +1,1 @@
+This is an invalid content to check that the file is still copied
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -183,8 +183,10 @@ class ConfigGenerator {
                         copyTo = this.webpackConfig.useVersioning ? '[path][name].[hash:8].[ext]' : '[path][name].[ext]';
                     }
 
+                    const copyFilesLoader = require.resolve('./webpack/copy-files-loader');
+                    const fileLoader = require.resolve('file-loader');
                     const copyContext = entry.context ? path.resolve(this.webpackConfig.getContext(), entry.context) : copyFrom;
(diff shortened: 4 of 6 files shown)


weaverryan:
I don't love this one, but, we're well-covered in tests, so let's see if it hopefully fixes things without side effects. Great job getting to the bottom of it!

Issue #527, opened 2019-02-21
Question: deny extract css from entry js


slince:
Hello ,  i dont want to encore extract css into a separate file，Is this possible?

Lyrkan:
Hi @slince,

Could you explain a bit more what you are trying to achieve ?
Do you want Webpack to add `<style>` tags instead ? Or maybe use CSS modules?

slince:
@Lyrkan   Yes, i just want webpack add style tag. Because css is very simple, it is not necessary to send an http request to download it.

Lyrkan:
I see, so you'd need to replace the `mini-css-extract-plugin` loader by the `style-loader`.

The bad news is: there is nothing built in the current version of Encore to do that... it may become a bit easier when #509 is released.

However you should be able to do it manually:

```js
// webpack.config.js
const Encore = require('@symfony/webpack-encore');
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

Encore
  // (...)
;

// Retrieve the generated config
const config = Encore.getWebpackConfig();

// Replace the mini-css-extract-plugin loader by style-loader
for (const rule of config.module.rules) {
  if (rule.use && rule.use[0] === MiniCssExtractPlugin.loader) {
    rule.use[0] = 'style-loader';
  }
}

// Make sure you export `config` and not
// `Encore.getWebpackConfig()` here
module.exports = config;
```

Kocal:
<sup>(I can't wait for #509 to be released :eyes:)</sup>

slince:
TKS

Issue #348, opened 2018-07-16
can not reload css when scss changed in dev server hot


lichnow:
in dev server environment sass file also be extract to a css file which imported in main.js.so , it can not been hoted in dev server. can use only `sass-loader`,`style-loader` to solve it in dev server enviroment in encore,or let us to configure it self by add a method `isDevServer` to check if in dev-server?

weaverryan:
Yea, that's correct - we *always* extract to a CSS file, which kills HMR. We did this on purpose - because we wanted to consistently output CSS files (not have different behavior between prod & dev environments). I've always wanted to add HMR for extracted CSS files, but it's always been a "todo" in the core webpack libs. For example, HMR is listed as a "TODO" on the https://github.com/webpack-contrib/mini-css-extract-plugin that we're using in the next version.

One option is to add a method to "opt in" to *not* extracting to CSS when in `dev-server` mode (and just use the style loader). That way, for people who know what they're doing and want HMR, they can opt into this.

Tech-Nomad:
Hey Ryan,
what do you mean by "One option is to add a method to "opt in" to not extracting to CSS when in dev-server mode (and just use the style loader). That way, for people who know what they're doing and want HMR, they can opt into this."? 

Is this anywhere documented? Where I have to set this option?

Tech-Nomad:
Why does this package not works with Encore?
https://github.com/shepherdwind/css-hot-loader

Lyrkan:
@Tech-Nomad What @weaverryan meant is that a method could be added to Encore to allow people not to use the `extract-text-webpack-plugin`/`mini-css-extract-plugin` when using `--dev-server`. That method doesn't exist yet.

AFAIK you should be able to use `css-hot-loader` with Encore, but:
* There's no easy way to add it (see the comments in #3... which probably won't work if you use the Webpack 4 version of Encore)
* You'll have to make sure that the versioning is not enabled when using it since the plugin requires that the filenames always stay the same

Tech-Nomad:
Thanks for the info Lyrkan! 
This feature is very important for my development workflow. So maybe it's less stress and work just to use original webpack in this case?

beany-vu:
I do agree with the @Tech-Nomad's opinion, the "live reload" feature or HMR should be available.

AubreyHewes:
FYI, Until it is an option in my projects I add the following to the `webpack.config.js` to enable hot style reloading in development. I accept the fact that it is not production css.

````javascript

...

[...]

Issue #256, opened 2018-02-05
How can we test our js components ?


Rebolon:
Hi,
I build vuejs or react components and i need to be able to run tests and i'm wondering how to do unit/e2e tests when we use `encore`. I don't find anythig about this in the documentation. 
Any idea ? 
thanks a lot

Lyrkan:
Hey @Rebolon,

Here is a small example I made a while ago that uses Karma/Jasmine: https://github.com/Lyrkan/encore-typescript-karma

If you look at the `karma.conf.js` file you'll see that you need to call `Encore.configureRuntimeEnvironment` before requiring your `webpack.config.js` file in order for Encore to know about which environment you want to use during your tests (you'll get an error if you forget to do that).

weaverryan:
Hey @Lyrkan!

This is a great answer. I think we should add something to the docs to help people. Wdyt?

Lyrkan:
Hi Ryan,

I thought about it when I was toying around with karma but wasn't really happy with some of the tweaks needed to make it work.

For instance, without that part the ManifestPlugin tries to create its files using an invalid path... still not sure why:

```js
// Set writeToFileEmit option of the ManifestPlugin to false
for (const plugin of webpackConfig.plugins) {
  if ((plugin instanceof ManifestPlugin) && plugin.opts) {
    plugin.opts.writeToFileEmit = false;
  }
}
```

Or the fact that you are supposed to remove the entry point since Karma handles it itself:

```js
// Remove entry property (handled by Karma)
delete webpackConfig.entry;
```

Rebolon:
Thanks for the answer @Lyrkan, i'll test it quickly.

Rebolon:
i've tried and get the following error:

```
{
    "message": "Uncaught ReferenceError: webpackJsonp is not defined\nat assets/js/vuejs/tests/index.js:1:1\n\nReferenceError: webpackJsonp is not defined\n    at assets/js/vuejs/tests/index.js:1:1",
    "str": "Uncaught ReferenceError: webpackJsonp is not defined\nat assets/js/vuejs/tests/index.js:1:1\n\nReferenceError: webpackJsonp is not defined\n    at assets/js/vuejs/tests/index.js:1:1"
  }
```

I've pushed the code on this branch: https://github.com/Rebolon/php-sf-flex-webpack-encore-vuejs/tree/feature/test-with-karma
But i think i've followed  your sample

Lyrkan:
[...]

PR #539, opened 2019-03-11
Add Encore.disableCssExtraction() to the public API

Lyrkan:
This PR adds an `Encore.disableCssExtraction()` method that allows to disable the `mini-css-extract-plugin` and use the `style-loader` instead.

It can be used to solve various problems that, until now, required a really ugly workaround that relied on our internal implementation (for instance the following commit probably broke some builds that used previous versions of it: https://github.com/symfony/webpack-encore/commit/68674433f9feb94e34c473c4b9cf5d7c40df515d#diff-8beacd21a12ca072bafa4e8e3f1aae6b).

Related issues: #3, #256, #348, #527

### DIFF
--- a/fixtures/js/css_import.js
+++ b/fixtures/js/css_import.js
@@ -1,0 +1,2 @@
+
+require('./../css/h1_style.css');
--- a/index.js
+++ b/index.js
@@ -1047,6 +1047,22 @@ class Encore {
     }
 
     /**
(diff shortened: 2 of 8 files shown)


weaverryan:
Yea, this IS useful. I fully realize that people may want to opt-in to the "inconsistent" experience between dev & prod because of nice things like HMR. Good idea!

PR #540, opened 2019-03-15
Update various dependencies and remove the LoaderOptionsPlugin

### LINKED ISSUE #428: resolve-url-loader magic behavior
When using encore for a project I was surprised to have the relative url(...) image references in my sass files resolved correctly even though the relative path were wrong (I moved the sass files but not the images). 

Turns out that the resolve-url-loader package will pretty much search your whole project directory for the image filenames and use anything that matches (starting with the correct relative path but if it does not resolve it will widen the search until it finds package.json and then do a recursive file search).

This is not only slow but also confusing and promotes the (accidental) use of erroneous paths in CSS files. The author of resolve-url-loader also [acknowledges](https://github.com/bholloway/resolve-url-loader/blob/master/packages/resolve-url-loader-filesearch/README.md) this: 

> The file-search "magic" was always a hack and solved problems back in the days when .css was distributed by Bower. Back then it was common for packages to be broken.
> Now days bundling CSS is common place and broken packages are quickly fixed. It should be possible to solve your use-case without a full file search.
> The author recommends you use the join option in resolve-url-loader@>=3.0.0 to make a more targeted fix.

Would you consider upgrading the resolve-url-loader dependency to ^3.0?

### LINKED ISSUE #534: How to exclude dirs from sass loader?
Hi. I have dir like /uploads/ where user store their photos. When running webpack with enableSassLoader() it makes stat through all this files (which take long time). The only solution for now is to set "resolveUrlLoader: false" but I want to use it for simplier paths. So is there any solution to disable /uploads/ dir?

Lyrkan:
This PR updates various dependencies that were either really outdated or subject to security issues.

Since most of those packages are listed in our dev dependencies it shouldn't have a big impact on most users, but the following changes could be an issue for some of them:

* The `css-loader` was updated from `^1.0.0` to `^2.1.1` ([breaking changes](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#200-2018-12-07))
* The `resolve-url-loader` was updated from `^2.3.0` to `^3.0.1` ([breaking changes](https://github.com/bholloway/resolve-url-loader/tree/master/packages/resolve-url-loader#version-3), closes #428 and closes #534)
* The minimum version of NodeJS was bumped from 6 to 8 (needed to update `zombie`, and v6's end-of-life happening in less than a month anyway)
* The [`LoaderOptionsPlugin`](https://webpack.js.org/plugins/loader-options-plugin/) was removed since it should not be needed anymore and caused a *really annoying* bug when used with the latest version of the `resolve-url-loader` (see https://github.com/symfony/webpack-encore/issues/428#issuecomment-473445008)


weaverryan:
Great job and thanks for summarizing the important changes. I'll include in the release.

PR #541, opened 2019-03-18
Fix the color of error messages when they are displayed on a light theme

### LINKED ISSUE #533: Color issue with "Install ... to use ..." feature
Ref: #520

If we have a light theme, the word `Install` is unreadable:
selection_301

It works fine with dark theme:
selection_302

It would be nice if it take the same color than `Running webpack ...`, `yarn dev-`, `to use`... which is the default color.

I quickly checked the source code on GitHub but I wasn't able to find where the problem apprears.

Thanks!

Lyrkan:
This PR fixes #533 by slightly modifying the default theme of `pretty-error` so it doesn't use a color for the error message.

**Before:**

image

**After:**

image

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@
 const WebpackConfig = require('./lib/WebpackConfig');
 const configGenerator = require('./lib/config-generator');
 const validator = require('./lib/config/validator');
-const PrettyError = require('pretty-error');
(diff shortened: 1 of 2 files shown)


weaverryan:
Nice one - thanks @Lyrkan!

PR #544, opened 2019-03-21
Allow to call configureBabel with an external Babel configuration

### LINKED ISSUE #543: Configuring the babel loader is not possible when using babelrc file
`configureBabel` currently has 2 arguments:

- a mandatory callback altering the Babel config
- an options argument allowing to configure some settings of the loader rule (excluding node modules or no for instance)

As the callback is incompatible with the usage of an external Babel config, this method throws an error when using it that way. But that means there is no way to configure options in such case.

Lyrkan:
This PR fixes #543 by allowing to call `Encore.configureBabel()` even if an external Babel configuration exists.

Basically:

* `null` (or any falsy value in reality, to keep it simple) is always allowed as a first argument
* a warning will be displayed if the first argument is a callback and something like a `.babelrc` file is detected
* a warning will be displayed if the second argument contains a non-whitelisted option (ie. not `includeNodeModules`/`exclude`) and something like a `.babelrc` file is detected

```js
// Allowed without any warning
Encore.configureBabel(null, {
  includeNodeModules: ['foo']
});

// Displays a warning and the
// callback is ignored
Encore.configureBabel(
  () => { /* ... */ }
);

// Displays a warning and the
// useBuiltIns option is ignored.
Encore.configureBabel(null, {
  useBuiltIns: 'foo',
  includeNodeModules: ['foo'],
});
```


Lyrkan:
Test failure is unrelated, it seems that Babel is now displaying a warning when `useBuiltIns` is set (which is the case by default in Encore) without also setting the `corejs` option.

See: https://github.com/babel/babel/pull/7646

index.js
```diff
@@ -820,12 +829,15 @@ class Encore {
...
+     *              file for instance). In this case you can set the option directly into
+     *              that configuration file.
      *
-     * @param {function} callback
+     * @param {function|false} callback
      * @param {object} encoreOptions
      * @returns {Encore}
      */
-    configureBabel(callback, encoreOptions = {}) {
+    configureBabel(callback = false, encoreOptions = {}) {
```
weaverryan: Does it need a default value?

lib/WebpackConfig.js
```diff
@@ -328,16 +328,23 @@ class WebpackConfig {
         this.useSourceMaps = enabled;
     }
 
-    configureBabel(callback, options = {}) {
-        if (typeof callback !== 'function') {
-            throw new Error('Argument 1 to configureBabel() must be a callback function.');
-        }
+    configureBabel(callback = false, options = {}) {
```
weaverryan: Less important, but same here - does `callback` need a default?

index.js
```diff
@@ -806,12 +811,16 @@ class Encore {
      *              A Webpack Condition passed to the JS/JSX rule that
      *              determines which files and folders should not be
      *              processed by Babel (https://webpack.js.org/configuration/module/#condition).
+     *              Can be used even if you have an external Babel configuration
+     *              (a .babelrc file for instance)
      *              Cannot be used if the "include_node_modules" option is
```
weaverryan: Not part of this PR, but should be `includeNodeModules`

index.js
Lyrkan: Probably not :)

index.js
```diff
@@ -820,8 +829,11 @@ class Encore {
      *              by individual polyfills. Using it with 'usage' will try to
      *              automatically detect which polyfills are needed for each file and
      *              add them accordingly.
+     *              Cannot be used if you have an external Babel configuration (a .babelrc
+     *              file for instance). In this case you can set the option directly into
+     *              that configuration file.
      *
-     * @param {function} callback
+     * @param {function|false} callback
```
stof: I suggest documenting as `function|null` rather than `function|false`. Using a boolean to represent the absence of value does not make much sense.

stof: btw, the error message even says `Argument 1 to configureBabel() must be a callback function or null.`

Lyrkan: Yeah, I initially went with `null` but for some reasons switched to `false` (and didn't update that message).

I think my reasoning was something like "that parameter is disabled and won't be used", but now that you point it out I agree that it makes more sense to put `null`.

weaverryan:
Thanks @Lyrkan!

PR #550, opened 2019-03-31
Clean up the public API file and improve a few other things

Lyrkan:
Now that we don't have many active PRs anymore I figured it was a good time to do some spring cleaning, especially in the `index.js` file (since we often recommend users to look at its content).

This PR:

* Moves the proxy creation into its own file (`lib/EncoreProxy.js`)
* Moves the content of `initializeWebpackConfig(...)` inside of the `WebpackConfig`'s constructor (we already had some checks related to the `RuntimeConfig` there anyway)
* Fixes a few JSDoc issues (missing or misplaced comments, types not resolving correctly, etc.)
* Puts examples from the `index.js` file's comments inside of Markdown code-blocks, which are supported by at least PHPStorm and VSCode:

  * In PHPStorm

  | Before | After |
  |--------|-------|
  | image | image |

  * In VSCode (note that the leading whitespaces and asterisk are caused by this bug: https://github.com/Microsoft/TypeScript/issues/23517)

  | Before | After |
  |--------|-------|
  | image | image |

* Indicates that `Encore.getWebpackConfig()` returns a Webpack configuration, which allows IDEs' autocompletion to work on it:

  image

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -12,7 +12,7 @@
 
 const parseRuntime = require('../lib/config/parse-runtime');
 const context = require('../lib/context');
-const chalk = require('chalk');
+const chalk = require('chalk').default;
 const logger = require('../lib/logger');
 
 const runtimeConfig = parseRuntime(
@@ -40,7 +40,7 @@ if (!runtimeConfig.isValidCommand) {
     }
     showUsageInstructions();
 
-    process.exit(1);
+    process.exit(1); // eslint-disable-line no-process-exit
 }
 
 if (runtimeConfig.helpRequested) {
--- a/index.js
+++ b/index.js
@@ -40,16 +40,24 @@ class Encore {
      * The public version of outputPath: the public path to outputPath.
      *
      * For example, if "web" is your document root, then:
-     *      .setOutputPath('web/build')
-     *      .setPublicPath('/build')
+     *
+     * ```
+     * Encore
+     *     .setOutputPath('web/build')
+     *     .setPublicPath('/build')
+     * ```
      *
      * This can also be set to an absolute URL if you're using
      * a CDN: publicPath is used as the prefix to all asset paths
      * in the manifest.json file and internally in webpack:
-     *      .setOutputPath('web/build')
-     *      .setPublicPath('https://coolcdn.com')
-     *      // needed when public path is absolute
-     *      .setManifestKeyPrefix('/build')
+     *
+     * ```
+     * Encore
+     *     .setOutputPath('web/build')
+     *     .setPublicPath('https://coolcdn.com')
+     *     // needed when public path is absolute
+     *     .setManifestKeyPrefix('/build')
+     * ```
      *
      * @param {string} publicPath
      * @returns {Encore}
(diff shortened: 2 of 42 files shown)


bin/encore.js
```diff
@@ -12,7 +12,7 @@
 
 const parseRuntime = require('../lib/config/parse-runtime');
 const context = require('../lib/context');
-const chalk = require('chalk');
+const chalk = require('chalk').default;
```
weaverryan: `.default`?

index.js
```diff
@@ -9,53 +9,16 @@
...
-    webpackConfig = new WebpackConfig(runtimeConfig);
-}
-
-// If runtimeConfig is already set webpackConfig can directly
-// be initialized here.
-if (runtimeConfig) {
-    initializeWebpackConfig();
-}
+let runtimeConfig = context.runtimeConfig;
+let webpackConfig = runtimeConfig ? new WebpackConfig(runtimeConfig, true) : null;
```
weaverryan: If there is no runtimeConfig, when is this variable initialized?

lib/config-generator.js
```diff
@@ -9,6 +9,7 @@
 
 'use strict';
 
+const WebpackConfig = require('./WebpackConfig'); //eslint-disable-line no-unused-vars
```
weaverryan: I didn't realize we needed to import the class in order for the jsdoc to function below. If that's the case, I'm surprised eslint isn't smart enough to know that this really is "used"

bin/encore.js
Lyrkan: Yeah... it doesn't change anything at runtime since `require('chalk')` is the same thing than `require('chalk').default` but it seems that due to how chalk is typed, VSCode (which uses TypeScript to resolve things) doesn't find its properties if you don't specify the `.default`:

image

image

index.js
Lyrkan: In `Encore.configureRuntimeEnvironment(...)`

That was already the case before, I basically just replaced calls to `initializeWebpackConfig(...)` by `new WebpackConfig(...)` (since the whole "validation" part is now done by the constructor).

weaverryan:
Thank you @Lyrkan!

lib/config-generator.js
stof: well, everything is scoped in JS. As our class is not a global variable, you cannot use it globally.

stof:
@Lyrkan https://github.com/Microsoft/TypeScript/issues/14377 seems to suggest a way to reference a type from another module in jsdoc without importing the file at runtime. It might be worth trying it.

Lyrkan:
> [Microsoft/TypeScript#14377](https://github.com/Microsoft/TypeScript/issues/14377) seems to suggest a way to reference a type from another module in jsdoc without importing the file at runtime. It might be worth trying it.

@stof Yep, in a first version I actually used:

```js
/** @typedef {import('./WebpackConfig')} WebpackConfig */
```

It works fine in VSCode but since that's not valid JSDoc I went for `require(...)` instead.

PR #555, opened 2019-04-07
Set "sourceType" to "unambiguous" by default in Babel's options

### LINKED ISSUE #552: require appears to not work with useBuiltIns: usage
See https://github.com/webpack/webpack/issues/4039#issuecomment-477779322

Found this when starting a tutorial, where I first show `require` and `module.exports` being used. Error is:

> get_nice_message.js:1 Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'

This normally happens if you mix require & import/export. But, that's not happening in this case. I haven't checked into the issue much yet...

Lyrkan:
This PR fixes #552 by setting the `sourceType` option of Babel to `unambiguous` by default.

For more details about the reasons behind this change, see https://github.com/symfony/webpack-encore/issues/552#issuecomment-480539276.


Lyrkan:
One of the jobs currently fails due to a bug in the new version of `mocha` released a few hours ago (see https://github.com/mochajs/mocha/issues/3866).

weaverryan:
Thank you @Lyrkan!

Issue #567, opened 2019-04-16
Document how to works with environment variables


Kocal:
Just a reminder to write a documentation for `Encore.configureDefinePlugin()`, and the integration with Symfony `.env` files and `process.env.*`.

+ https://github.com/motdotla/dotenv

```js
Encore.configureDefinePlugin(options => {
  options['process.env'].MY_OPTION = JSON.stringify('foo');
});
```

For example for configuring sentry DNS from `.env` file:
```js
Encore.configureDefinePlugin(options => {
  const env = dotenv.config();
    
  if (env.error) {
    throw env.error;
  }

  options['process.env'].SENTRY_DSN = JSON.stringify(env.parsed.SENTRY_DSN);
});
```

Lyrkan:
We'll have to be careful about that one: there could be confusion related to the fact that it will be interpretated at build time and not on the server it'll be deployed to (like Symfony).

Kocal:
Yup, we should note that

Kocal:
In fact, what about adding a new method like `.injectEnvironmentVariables(['NODE_ENV', '...'])` that will:
- resolve environment variables like Symfony do
- automatically inject them with DefinePlugin + `JSON.stringify`

?

It can be a bit hard to implement, but it will be 100 times more user friendly

anaelChardan:
What about this issue ? Would be very helpful ;)

Kocal:
A proposal has been opened on #568, but it's totally not an easy feature in fact :p

yaotzin1:
I start using https://www.npmjs.com/package/dotenv-webpack in my project. 

After installation:

In webpack.config.js import it:

`const Dotenv = require('dotenv-webpack');`

Add it as plugin:

`    .addPlugin(new Dotenv(
        {
            path: './.env.local',
            systemvars: true
        }
    ))`

start using it in code:

`process.env.npm_package_version`

veewee:
When used wrong, dotenv-webpack can leak e.g. the database connection string into public JS code.
I made a little plugin based on dotenv and define that only includes `PUBLIC_` prefixed env vars from both .env and .env.local

```js
.addPlugin((() => {
        const parse = (envs) => envs.reduce((env, current) => {
            if (current.errors) {
                console.error(current.errors);
            }

            return {
                ...env,
                ...Object.fromEntries(
                    Object.entries(current.parsed ?? {})
                        .filter(([key]) => key.startsWith('PUBLIC_'))
[...]

PR #568, opened 2019-04-22
WIP: Automatically inject environment variables from `.env` files

Kocal:
Following #567.

This PR aims to inject environment variables (when calling `Encore.injectEnvironmentVariables()`) from `process.env` and `.env` files, like Symfony DotEnv component does.

Environment variables will be loaded from:
  - `process.env`
  - `.env` if exists, otherwise `.env.dist if exists`
  - `.env.local` if exists
  - in production environment:
    - `.env.prod` if exists
    - `.env.prod.local` if exists
  - in development environment:
    - `.env.dev` if exists
    - `.env.dev.local` if exists

`.env` files parsing is actually done by [`dotenv` package](https://github.com/motdotla/dotenv#readme), but we will have to write our own implementation in order to support the following syntaxes:
```.env
DB_USER=user # a comment 
DB_PASSWORD=${DB_USER}_passwd # include the user as a prefix
DATE=$(date) # exec shell commands
```

Note that it will not pollute `process.env` global variable. Environment variables will be stored in a local object that will be used with `DefinePlugin`.


Lyrkan:
Hi @Kocal,

I'm a bit skeptical about this PR.
While I really think that we should document and/or help users with env vars, what I'm less convinced about is that Encore should be reading `.env` files.

What I mean by this is: I don't think the `.env` files used by Symfony translate well to JS. As I said in #567 my main concern is the eventual confusion coming from the fact that in one case the file will be read at build time, and for the other one during runtime.

The environment names could also be an issue. With Encore you can only use `dev` or `prod` (well, there is also `dev-server` but that doesn't count), whereas with Symfony you could have *any* name. What if I want it to load `.env.test` for instance?

The last point that bothers me a bit is that we would have to implement our own reader.
Even though that's something perfectly do-able in theory it may end-up being really annoying for the simple reason that we would have to ensure it will work the same way than the version of the DotEnv component that was installed by Composer (which could change over time).

Don't get me wrong, I'm not saying ":-1:", just wondering if that's really something we should do :)

Kocal:
:hand: 

> I'm a bit skeptical about this PR.
While I really think that we should document and/or help users with env vars, what I'm less convinced about is that Encore should be reading .env files.

To be honest I'm a bit skeptical too, maybe this PR does _too much_ things. :thinking: 

> What I mean by this is: I don't think the .env files used by Symfony translate well to JS. As I said in #567 my main concern is the eventual confusion coming from the fact that in one case the file will be read at build time, and for the other one during runtime.

True, we should document the fact that environment variables are _injected_ during build time, and document that can be used for tree-shaking (e.g.: `if (process.env.NODE_ENV === 'development') { ... }`, this code will be removed when building for production).

> The environment names could also be an issue. With Encore you can only use dev or prod (well, there is also dev-server but that doesn't count), whereas with Symfony you could have any name. What if I want it to load .env.test for instance?

Yes we are doomed here. It's something I never liked with Encore, you can only have two environments `production` and `dev`. :disappointed: 

I've just checked the source code and we can use `NODE_ENV` env var to define a custom env:
Sélection_586

but it seems it's overloaded by the following code:
Sélection_585

Maybe we should not overload `runtimeConfig.environment` if it's already defined?

Another good approach is what `vue-cli` does with [Modes and Environment Variables](https://cli.vuejs.org/guide/mode-and-env.html). If you are not familiar with `vue-cli`, you have three modes (**not** envs) by default:
  - `development` when running `vue-cli-service serve`
  - `test` when running `vue-cli-service test:unit`
  - `production` when running `vue-cli-service build`

You can overwrite the default mode for a command with `--mode <mode>` arg, for example `vue-cli-service build --mode development` will build the project in `development` mode.

For environment variables, `vue-cli` load them like this:
  - `.env`
  - `.env.local`
  - `.env.[mode]`
  - `.env.[mode].local`

(Note that you can also define `NODE_ENV` env var from `.env` files.)

So if with Encore we can do something like `encore build --mode test`, that would be ideal. :+1: 
(And maybe make `encore dev` and `encore production` shortcuts to `encore build --mode [mode]`?)

> ... for the simple reason that we would have to ensure it will work the same way than the version of the DotEnv component that was installed by Composer (which could change over time).

Ah yes, I didn't think about this.
Maybe Encore can execute a PHP script that will returns a JSON encoded object? :smile: :eyes: 

I'm laughing about this solution, but if we can use the Symfony DotEnv directly that would be awesome. 
But I'm also thinking, what happens if we are not inside a Symfony project? Fallback on `dotenv` package?

weaverryan:
1) About the "mode" stuff, that *is* interesting. Webpack now has a `mode` option (though I think it also only supports 2 values) - and our system actually pre-dates that. Having something more flexible is interesting - but we also need to "stay in the spirit" of what the Webpack mainstream is doing. But generally speaking, this makes sense to me - we could have an `Encore.getMode()`, which people could use to make different decisions. And we could also expose that in the `DefinePlugin` if we want to allow it to be used in code.

2) About the `.env` stuff, it's also interesting, but probably not something that will be used by *many* users. Could we document it instead? Show using the `dotenv` package to populate the DefinePlugin? And of course, tell people that there are certain syntaxes that will work in `DotEnv` that won't in `dotenv`.

Kocal:
1. Be careful, I'm not speaking about Webpack modes but about `vue-cli` modes, they are totally differents.
In `vue-cli`, modes are only used to load `.env.[mode]` and `.env.[mode].local` files.
Webpack mode is defined to `production` if `process.env.NODE_ENV` equals to `production` (see [production config](https://github.com/vuejs/vue-cli/blob/dev/packages/@vue/cli-service/lib/config/prod.js#L12)).

For example you can create the following file:
```env
# .env.staging
VUE_APP_API_URL=https://staging.example.com/api

NODE_ENV=production # will set Webpack production mode
```
and run `yarn vue-cli-service build --mode staging`. Your app will be built for a **staging** env but it will still be optimized due to Webpack **production** mode.

`vue-cli` modes gives us a lot of flexibility that actually we don't have with Encore (because we don't handle `.env` files, so we didn't need mods).

2. Yep we can document it, but if we implement Encore modes like `vue-cli` does, it's essential to handle `.env` files, otherwise it's useless to implement a modes system.

Lyrkan:
What about adding a CLI option that'd allow to load `.env` files if they exist?

For instance:

```bash
# Loading order: env vars, .env, .env.test
# If `.env` doesn't exist only `.env.test` is loaded
$ yarn encore prod --env-file .env --env-file .env.test
```

That may be a bit verbose but with npm scripts it can be done relatively quickly.

In addition we could have some doc that explain how to use all of that.
Maybe in a page that shows all the ways data can be passed to JS code (that's a frequently asked question) ?
Some ideas:

* how to do it at build time (by using env vars)
* how to do it at runtime (by exposing things globally, by using data attributes, by getting parameters from an API, etc...)

Kocal:
> What about adding a CLI option that'd allow to load .env files if they exist?

Ah why not, I think it's the  better alternative for not adding:
 - `vue-cli`-like modes
 - `.env` files support
 - `NODE_ENV` handling for Webpack mode or everywhere else (`.isProduction()`)

to Encore at the same time :+1: 

> ```
> # Loading order: env vars, .env, .env.test
> ```

What happens if `--env-file` is not specified? `process.env` is still wholly injected or not?

> In addition we could have some doc that explain how to use all of that.
> Maybe in a page that shows all the ways data can be passed to JS code (that's a frequently asked question) ?

That a nice idea, but I think Env Vars should have their own page (to make it easier to find on https://symfony.com/doc/current/frontend.html), no?

Lyrkan:
> What happens if --env-file is not specified? process.env is still wholly injected or not?

If you're talking about what would be available from the entrypoints: nothing (even if `--env-file` are specified).

The env vars would however be available in the `webpack.config.js`, allowing them to be used through the `DefinePlugin`.

That option would simply be a shortcut for:

* requiring `dotenv` (or displaying a nice error message if the dependency isn't installed)
* loading the specified files

So, if you don't specify `--env-file` you'll get the same result as today.

> That a nice idea, but I think Env Vars should have their own page (to make it easier to find on symfony.com/doc/current/frontend.html), no?

I'd say that it depends on the length of the page, IIRC having really short ones should be avoid.

tristanbes:
The current implementation of env variable inside Webpack Encore is broken.

Let's say you have 

```
// .env
API_KEY=XXXX
```

```
// .env.dev
API_KEY=YYYY
```


```.js
// webpack.config.js

Encore.configureDefinePlugin((options) => {
    const dotenvEnv = dotenv.config();
    if (dotenvEnv.error) {
      throw dotenvEnv.error;
    }

    const env = { ...dotenvEnv.parsed, ...process.env };

    ['API_KEY'].forEach((envVar) => {
      options['process.env'][envVar] = JSON.stringify(env[envVar]);
    });
  });
```

When you're running `yarn dev-server` or `yarn dev`, it will just output 
`console.log(process.env.API_KEY) // XXXX`

instead of `YYYY`

It just ignore all logic (= ignores `.env.prod` `.env.dev` `.env.local`...)

More than one developer will fall into this trap;

Kocal:
> The current implementation of env variable inside Webpack Encore is broken.

It can be broken since there is no way to define them.

Lyrkan:
> It just ignore all logic (= ignores .env.prod .env.dev .env.local...)
>
> More than one developer will fall into this trap;

Well, that's actually an argument for not implementing this in Encore.

In your snippet you are using `dotenv` that only reads `.env` by default. They even recommend [**not** using multiple .env files in their documentation](https://www.npmjs.com/package/dotenv#should-i-have-multiple-env-files) which is kind of the opposite of Symfony.

We could probably copy what the DotEnv component of Symfony does, but in my opinion that wouldn't be a good idea either. As I explained in an earlier comment Symfony environments don't really apply to building assets since one is evaluated at runtime and the other one at build time. It could confuse some developers as well.

That being said you can implement any logic you want in your code by using `Encore.isProduction()`/`Encore.isDev()` and the [`path` option of `dotenv`](https://www.npmjs.com/package/dotenv#path).

yaotzin1:
maybe instead of inviting the wheel again, it is better to integrate with react-create-app?

Kocal:
To integrate what?

yaotzin1:
> To integrate what?
https://create-react-app.dev/docs/getting-started/

Tools provided by Vue or react for example.

yaotzin1:
Nevermind, I fixed the problem with loading .env files using dotenv-webpack plugin...

Issue #571, opened 2019-04-30
Node.js version support


kpower:
Short. What's lowest node.js version expected to be supported (in webpack encore's stable release)?

Long.
I've tried to install webpack on Shared Hosting with CloudLinux (a) and on VPS hosting with CentOS 7 (b). Both have problems with `yarn install` when using Node.js > 6.x ("Unhandled rejection TypeError: child.send is not a function" and "An unexpected error occurred: "EMFILE: too many open files, copyfile...").
Finally I've solved my problem: the only node version that can correctly make `yarn install` is 6.*. But to use it, I need @symfony/webpack-encore ^0.24.0 - because 0.27 needs Node.js 8.0+. So I expect in future releases I will lost this option (after 1.0 would be released, it's not a good solution to use 0.* versions I think). Am I right?

Kocal:
Given the [Travis configuration](https://github.com/symfony/webpack-encore/blob/master/.travis.yml#L43), the lowest compatible version of Node is 8.

Lyrkan:
> Short. What's lowest node.js version expected to be supported (in webpack encore's stable release)?

I'd consider the current version a stable release, but if you're talking about `1.x.x` (which should be released for Webpack 5) it could change based on the minimum version required by the dependencies used in Encore.

That being said I think we'll always try to support at least the active LTS versions.
As you can see from [this page](https://nodejs.org/en/about/releases/), v6 won't be maintained anymore (as of tomorrow) which is why we switched to v8 in #540.

Another solution to your issue could be to do your build before deploying on your server (which won't even need Node.js at all).

kpower:
Thanks for your comments (building locally was my final solution) and answer about 6.x future.

stof:
if you get an error while running `yarn install`, that's an issue with the node version supported by yarn (or maybe with the version of yarn installed on your server), not with the version of node supported by webpack-encore (as you are not running Encore yet when running that command).

Kocal:
In this case yes, but yarn follows [`engines.node`](https://docs.npmjs.com/files/package.json#engines) configuration from `package.json` and will fail if Node version does not match.

PR #572, opened 2019-04-30
doc: add badge for lowest supported Node version

Kocal:
Following of #571, it can be nice to have a badge that shows the lowest supported version of Node.

Under the hood, it uses `engines.node` from [`package.json`](https://github.com/symfony/webpack-encore/blob/master/package.json#L24)


Lyrkan:
Yep, good idea

Thank you @Kocal :)

PR #583, opened 2019-05-31
minor #583 chore: replace some `.indexOf() === -1` with `.includes()` (Kocal)

Kocal:
Some small internal changes.

It replaces some calls to `.indexOf()` with `.includes()` which is more readable:
- `a.indexOf(b) === -1` -> `!a.includes(b)`
- `a.indexOf(b) !== -1` -> `a.includes(b)`


Lyrkan:
Thanks @Kocal

PR #594, opened 2019-06-19
Add support for .postcss files

### LINKED ISSUE #592: Add a .postcss test in enablePostCssLoader
I'm not sure how to format this issue, but everything is explained in this ticket: https://github.com/vuejs/vetur/issues/1330

Adding the following loader solves my issue:

```js
.addLoader({
    test: /\.postcss$/,
    use: [
        'vue-style-loader',
        {
            loader: 'css-loader',
            options: { importLoaders: 1 }
        },
        'postcss-loader'
        ]
})
```

I don't think I should *have to* add this manually. This loader is related to both `Vue` and `PostCSS`, so maybe a new method would solve the problem.

Lyrkan:
This PR closes #592 by making the `.css` rule also process `.postcss` files when `Encore.enablePostCssLoader()` has been called.

It seems that it can be useful in a Vue.js context in order to have proper syntax highlighting in IDEs by using `lang="postcss"`, as explained on https://vue-loader-v14.vuejs.org/en/features/postcss.html:

> Since vue-loader handles PostCSS on its styles internally, you only need to apply postcss-loader to standalone CSS files. There's no need to specify lang="postcss" on a style block if there is a PostCSS config file in your project.
>
> Sometimes the user may want to use lang="postcss" only for syntax highlighting purposes.

### DIFF
--- a/fixtures/css/postcss_extension.postcss
+++ b/fixtures/css/postcss_extension.postcss
@@ -1,0 +1,1 @@
+@import "autoprefixer_test.css";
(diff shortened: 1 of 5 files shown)


weaverryan:
Thanks @Lyrkan! Looks very sensible - great tests, as usual, which is HUGE :)

Issue #596, opened 2019-06-24
upgrade packages


dvdknaap:
Hi,

I've update my url-loader package to 2.0.0 but when i run yarn build i get the warning message that webpack encore requires version ^1.0.1.

Could you upgrade the packages to the newest versions to fix these errors?

Thanks,

Lyrkan:
Will be fixed by #599.

Note that what you currently get is only a warning, everything should work fine if you use `url-loader@^2`.

PR #598, opened 2019-06-28
Webpack dev server updated from 3.1.14 to 3.7.2

gustawdaniel:
For the sake of #596 packeges webpack-dev-server and url-loader are updated to newest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "stylus-loader": "^3.0.2",
     "ts-loader": "^5.3.0",
     "typescript": "^2.3.4",
-    "url-loader": "^1.0.1",
+    "url-loader": "^2.0.1",
     "vue": "^2.3.4",
     "vue-loader": "^15.0.11",
     "vue-template-compiler": "^2.3.4",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "webpack": "^4.20.0",
     "webpack-chunk-hash": "^0.6.0",
     "webpack-cli": "^3.0.0",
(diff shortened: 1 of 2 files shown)


Lyrkan:
Hi @gustawdaniel,

I don't think that updating `webpack-dev-server` is required, our current range (`^3.1.14`) already allows people to use `3.7.2`.

As for `url-loader@^2` there are some things to consider as I noted on #596:
* it should already be possible to use `url-loader@^2` with Encore since that's a dev-dependency (maybe with a warning but it won't prevent it from working fine)
* the version is quite recent so it may be a good idea to wait for a bit before updating (for instance `2.0.0` introduced the following regression which was fixed 3 days ago: https://github.com/webpack-contrib/url-loader/issues/183)
* this is a major bump which introduces breaking changes: one of them being the min Node.js version set to `>=8.9.0`. It means that people using `>= 8.0.0 <8.9.0` won't be able to use the `url-loader` without a warning since **our** version requirement is `8.* || >= 10.*`). One solution to that would be to allow both `url-loader@^1` and `url-loader@^2` (Encore doesn't set any option so it should not break anything).

gustawdaniel:
Ok. So I closing this pull request. Thanks for explaination @Lyrkan

Lyrkan:
As I said I'm not entirely against this, having something like `"url-loader": "1.* || 2.*"` would be a valid way of doing it without requiring people with to update Node.js or the `url-loader`.

gustawdaniel:
I removed form so there is unknown repository so i created it again in:

https://github.com/symfony/webpack-encore/pull/599

PR #602, opened 2019-07-07
Update license year

Arman-Hosseini:


Lyrkan:
Thank you @Arman-Hosseini

PR #603, opened 2019-07-08
Allow to configure the resolve-url-loader

diegocardoso93:
This PR solve the error below, occuring on some cases:
```
Error: resolve-url-loader: CSS error
  source-map information is not available at url() declaration (found orphan CR, try removeCR option)
```
See discussion: bholloway/resolve-url-loader#107

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -116,13 +116,7 @@ class WebpackConfig {
         this.copyFilesConfigs = [];
         this.sassOptions = {
             resolveUrlLoader: true,
-            resolveUrlLoaderOptions: {
-                engine: 'postcss',
-                keepQuery: false,
-                removeCR: false,
-                debug: false,
-                silent: false
-            }
+            resolveUrlLoaderOptions: {}
         };
         this.preactOptions = {
             preactCompat: false
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -115,7 +115,14 @@ class WebpackConfig {
         // Features/Loaders options
         this.copyFilesConfigs = [];
         this.sassOptions = {
-            resolveUrlLoader: true
+            resolveUrlLoader: true,
+            resolveUrlLoaderOptions: {
+                engine: 'postcss',
+                keepQuery: false,
+                removeCR: false,
+                debug: false,
+                silent: false
+            }
         };
         this.preactOptions = {
             preactCompat: false
--- a/lib/loaders/sass.js
+++ b/lib/loaders/sass.js
@@ -32,7 +32,7 @@ module.exports = {
                 loader: 'resolve-url-loader',
                 options: {
                     sourceMap: webpackConfig.useSourceMaps,
-                    removeCR: true
+                    ...webpackConfig.sassOptions.resolveUrlLoaderOptions
                 }
             });
         }
--- a/lib/loaders/sass.js
+++ b/lib/loaders/sass.js
@@ -30,10 +30,12 @@ module.exports = {
             // entry file, not the file that contains the url()
             sassLoaders.push({
                 loader: 'resolve-url-loader',
-                options: {
-                    sourceMap: webpackConfig.useSourceMaps,
-                    ...webpackConfig.sassOptions.resolveUrlLoaderOptions
-                }
+                options: Object.assign(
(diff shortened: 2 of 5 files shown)


Lyrkan:
Hey @diegocardoso93

`resolve-url-loader@3.1.0` was already allowed by the current version range (`^3.0.1`) so the question is: is `removeCR: true` a good default value?

Do you know if it could have any drawbacks on setups that don't need it to be `true` (ie. why it isn't already `true` by default in the `resolve-url-loader`)?

If that's the case it could  be better to allow setting it easily (maybe using a new option in `enableSassLoader`) and detect the error (using `friendly-errors-webpack-plugin`) to display a comprehensive message. What do you think?

diegocardoso93:
Hello @Lyrkan 
Totally agree, it can be better if the decision of enable `removeCR` option can be take by user.
Could you do this change?
Thanks.

diegocardoso93:
@Lyrkan I've just updated with the `resolveUrlLoaderOptions`.
I don't sure if the option variable name is the best.
Can you take a look?

lib/WebpackConfig.js
```diff
@@ -115,7 +115,14 @@ class WebpackConfig {
         // Features/Loaders options
         this.copyFilesConfigs = [];
         this.sassOptions = {
-            resolveUrlLoader: true
+            resolveUrlLoader: true,
+            resolveUrlLoaderOptions: {
+                engine: 'postcss',
```
Lyrkan: Is there a reason why you explicitely set the value of those options there?
Couldn't we go with `resolveUrlLoaderOptions: {}` instead?

lib/loaders/sass.js
```diff
@@ -31,7 +31,8 @@ module.exports = {
             sassLoaders.push({
                 loader: 'resolve-url-loader',
                 options: {
-                    sourceMap: webpackConfig.useSourceMaps
+                    sourceMap: webpackConfig.useSourceMaps,
+                    ...webpackConfig.sassOptions.resolveUrlLoaderOptions
```
Lyrkan: I don't think that the spread operator is allowed in this context with the version of ESLint we are currently using (which is why the "lint" job fails on Travis).

So two choices there:
* update ESLint (but maybe that's a chore for another PR)
* use `Object.assign()`

stof: the issue is not the eslint version, but the node.js version.

Lyrkan: The linter is executed using Node 10 which should supports this.
It should even be ok under Node 8 since that job is green: https://travis-ci.org/symfony/webpack-encore/jobs/560124277

```
$ nvm use v8.15.1
Now using node v8.15.1 (npm v6.4.1)

$ node
> const foo = { foo: 1, bar: 2 };
undefined
> const bar = { ...foo, bar: 3 };
undefined
> bar
{ foo: 1, bar: 3 }
```

Lyrkan: Maybe that could be solved without updating eslint though, by changing the following option of our `.eslintrc.js` to `2018`:

https://github.com/symfony/webpack-encore/blob/9a3cb43e880f10c54e8319b76433058137da78d2/.eslintrc.js#L9

Lyrkan: Actually you were partially right @stof, using `ecmaVersion: 2018` is triggering other errors:

```
  99:22  error  Spread properties are not supported yet on Node 8.0.0                                                           node/no-unsupported-features
  99:22  error  Rest/spread properties are not supported until Node.js 8.3.0. The configured version range is '8.* || >= 10.*'  node/no-unsupported-features/es-syntax
```

We should probably change that range since most Webpack loaders are now requiring `>= 8.9.0` (for instance: [`css-loader`](https://github.com/webpack-contrib/css-loader/pull/948), [`url-loader`](https://github.com/webpack-contrib/url-loader/pull/182), [`file-loader`](https://github.com/webpack-contrib/file-loader/blob/d441daa955246967686d24a49da83329fad6d1b0/package.json#L12), ...)

But as I said earlier, maybe it should be handled in another PR.

diegocardoso93: I just updated the PR with changes requested.

weaverryan:
Very sensible & clean addition. Thank you Diego!

PR #627, opened 2019-08-13
Allow to overwrite pre-defined aliases using addAliases()

### LINKED ISSUE #625: We can not choose a proper build for Vue.js
I would like to use the Vue.js build like defined by my alias : 

```
{
      'vue$': Encore.isProduction() ? 'vue/dist/vue.runtime.min.esm': 'vue/dist/vue.runtime.esm',
}
```
but it is not possible because of this file which rewrites my alias in an unwanted way. 

https://github.com/symfony/webpack-encore/blob/5640271a87e63048b8ae261d210922b4972cab69/lib/config-generator.js#L103

Why this build precisely? Why is it not configurable? The solution I found to work around the problem:

```
let config = Encore.getWebpackConfig()
config.resolve.alias = {
      'vue$': 'vue/dist/vue.runtime.esm',
    }
module.exports = config
```

I'm using the 0.28.0 version of "@symfony/webpack-encore"

Lyrkan:
Currently pre-defined aliases (used by `vue` and `preact-compat`) are added after the ones from `Encore.addAliases(...)`, which prevents overwriting them using this method.

It makes more sense to change that order so aliases from `Encore.addAliases(...)` are always added last (fixes #625).

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -96,7 +96,7 @@ class ConfigGenerator {
 
         config.resolve = {
             extensions: ['.wasm', '.mjs', '.js', '.json', '.jsx', '.vue', '.ts', '.tsx'],
-            alias: Object.assign({}, this.webpackConfig.aliases)
+            alias: {}
         };
 
         if (this.webpackConfig.useVueLoader) {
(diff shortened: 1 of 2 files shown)


weaverryan:
Thank you @Lyrkan!

Issue #638, opened 2019-09-13
Autoprefixer comments are stripped out when using sass-loader


wimhendrikx:
With [version 7.3.0 of sass-loader](https://github.com/webpack-contrib/sass-loader/releases/tag/v7.3.0), [compression](https://github.com/webpack-contrib/sass-loader/pull/723) is also done during the production build process. Apparently, this happens before autoprefixing kicks in (postcss-loader), which causes the necessary grid comments (```/* autoprefixer grid: on */```) to be removed, thus preventing Grid autoprefixing for IE during production.

I have found 2 possible fixes:

1. Adding a '!' in the comment => ```/*! autoprefixer grid: on */```
2. Disable compression in the sass-loader ```enableSassLoader(function (options) { options.outputStyle = 'expanded'; })```

Personally, the 2nd fix seems to be the best solution, but maybe this can be provided by default when compression is already done by postcss-loader?

weaverryan:
Indeed - because CSS minification is handled with `mini-css-extract-plugin`, I can't think of any reason for us to want Sass to do this for us. So I also vote for option (2) - specifically, making the `outputStyle` `nested` by default (because I think this is the default setting in dev?). @wimhendrikx would you be able to create a PR for this? The relevant file should be https://github.com/symfony/webpack-encore/blob/master/lib/loaders/sass.js

Lyrkan:
> Indeed - because CSS minification is handled with mini-css-extract-plugin, I can't think of any reason for us to want Sass to do this for us

People can opt-out from using the `mini-css-extract-plugin`, so that's a (small) breaking change. Not a big issue though since they would still be able to change `options.outputStyle` when calling `Encore.enableSassLoader()`.

> specifically, making the outputStyle nested by default (because I think this is the default setting in dev?)

That point may cause a bigger issue though, because Dart Sass does seem to support `nested`: https://sass-lang.com/documentation/js-api#outputstyle

> * "expanded" (the default for Dart Sass) writes each selector and declaration on its own line.
> * "compressed" removes as many extra characters as possible, and writes the entire stylesheet on a single line.
> * "nested" (the default for Node Sass, not supported by Dart Sass) indents CSS rules to match the nesting of the Sass source.
> * compact (not supported by Dart Sass) puts each CSS rule on its own single line.

We'll have to be careful about setting options that are not directly handled by the loader since they are directly passed to the chosen implementation which may or may not support them.

wimhendrikx:
[...]

PR #639, opened 2019-09-16
bug #638 sass-loader should not do css minification

wimhendrikx:
Fixes a BC break from the sass-loader minifying the CSS which is already done with mini-css-extract-plugin.

I chose to use outputStyle 'expanded' so it's also compatible with Dart Sass.


weaverryan:
Thank you @wimhendrikx!

PR #640, opened 2019-09-17
fix(copy-files): use context realpath

QWp6t:
In scenarios where context is within a symlink, `path.relative()` will return an undesired path because the resource path follows symlinks. We resolve this by using `fs.realpathSync(options.context)`.

### DIFF
--- a/lib/webpack/copy-files-loader.js
+++ b/lib/webpack/copy-files-loader.js
@@ -13,6 +13,7 @@ const LoaderDependency = require('webpack/lib/dependencies/LoaderDependency');
 const fileLoader = require('file-loader');
 const loaderUtils = require('loader-utils');
 const path = require('path');
+const fs = require('fs');
 
 module.exports.raw = true; // Needed to avoid corrupted binary files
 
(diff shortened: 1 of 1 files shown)


stof:
don't you need to do this both for the resource path and the content though ?

QWp6t:
In my testing, which was not very extensive at all, the symlink had already been followed for the `resourcePath`. I didn't investigate where it gets it, though, or if there are any caveats.

At any rate, I can do the same for the resource path too if you want.

weaverryan:
> Could you maybe add some functional tests using symlinks?

Yes, this would be great! It can be added in this section - probably we could just create a symlink at the start of the test, make sure that file is copied, then clean up the symlink after. Can you try that @QWp6t?

Thanks!

https://github.com/symfony/webpack-encore/blob/master/test/functional.js#L1753

QWp6t:
Hey guys, just wanted to stop by to mention I haven't forgotten about this PR. Yes, I can add some tests, but I have a deadline coming up, so it'll have to wait a bit. I think you guys have access to push to my fork if you're wanting to fast-track this for whatever reason, otherwise I'll make the changes you requested just as soon as I'm finished with work, probably end of next week.

Have a good week! 🍻

PR #642, opened 2019-09-29
Allow to configure babel env preset

ihmels:
This PR allows you to configure `@babel/preset-env` by calling `Encore.configureBabelPresetEnv()` without the need to create a `.babelrc` and lose the ability to activate presets automatically.

```js
Encore
    .configureBabelPresetEnv((config) => {
        config.corejs = 3;
        config.useBuiltIns = 'usage';
        config.include = ['foo-plugin'];
    });
```
The options `corejs` and `useBuiltIns` of `Encore.configureBabel()` should be deprecated.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -140,6 +140,7 @@ class WebpackConfig {
         this.lessLoaderOptionsCallback = () => {};
         this.stylusLoaderOptionsCallback = () => {};
         this.babelConfigurationCallback = () => {};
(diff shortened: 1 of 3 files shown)


weaverryan:
This makes sense - it's the one preset we add, so indeed, why not allow it to be configured. Thanks Yannick for the nice PR!

stollr:
Would be nice if an explanation or link to a resource could be added, where it is explained how to use this `configureBabelPresetEnv` for a newbie in the webpack/encore/babel/react world ;-)

Issue #573, opened 2019-05-01
Include async chunks in the entrypoints file


stof:
To preload or eagerly load async chunks on some pages, we would need to know the list of files necessary for that chunk (and to assign a name explicitly for the chunk to know how to get the right one, but that part is already possible with webpack and happens in the source code when importing the async chunk). AFAICT, including them in the entrypoints.json file should be enough to solve this. But I'm not sure `assets-webpack-plugin` supports that.

stof:
@weaverryan do you think this is possible.

Lyrkan:
@stof What would be the difference with `/* webpackPreload: true */` and `/* webpackPrefetch: true */`?

stof:
These won't work in the context of Symfony + Encore, as we don't render the HTML with webpack and so it cannot output a `<link rel="preload">` tag.

But my use case is indeed to be able to add such `<link>` tag.
Thus, in my case, I would like to preload some imported modules only on some of the pages, not everytime the main bundle is loaded (as I know some of my dynamic imports will be used only for specific features, for which I can know which page are very likely to use them and so should preload the chunk). So I would not want Encore to force bundling this knowledge into the entrypoint data itself. Having files needed for the different dynamic chunks (similar to how files are listed for entrypoints) would allow to render these.

Lyrkan:
> These won't work in the context of Symfony + Encore, as we don't render the HTML with webpack and so it cannot output a <link rel="preload"> tag.

I'm 90% sure it would also work in a Symfony context. Having these inline directives when doing a dynamic import will make Webpack add `<script>` or `<link>` tags at runtime (I know for sure that it works for `<script>` tags since I use it on some projects).

So, as long as you have the right public path you should end-up with your JS/CSS files preloaded as soon as the main JS file is executed.

I'm not sure it would cover your whole use case though since prefetching chunks only on some pages would require you to have the dynamic import in files that are specific to those pages. I'll look into it, but I don't think exposing async chunks in the `entrypoints.json` will be an easy thing to do.

Lyrkan:
Update: it seems that it actually doesn't work if you `import(/* webpackPrefetch: true */ 'something.css')` because Webpack will add a `<link>` tag for the generated (and almost empty) `.js` chunk instead of the `.css` file.

We'd probably have the same issue if we tried to add async chunks info to the `entrypoints.json` file.

Lyrkan:
[...]

PR #645, opened 2019-10-05
Update Webpack to v5 (+ other dependencies)

Lyrkan:
Last week Webpack added a compat layer to its v5 alpha that allows the `mini-css-extract-plugin` to run on it (with deprecated messages). Since we include that plugin by default (and a lot of our tests relie on it) it was the main thing blocking us from preparing the migration.

---

I basically started from #564 which was updating some dependencies, enabling CSS HMR when needed and adding a `configureMiniCssExtractPlugin(...)` method, but with a few changes:
* no more inconsistent hashes checks for the `enableVersioning applies to js, css & manifest` test: it seems to be working fine by default
* no more embedding the `webpack-manifest-plugin` into our code: https://github.com/danethurber/webpack-manifest-plugin/issues/167 is still an issue but @mastilver has been working on the project lately (which is why the plugin works with Webpack 5) and a fix has already been merged on the `next` branch, so it's probably only a matter of time now :). **Edit: Fixed in `webpack-manifest-plugin@^3.0.0-rc`**
* removal of Node 8 compatibility

So now, about the state of that PR:

**0 failing test left**:
* ~6 tests that will probably be fixed by https://github.com/webpack/webpack/pull/10661 in `webpack@5.0.0-beta.15`: `Uncaught Error: Error when running the browser: Error: Error when running the browser: ReferenceError: mod is not defined`~
* ~All the 7 tests related to the `vue-loader` are failing with a `Cannot find module 'webpack/lib/RuleSet` error message (see: https://github.com/vuejs/vue-loader/issues/1599).~
* ~1 test related to the `webpack-manifest-plugin` issue previously mentioned~
* ~1 test related to `createSharedEntry()` which doesn't seem to work properly~
* ~1 test related to Babel that doesn't transform an arrow function as expected~

**A lot of deprecation notices** (but most, if not all, of them are triggered by vendors), for instance:
* `Compilation.hooks.normalModuleLoader was moved to NormalModule.getCompilationHooks(compilation).loader`
* `Module.id: Use new ChunkGraph API`
* `Module.updateHash: Use new ChunkGraph API`
* `Chunk.modulesIterable: Use new ChunkGraph API`
* `ChunkGroup.getModuleIndex2 was renamed to getModulePostOrderIndex`
* `Compilation.chunks was changed from Array to Set (using Array method 'reduce' is deprecated)`
* `chunk.files was changed from Array to Set (using Array method 'reduce' is deprecated)`

**Some modules do not declare they are compatible with Webpack 5 yet** *(warning messages during `yarn install`)*: this shouldn't be an issue **unless** those modules require a major version upgrade to be officialy compatible (in which case breaking changes could impact us).

**We're still using `webpack-cli@3`** which may not support Webpack 5. It currently seems to be OK but we should probably upgrade to `webpack-cli@4` (currently in beta). I took a quick glance at it and it probably won't be an easy thing to do, mainly because of how our "runtime context" works and how the new version of the CLI calls Webpack (through another process).


Lyrkan:
I fixed the failing Babel test which wasn't really related to Babel but to the new [`output.ecmaVersion`](https://github.com/webpack/webpack/commit/e507392094b373f993ad2124926912ca446966bb#diff-8588527c440c140a86b57694ee7a97bc) option of Webpack 5.

This option tells Webpack which ES version it should use for the code it generates (for instance JSONP templates) which is **not** transpiled by Babel.

By default it is set to `2015`, meaning that it will use things like arrow functions, hence the failing test.

Since some users may not want that behavior (IE11 does not support arrow functions) I changed the default value in Encore to 5.

---

Regarding the deprecation messages I confirm that they are all triggered by plugins (and maybe by the Webpack core itself for some of them):

* `Compilation.hooks.normalModuleLoader was moved to NormalModule.getCompilationHooks(compilation).loader` => `mini-css-extract-plugin` / `vue-loader`
* `Module.id: Use new ChunkGraph API` => `mini-css-extract-plugin`
* `Module.updateHash: Use new ChunkGraph API` => `mini-css-extract-plugin`
* `Chunk.modulesIterable: Use new ChunkGraph API` => `mini-css-extract-plugin`
* `ChunkGroup.getModuleIndex2 was renamed to getModulePostOrderIndex` => `mini-css-extract-plugin`
* `Compilation.chunks was changed from Array to Set (using Array method 'reduce' is deprecated)` => `mini-css-extract-plugin`
* `chunk.files was changed from Array to Set (using Array method 'reduce' is deprecated)` => `webpack-manifest-plugin`

--- 

Not really related but I also removed old deprecation messages since they have all been there for a while now.

Lyrkan:
It seems that the hack we are using for `Encore.createSharedEntry()` doesn't work anymore.

The tests are currently failing because of the following change:

> optimization.splitChunks test no longer matches chunk name
> MIGRATION: Use a test function `(module, { chunkGraph }) => chunkGraph.getModuleChunks(module).some(chunk => chunk.name === "name")`

Since we are directly checking if the chunk name matches `sharedCommonsEntryName` the cache group is never used and no real shared entry is created.

Sadly, after fixing that part we get the following error message:

> BREAKING CHANGE: webpack < 5 used to allow to use the entrypoint as splitChunk. This is no longer allowed. Remove this entrypoint and add modules to cache group's 'test' instead.
> If you need modules to be evaluated on startup, add them to the existing entrypoints (make them arrays). See migration guide of more info.

Which is exactly how `Encore.createSharedEntry()` works:

https://github.com/symfony/webpack-encore/blob/5640271a87e63048b8ae261d210922b4972cab69/lib/config-generator.js#L517-L527

Kocal:
Is there still a legitimate reason to use `Encore.createSharedEntry()` ?

Lyrkan:
@Kocal I'd say that it's useful for people that are not using the Encore Bundle since they can't predict the names of the shared chunks Webpack will generate for them when using `splitEntryChunks()`.

index.js
```diff
@@ -916,6 +916,34 @@ class Encore {
...
+     *         // pluginConfig.chunkFilename = '[id].css';
+     *     }
+     * );
+     * ```
+     *
+     * @param {function} loaderOptionsCallback
+     * @param {function} pluginOptionsCallback
+     * @returns {Encore}
+     */
+    configureMiniCssExtractPlugin(loaderOptionsCallback, pluginOptionsCallback = () => {}) {
```
stof: should this be bundled in the webpack 5 upgrade, or be submitted separately ?

test/WebpackConfig.js
```diff
@@ -546,7 +546,7 @@ describe('WebpackConfig object', () => {
 
         it('Calling with "includeNodeModules" option', () => {
             const config = createConfig();
-            config.configureBabel(() => {}, { include_node_modules: ['foo', 'bar'] });
+            config.configureBabel(() => {}, { includeNodeModules: ['foo', 'bar'] });
```
stof: should already be done in the existing version

index.js
Lyrkan: I originally added it into #564 in case people wanted to disable the new `hmr` option (which is also included in this one). So, maybe we could ship it separately but it'd be preferable if it was done before the `mini-css-extract-plugin` upgrade.

test/WebpackConfig.js
Lyrkan: I remember keeping the `include_node_modules` check on purpose since we only deprecated that version. But maybe we should have had an additional check for the uppercase version.

test/bin/encore.js
```diff
@@ -39,6 +39,7 @@ describe('bin/encore.js', function() {
             `
 const Encore = require('../../index.js');
 Encore
+    .enableSingleRuntimeChunk()
```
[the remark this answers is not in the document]
Lyrkan: We could but that's not a big issue since the current version allows not to call `enableSingleRuntimeChunk()`/`disableSingleRuntimeChunk()` (unless you also called `createSharedEntry()` it defaults to the `disableSingleRuntimeChunk()` behavior with deprecation messages).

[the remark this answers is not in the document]
stof: I mean, we should avoid using deprecated API in our testsuite, unless we are specifically testing the BC layer (in which case the test gets removed when dropping it, or changed to a test asserting an exception).
That's the policy we have in Symfony itself (we even make the testsuite fail if deprecations are triggered in non-legacy tests).

weaverryan:
About `createSharedEntry()`, I think we should try this (and see how it looks/feels):

A) Remove `createSharedEntry()` (if we like the final solution, we can deprecate it in a Webpack4-supporting Encore release before releasing this)

B) Replace it with some interface to make it easier/obvious to the user to add their own cache group - e.g. to get some end-config that looks similar to this: https://webpack.js.org/plugins/split-chunks-plugin/#split-chunks-example-3

It could be something like:

```
.addCacheGroup('vendor', {
    test: /[\\/]node_modules[\\/](react|react-dom)[\\/]/,
})
```

That would be the lowest-level way to handle this. We might choose to instead/also add something more user-friendly:

```
.addCacheGroup('vendor', {
    # could be an optional "key" we add... the user could use this *or* test (but not both)
    node_modules: ['react', 'react-dom'],

    # or another possibility, where we allow mixing "node_modules" modules and also specific paths
    modules: ['react', 'react-dom', './assets/js/main'],
})
```

Not sure how far we need to go. If we did the first and added good documentation with a "vendors" example, that will go a long way.

Lyrkan:
Updated the PR with `webpack@^5.0.0-beta` and `webpack-manifest-plugin@^3.0.0-rc`:

* https://github.com/danethurber/webpack-manifest-plugin/issues/167 is now fixed and everything seems to be working fine with Webpack 5 :)
* Webpack has bumped its minimum requirement for Node's version to `10.13.0`, I haven't changed our test matrix yet but I can't see a reason *not* to do it since Node 8 goes EOL at the end of the year anyway.
* I opened an issue on `vue-loader`'s repository about the `webpack/lib/RuleSet` thing that makes it incompatible with Webpack 5 (https://github.com/vuejs/vue-loader/issues/1599)
* I like the `addCacheGroup()` idea to replace `createSharedEntry()`, I might give it a shot soon

Lyrkan:
PR updated:
* I bumped the version of a few dependencies (including webpack and vue)
* The issue we previously had with Vue.js has been fixed
* I removed `createSharedEntry()` since we already deprecated it in `master`
* Node 8 does not work anymore with some loaders, so I removed it from our test matrix

There are some new failing tests, but they all have the same cause and should be fixed in the next release of Webpack (see https://github.com/webpack/webpack/pull/10661).

weaverryan:
Awesome @Lyrkan! Thanks for keeping this up to date - we should be able to get Encore released fast once Webpack gets here - whenever that is :).

lib/friendly-errors/transformers/missing-loader.js
```diff
@@ -32,7 +32,7 @@ function isErrorFromVueLoader(filename) {
     }
 
     // vue3
-    if (filename.includes('vue-loader/dist??')) {
+    if (/vue-loader\/dist(\/index\.js)?\?\?/.test(filename)) {
```
Lyrkan, line 35: Looks like the filename now includes `/index.js`.
I don't know if that's always the case so I also kept supporting the version without it.

test/functional.js
```diff
@@ -943,23 +749,28 @@ describe('Functional tests using webpack', function() {
                 config.addEntry('page2', './preact/main');
 
                 // Move Vue.js code into its own chunk
-                config.addCacheGroup('vuejs', { test: /[\\/]node_modules[\\/]vue[\\/]/ });
+                const isVue2 = getVueVersion(config) === 2;
+                config.addCacheGroup('vuejs', {
+                    test: isVue2 ?
+                        /[\\/]node_modules[\\/]vue[\\/]/ :
+                        /[\\/]node_modules[\\/]@vue[\\/]/
+                });
```
Lyrkan, line 757: Quick explanation about why this is needed.
This is actually not directly related to Webpack 5 but because of how `vue` is now splitted into multiple packages under the `@vue` scope.

The imported module when you `import vue` does not do much:

```js
import { warn } from '@vue/runtime-dom';
export * from '@vue/runtime-dom';

// This entry exports the runtime only, and is built as
const compile = () => {
    if ((process.env.NODE_ENV !== 'production')) {
        warn(`Runtime compilation is not supported in this build of Vue.` +
            ( ` Configure your bundler to alias "vue" to "vue/dist/vue.esm-bundler.js".`
                ) /* should not happen */);
    }
};

export { compile };
```

With Webpack 4 the cache group seemed to worked because it created a shared file with the code of the `compile` function and the declaration of the multiple exports from `@vue/runtime-dom` (without their actual code... so not really useful).

With Webpack 5 I'm guessing that the `compile` function gets tree shaked and nothing matches the `node_modules/vue` test of the cache group anymore and no shared file is created.

Kocal:
Webpack 5 has been officially released https://webpack.js.org/blog/2020-10-10-webpack-5-release/ 🎉

Lyrkan:
Updated with `webpack@5.0.0` and `webpack-cli@4.0.0`.

Everything seems to work but it's still not done yet:

* there are some deprecation notices left:
  * `last-call-webpack-plugin` (used by `optimize-css-assets-webpack-plugin`): `DeprecationWarning: optimizeChunkAssets is deprecated (use Compilation.hook.processAssets instead and use one of Compilation.PROCESS_ASSETS_STAGE_* as stage option)`
  * `webpack-manifest-plugin` (which doesn't seem to be maintained anymore): `DeprecationWarning: Compilation.hooks.normalModuleLoader was moved to NormalModule.getCompilationHooks(compilation).loader`
  * `delete-unused-entries-js-plugin.js` (on our side but I don't know how to fix it yet): `DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated`

* there are warnings when using Vue files with `<style>` tags in our tests, not sure why, for instance: `ModuleDependencyWarning: export 'default' (reexported as 'default') was not found in '-!../../../../node_modules/mini-css-extract-plugin/dist/loader.js??clonedRuleSet-93.use[0]!../../../../node_modules/css-loader/dist/cjs.js??clonedRuleSet-93.use[1]!../../../../node_modules/vue-loader/dist/stylePostLoader.js!../../../../node_modules/vue-loader/dist/index.js??ruleSet[0].use[0]!./Hello.vue?vue&type=style&index=0&id=614f9072&scoped=true&lang=css' (possible exports: )`

stof:
@Lyrkan given that webpack 5 supports non-JS chunks natively, do we even need the `delete-unused-entries-js-plugin.js` ? the solution could be to remove it.

>     * `webpack-manifest-plugin` (which doesn't seem to be maintained anymore)

Given that the plugin is unmaintained and that Encore needs to customize parts of it anyway (and is blocked from implementing https://github.com/symfony/webpack-encore/issues/573 due to the plugin implementation), wouldn't it be easier to move away from that plugin in favor of a dedicated plugin in Encore ?

ankurk91:
https://github.com/shellscape/webpack-manifest-plugin

ankurk91:
optimize-css-assets-webpack-plugin can be replaced by

https://github.com/webpack-contrib/css-minimizer-webpack-plugin/

stof:
@Lyrkan do you have any plan for when you will continue this work?

Lyrkan:
@stof Currently waiting for a new version of the webpack-manifest-plugin to be released since there is a new maintainer (https://github.com/shellscape/webpack-manifest-plugin/issues/222). Having our own plugin could also be a solution but it would be one more thing to maintain from our side and I fear that we won't have enough people willing to do it.

> given that webpack 5 supports non-JS chunks natively, do we even need the delete-unused-entries-js-plugin.js ? the solution could be to remove it.

We may still need it for the hack used by `Encore.copyFiles(...)`... but I'll have to do some tests since there has been some improvements in Webpack 5 regarding empty chunks removals.

> optimize-css-assets-webpack-plugin can be replaced by webpack-contrib/css-minimizer-webpack-plugin

Thanks, I'll look into that.

ankurk91:
I would like to see cache to be implemented.

https://github.com/webpack/changelog-v5/blob/master/guides/persistent-caching.md

This can greatly improve build time.

stof:
@ankurk91 this PR is focused on making existing Encore features work with webpack 5, to be able to migrate.
Configuring the webpack cache would indeed be great, but to me, this can be left as a follow-up work as the cache is not mandatory to migrate from v4 to v5.

mmarton:
Not sure if this needs any changes (but if it does this seems the right place) 

> LibSass and the packages built on top of it, including Node Sass, are deprecated https://sass-lang.com/blog/libsass-is-deprecated

stof:
sass-loader is compatible with the `sass` package, not only with `node-sass`

lib/loaders/css-extract.js
```diff
@@ -36,15 +36,11 @@ module.exports = {
         }
 
         // Default options
```
stof, line 38: the comment should be removed as you removed the documented variable

shellscape:
An update for you all here: https://github.com/shellscape/webpack-manifest-plugin/issues/222#issuecomment-736969974

FWIW the 3.0 RC has basically been tossed out the window and I'm reapplying fixes in there atop the refactored codebase. New version dropping soon.

Lyrkan:
Everything should be good now!

The failing tests are unrelated (probably introduced by #865 since we use the `sass-loader` version in them).

weaverryan:
@Lyrkan could you fix up the yarn.lock conflict one more time? Thanks!

lib/config-generator.js
```diff
@@ -632,7 +542,7 @@ class ConfigGenerator {
 
     buildWatchOptionsConfig() {
         const watchOptions = {
-            ignored: /node_modules/
+            ignored: 'node_modules'
```
weaverryan, line 545: Can you explain this change?

weaverryan (commented):
I'm still reviewing, but it's looking really good so far!

weaverryan (commented):
Review complete! This is really nice, and it deletes more code than it adds - always wonderful :). There is one significant thing that I'm wondering about for Webpack 5 - Asset Modules - https://webpack.js.org/guides/asset-modules/

It *seems* like using file-loader, url-loader, etc for images, fonts, etc is now "non-standard", and instead we should use Asset Modules. If I'm correct, ideally we would refactor to use that immediately so that we can use the breaking changes in the release to remove file-loader, url-loader, etc. But, we could do this in a different PR.

A few other things I'm thinking about:

A) I'm wondering if we could start using webpack-manifest-plugin also to generate the entrypoints.json file. I have no tried it yet, but the idea would be to use the `useEntryKeys` option. This would mean we would need to use the plugin 2 times... once for manifest.json and again for entrypoints.json... which I think is ok, but a bit non-standard. We might also (but not sure) need to remove or move the "sha-" checksums - I'm not sure if those could be done with webpack-manifest-plugin

B) A lot of defaults were changed on `stats`. We should revisit if we can remove much/all of our "overriding" and allow encore to have a more standard output.

C) Finally, it might be nice to have a new method for the persistent caching - maybe `.enableBuildCache()`, with some config to allow them to add more files / configure the cache. We could, by default, at least add `webpack.config.js` to the cache... so it "should" work out of the box for most users.

Cheers!

package.json
```diff
@@ -33,23 +33,23 @@
...
     "resolve-url-loader": "^3.1.2",
     "semver": "^7.3.2",
     "style-loader": "^1.1.3",
-    "terser-webpack-plugin": "^1.1.0",
+    "terser-webpack-plugin": "^4.2.3",
     "tmp": "^0.2.1",
-    "webpack": "^4.36.0",
-    "webpack-cli": "^3.0.0",
+    "webpack": "^5",
+    "webpack-cli": "^4",
```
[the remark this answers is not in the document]
Lyrkan: IIRC when I initially wrote that webpack-cli was in beta and the way it started webpack prevented it from reading the runtime context. The issue wasn't there anymore when I upgraded it to the stable version.

lib/config-generator.js
```diff
@@ -454,8 +424,6 @@ class ConfigGenerator {
         // Dump the manifest.json file
         manifestPluginUtil(plugins, this.webpackConfig);
 
-        versioningPluginUtil(plugins, this.webpackConfig);
```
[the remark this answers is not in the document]
Lyrkan: `HashedModuleIdsPlugins` is outdated and has been replaced by `optimization.moduleIds: 'deterministic'` (https://webpack.js.org/migrate/5/#update-outdated-options).

Based on this page that's the default behavior for production builds: https://github.com/webpack/changelog-v5/blob/master/README.md#deterministic-chunk-module-ids-and-export-names

lib/config-generator.js
```diff
@@ -510,29 +476,12 @@ class ConfigGenerator {
                 terserPluginUtil(this.webpackConfig),
                 optimizeCssAssetsUtil(this.webpackConfig)
             ];
-        } else {
-            // see versioning.js: this gives us friendly module names,
-            // which can be useful for debugging, especially with HMR
-            optimization.namedModules = true;
```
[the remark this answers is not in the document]
Lyrkan: `optimization.namedModules` has been replaced by `optimization.moduleIds: 'named'`: https://webpack.js.org/migrate/5/#update-outdated-options

This is the default value for development builds: https://webpack.js.org/configuration/mode/#mode-development

lib/config-generator.js
Lyrkan: When I did that change using a RegExp was not allowed for that option anymore.

It seems that it was added back by https://github.com/webpack/webpack/pull/11498.

test/functional.js
```diff
@@ -403,24 +421,24 @@ describe('Functional tests using webpack', function() {
                     if (!process.env.DISABLE_UNSTABLE_CHECKS) {
```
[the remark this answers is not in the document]
Lyrkan: Most likely, hashes will still change depending on the versions of the dependencies used by our tests.

test/functional.js
```diff
@@ -1249,22 +1075,6 @@ module.exports = {
...
-                // check that babel transformed the arrow function
-                webpackAssert.assertOutputFileDoesNotContain(
-                    'main.js',
-                    '=>'
-                );
-
-                done();
-            });
-        });
-
```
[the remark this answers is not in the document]
Lyrkan: Well, it was a problem, but I'm not sure why I entirely removed it 😅

The test was failing because Babel does not apply to all the code that is generated by Webpack... and by default that one will now contain arrow functions.

I added the test back but made it check for `class` replacement instead.

Lyrkan:
> There is one significant thing that I'm wondering about for Webpack 5 - Asset Modules - webpack.js.org/guides/asset-modules
> 
> It seems like using file-loader, url-loader, etc for images, fonts, etc is now "non-standard", and instead we should use Asset Modules. If I'm correct, ideally we would refactor to use that immediately so that we can use the breaking changes in the release to remove file-loader, url-loader, etc. But, we could do this in a different PR.

Hm, I kept `file-loader` and `url-loader` to avoid adding other sudden breaking changes.
Both of them still work and we'll actually still need the `file-loader` for `Encore.copyFiles()`.
I think we could try to progressively remove them later, maybe by using asset modules by default unless the user calls `configureUrlLoader(...)`.

> A) I'm wondering if we could start using webpack-manifest-plugin also to generate the entrypoints.json file. I have no tried it yet, but the idea would be to use the useEntryKeys option. This would mean we would need to use the plugin 2 times... once for manifest.json and again for entrypoints.json... which I think is ok, but a bit non-standard. We might also (but not sure) need to remove or move the "sha-" checksums - I'm not sure if those could be done with webpack-manifest-plugin

Hm, I haven't really thought about that to be honest, in my opinion both plugins do different things.
For the checksums I know that it could now be done automatically by the `assets-webpack-plugin` (that wasn't the case when we added that feature), but not with the same format than the one we currently have.

> B) A lot of defaults were changed on stats. We should revisit if we can remove much/all of our "overriding" and allow encore to have a more standard output.

👍

> C) Finally, it might be nice to have a new method for the persistent caching - maybe .enableBuildCache(), with some config to allow them to add more files / configure the cache. We could, by default, at least add webpack.config.js to the cache... so it "should" work out of the box for most users.

Things can easily go wrong with persistent caching... even only enabling it for the config file could lead to some issues if not done properly (for instance if the config use values coming from environment variables, which is far from an edge case imo: https://github.com/webpack/changelog-v5/blob/master/guides/persistent-caching.md#version).

I also think that we should provide some way to enable/configure it in Encore but that should probably be treated separately after thinking for a while about how we really want to handle that.

lib/config-generator.js
weaverryan: I see it! We should probably change this back then?

weaverryan:
> Hm, I kept file-loader and url-loader to avoid adding other sudden breaking changes.
Both of them still work and we'll actually still need the file-loader for Encore.copyFiles().

The migration page - https://webpack.js.org/migrate/5/ - suggests that we should migrate away to use asset modules:

> Several commonly-used loaders like raw-loader, url-loader, and file-loader should be replaced using the new built-in Asset Modules feature.

So, I think we should make this change now (or at least try it). However, we can do it in another PR after we merge this. And we can keep using file-loader for `Encore.copyFiles()` - I don't see any issue with that.

> Hm, I haven't really thought about that to be honest, in my opinion both plugins do different things.
For the checksums I know that it could now be done automatically by the assets-webpack-plugin (that wasn't the case when we added that feature), but not with the same format than the one we currently have.

It's something we can investigate after we merge this PR. If we can make Encore smaller and more standard, cool :)

> I also think that we should provide some way to enable/configure it in Encore but that should probably be treated separately after thinking for a while about how we really want to handle that.

Concerns heard! Let's definitely consider it in another PR. It would be a nice "sweet new feature!" to have when we release the Webpack5-powered Encore ;).

weaverryan:
Thank you for this HUGE work @Lyrkan! A PR created more than 1 year agoto track Webpack 5, and now I caan happily merge it :). I will create an issue to track the final few things that we're considering before releasing this new Webpack-5-powered version. Thanks!

Kocal:
Nice!! :D

lib/config-generator.js
weaverryan: FYIU - this has been reverted

PR #649, opened 2019-10-08
Allow to use the [N] placeholder in copyFiles()

### LINKED ISSUE #647: copyFiles does not support pattern match placeholder
In https://github.com/webpack-contrib/file-loader#placeholders there's the pattern match placeholder `[N]` documented:
> The n-th match obtained from matching the current file name against the `regExp`.

I would expect that `copyFiles` would use the matches from `pattern`, though `[N]` is not replaced. For example in the following code
```
Encore.copyFiles({
     from: './assets/',
     to: '[path][0]',
     pattern: /.*/,
});
```
`[0]` should be replaced by the entire tested string, though the resulting filename is always `[0]`.

Lyrkan:
This PR should fix #647 by passing the `regExp` option to the `file-loader` used by `Encore.copyFiles()`.

It's a bit hacky since it changes the loader's options from the loader itself, but we should probably be fine for a while (I also tested it with `webpack@5.0.0-alpha.29` and the latest version of `file-loader@4.2.0`).

### DIFF
--- a/lib/webpack/copy-files-loader.js
+++ b/lib/webpack/copy-files-loader.js
@@ -56,7 +56,7 @@ module.exports.default = function loader(source) {
     // The "source" part of the regexp is base64 encoded
     // in case it contains characters that don't work with
     // the "inline loader" syntax
-    const pattern = new RegExp(
+    const pattern = options.regExp || new RegExp(
         Buffer.from(options.patternSource, 'base64').toString(),
         options.patternFlags
     );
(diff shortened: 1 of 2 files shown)


Lyrkan:
Failing tests are unrelated (see #650)

weaverryan:
Sorry for the delay. This is a pretty wild, in-depth solution. But it's well-covered with a test and I appreciate you checked it in Webpack 5 alpha!

PR #650, opened 2019-10-08
Fix hashes in the "Copy with versioning enabled" test

Lyrkan:
Fix a broken test on develop following the merge of #648.

Not sure why the main.js and runtime.js hashes changed, but their content stayed exactly the same.

### DIFF
--- a/test/functional.js
+++ b/test/functional.js
@@ -1989,8 +1989,8 @@ module.exports = {
                         expect(config.outputPath).to.be.a.directory()
                             .with.files([
                                 'entrypoints.json',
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks for catching that for me @Lyrkan :)

stof:
> Not sure why the main.js and runtime.js hashes changed, but their content stayed exactly the same.

That's because webpack 4 is not hashing the output file. The hash is based on input files and some parts of the webpack config.
This will probably change in webpack 5 as they are working on better long-term caching.

PR #669, opened 2019-11-17
Fixing postcss error: newer versions seem to have a "Syntax Error: name

weaverryan:
When enabling PostCSS, but not having a PostCSS config file, we attempt to catch the error so we can print out an example PostCSS file. In a project I just started, this was not happening. It appears that the `e.name` was `SyntaxError`. Since we're already looking for a very specific error message, I think the name was overkill anyway (though I did check for `e.message` just as an extra sanity check).


Lyrkan:
> It also makes me wonder why we have lodash in our dependencies, it doesn't seem needed at all.

Found out why: it was needed when we embedded a patched version of the `webpack-manifest-plugin`.

We could probably remove it safely now and also move `fs-extra` to `devDependencies`.

weaverryan:
Nice research! I did both things - tests pass locally, let's see how if the CI is happy :)

PR #675, opened 2019-11-23
Added support for image optimiziation with image-webpack-loader

aschempp:
as discussed with @weaverryan here's a first POC for adding image optimization. Looking for general feedback as this is my first time working with Encore, so I'm not sure if my approaches make sense.

~~The method and feature name are kinda confusing due to the dependencies. We're using `image-webpack-loader`, which internally uses `imagemin` to _optimize_ images. Not sure if stuff should be named after the loader (as of now), or rather something like `enableImageOptimization()` or name the feature `imagemin` instead of `image-webpack`…~~

To enable image optimization, install `image-webpack-loader` and call `enableImagemin()` in the Encore config. Further image configuration is available by passing options to the method exactly as in [`image-webpack-loader`](https://github.com/tcoopman/image-webpack-loader#options).

```js
Encore.enableImagemin({
    mozjpeg: {
        progressive: true,
        quality: 65
    },
    // optipng.enabled: false will disable optipng
    optipng: {
        enabled: false,
    },
    pngquant: {
        quality: [0.65, 0.90],
        speed: 4
    },
    gifsicle: {
        interlaced: false,
    },
    // the webp option will enable WEBP
    webp: {
        quality: 75
    }
});
```

#SymfonyHackday


lib/config-generator.js
```diff
@@ -400,6 +401,14 @@ class ConfigGenerator {
             }));
         }
 
+        if (this.webpackConfig.useImageWebpackLoader) {
+            rules.push(applyRuleConfigurationCallback('image-webpack', {
+                test: /\.(gif|png|jpe?g|svg)$/i,
+                loader: 'image-webpack-loader',
+                options: imageWebpackLoaderUtil.getOptions(this.webpackConfig)
+            }));
+        }
```
Lyrkan: The new loader should probably be part of the existing "images" rule:

https://github.com/symfony/webpack-encore/blob/53df05e04f87045f7fbe064b7802f49e26feaff9/lib/config-generator.js#L302-L306

See https://webpack.js.org/concepts/loaders/#configuration for how loaders are evaluted.

index.js
```diff
@@ -1097,6 +1097,20 @@ class Encore {
         return this;
     }
 
+    /**
+     * Enables image optimization with imagemin.
+     *
```
Lyrkan: Maybe add a note there that it will only be enabled for productions build by default + an example of how to use it with/without a callback.

aschempp:
I finally got around to update this PR. It's pretty much a rewrite of the original implementation, thanks to the suggestions @Lyrkan. I have also update the initial PR description.

One caveat is that the signature of the images loaders has changes due to the nature of Webpack and chaining loaders. I'm not sure if this is really a problem, and we could try to keep BC of the signature for as long as Imagemin is not enabled if necessary?

aschempp:
failing tests don't see related to the PR 🙃

index.js
```diff
@@ -1160,6 +1160,50 @@ class Encore {
...
+     *     webp: {
+     *         quality: 75
+     *     }
+     * });
+     * ```
+     *
+     * @param {object} options
+     * @return {Encore}
+     */
+    enableImagemin(options = {}) {
```
Lyrkan: That would work but we have some kind of pattern to set loaders' options using a callback, I think it would be better for consistency to use it there too.

It shouldn't require many changes, basically:
* have `enableImagemin` accept a function as its first argument instead of an object
* store this function instead of an object in `lib/WebpackConfig.js`
* use `applyOptionsCallback` in `lib/loaders/images.js` to apply that function on a "default options" object

lib/WebpackConfig.js
```diff
@@ -725,6 +734,18 @@ class WebpackConfig {
...
+    enableImagemin(options = {}) {
+        this.useImagemin = true;
+
+        for (const optionKey of Object.keys(options)) {
+            if (!(optionKey in this.imageminOptions)) {
+                throw new Error(`Invalid option "${optionKey}" passed to enableImagemin(). Valid keys are ${Object.keys(this.imageminOptions).join(', ')}`);
+            }
+
+            this.imageminOptions[optionKey] = options[optionKey];
+        }
```
Lyrkan: That kind of validation is usually only done for Encore-specific options (the reason being that it would, for instance, disallow the valid `plugins` option in this case).

index.js
aschempp: I actually implemented it that way first 😂 
Then I noticed `enablePreactPreset` uses an options object, `enableTypeScriptLoader` uses a callback, `eslintLoaderOptionsOrCallback` supports both types … I wasn't sure about the best or *latest* approach here?

lib/WebpackConfig.js
aschempp: so we shouldn't validate the configuration at all, right?

index.js
Lyrkan: Usually it goes like this:
* Encore-specific options are set using objects that are validated directly when the related method is called (that's the case for `enablePreactPreset()`).
* Loaders/plugins options are set using callbacks that are applied on a default options object when building the Webpack config. Sometimes (for instance in `enableTypeScriptLoader`) an object is also allowed but in reality it's just a shorthand that creates a callback. In this case we don't really check what the callback does unless we have a really good reason to do it (for instance if we know for sure that setting an option breaks another feature)

lib/WebpackConfig.js
Lyrkan: > so we shouldn't validate the configuration at all, right?

Exactly

index.js
aschempp: Adjusted accordingly in 700b608eb478b7242faf0d624cbbbf16f9bb6fe5

aschempp:
any feedback on this PR? 🙃

weaverryan:
Sorry for the delay - I have this on my list to check soon - it's something that I think would be super cool :)

lib/loaders/images.js
```diff
@@ -0,0 +1,63 @@
...
+        const imagesLoaders = [{
+            loader: loaderName,
+            options: loaderOptions
+        }];
+
+        if (webpackConfig.useImagemin) {
+            loaderFeatures.ensurePackagesExistAndAreCorrectVersion('imagemin');
+
+            const imageminOptions = {
+                disable: !webpackConfig.isProduction(),
```
weaverryan: Hmm, I'm not sure about this. I could see a user enabling this feature while developing, then checking the images and saying "yep! they look ok still!. Then on deploy, their images are now a "lower quality". I think we should just allow the user to control whether or not this is disabled entirely.

aschempp: 🤔  that should just be the default value, and one can just override that in the webpack/encore configuration?

weaverryan: If I call `Encore.enableImagemin()`, I would expect that my images would be optimized. NOT optimizing them is unexpected. So I think Encore needs to not have this `disable` option at all. Nice and simple: if they call `Encore.enableImagemin()`, we enable it. Then, if they choose to, they could decide to disable in dev.

I *do* realize why you're doing this... it's just too unexpected. One other option would be to add an `isEnabled` flag as the *first* argument to `Encore.enableImagemin()`. Then we would document you using it like this:

```
Encore.enableImagemin(Encore.isProduction());
```

weaverryan:
Ping! I think we're close - but there is one last big issue :)

lib/loaders/images.js
aschempp: Sorry for the delay. I decided to remove the disabled-in-dev default. You're probably right that it's unexpected, and it's easy to add. I thought about the additional option as well, but it just adds unnecessary complexity, I can achieve the same using an options object.

aschempp:
Finally, tests are fixed! Not sure how to fix the frozen yarn file (I'm not using yarn 🙈 )

Kocal:
> I'm not using yarn

But you should! :smile: 

The error happens because `--frozen-lockfile` actually check if the `package.json` and `yarn.lock` files are synchronized. If it's not the case, it fails.

Here, it's the case. You've added `image-webpack-loader` dependency in the `package.json` file (https://github.com/symfony/webpack-encore/pull/675/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) but the `yarn.lock` file has not been updated.

You need to run `yarn` and commit updates, but the next time please run `yarn add --dev image-webpack-loader`.

aschempp:
> > I'm not using yarn
> 
> But you should! 😄

haha, yeah, that's an _opinion_ 😝 

> You need to run `yarn` and commit updates, but the next time please run `yarn add --dev image-webpack-loader`.

I do basically understand why this happens (I'm familiar with lock files). I'm wondering about the work flow, do I need to rebase on every commit to master to re-generate the lock file? That seems pretty cumbersome, is there a documentation what these tests are for?

Kocal:
Yes the ideal is to rebase on master, `yarn.lock` conflicts won't be an issue since Yarn resolves them easily.

It seems that you already rebased your PR, so just run `yarn` and it should be fine.

stof:
> haha, yeah, that's an _opinion_ stuck_out_tongue_closed_eyes

no. That's a fact. The webpack-encore project uses Yarn (Classic) for its dependency management. So when contributing to the project, you need to use it.

aschempp:
> no. That's a fact. The webpack-encore project uses Yarn (Classic) for its dependency management. So when contributing to the project, you need to use it.

Obviously. I was referring to the 😄 emoji looking more like a general hint 😉 

`yarn.lock` updated in 14595cf0499ba6c6da9c8cadd9e2673c233a9c90

taylankasap:
What is this waiting for? Anything I (a random developer) can help with?

Rodrigo001-de:

infabo:
@Rodrigo001-de awesome non-related comment!

Rodrigo001-de:
@infabo sorry

aschempp:
Updated to fix conflicts. Any news on merging this?

georgringer:
hey @weaverryan it would be great if you could review this, let's merge that nice feature

joworeiter:
hi, anything new, when the feature gets merged?

jbty:
Hi, any update on this ?

Jupi007:
Hi, it is planed to fix conflicts and merge this soon?

aschempp:
Unfortunately, this PR has become unusable due to a lot of changes/improvements in Encore. I guess I can re-implement the feature but I don't like to spend the time if it's not gonna be merged for another year… @weaverryan ?

Jupi007:
> Unfortunately, this PR has become unusable due to a lot of changes/improvements in Encore. I guess I can re-implement the feature but I don't like to spend the time if it's not gonna be merged for another year… @weaverryan ?

That's sad, this feature could be awesome for many people.

Currently I manually optimize my images and just copy them into `/public/build`. Sorry for the offtopic @aschempp, but do you know how to manually integrate **image-webpack-loader** into encore?

aschempp:
> Sorry for the offtopic @aschempp, but do you know how to manually integrate **image-webpack-loader** into encore?

This is what I have in all my `webpack.config.js` (in addition to requiring `image-webpack-loader` in the `package.json`). Unfortunately it does not seem to work with `copyFiles()` but only assets that were required in scripts and stylesheets.

```
    // optimize and minify images
    .addLoader({
        test: /\.(gif|png|jpe?g|svg)$/i,
        use: [ 'image-webpack-loader' ]
    })
```

Jupi007:
Thanks a lot @aschempp :)

weaverryan:
Hi!

yes, sorry to let this completely get stuck - that’s my fault. Per the last comment from @aschempp, since this is quite conflicted now and enabling it is so straightforward on your app, another option is just to document it.

I’m going to close for now. If someone wants to reopen with the updated version - or make a docs PR, that would be great. Again, my apologies @aschempp - you did really nice work here, sometimes it’s hard for me to keep up - sorry :/

aschempp:
no hard feelings @weaverryan, one more thing off my plate myself 😆

PR #677, opened 2019-12-03
Updated - mini-css-extract-plugin

daymerfillad:
The latest versions of Vuetify need newer versions of mini-css-extract-plugin, they require version 0.8


weaverryan:
Closing for now - unless we can get some more feedback from the author :).

Thanks!

PR #680, opened 2019-12-16
Add Encore.addCacheGroup() method and depreciate Encore.createSharedEntry()

Lyrkan:
As discussed in https://github.com/symfony/webpack-encore/pull/645#issuecomment-538787669 our `Encore.createSharedEntry()` hack won't work anymore with Webpack 5.

Since it was mostly something that was added to make the transition from Webpack 3 to Webpack 4 less painful it's probably time to depreciate it and encourage people to use cache groups properly instead.

This PR adds a new `Encore.addCacheGroup()` method that, as its name implies, simply adds a new cache group to the config:

```js
Encore.addCacheGroup('vendor', {
  test: /[\\/]node_modules[\\/]react/
});
```

To make it a bit easier in case people want to directly reference things from the `node_modules` directory I also added a `node_modules` option which is basically a shorthand that sets the `test` option:

```js
Encore.addCacheGroup('vendor', {
  node_modules: ['react', 'react-dom']
});
```

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -14,6 +14,7 @@ const fs = require('fs');
 const crypto = require('crypto');
 const RuntimeConfig = require('./config/RuntimeConfig'); //eslint-disable-line no-unused-vars
 const logger = require('./logger');
+const regexpEscaper = require('./utils/regexp-escaper');
 
 /**
  * @param {RuntimeConfig|null} runtimeConfig
@@ -84,6 +85,7 @@ class WebpackConfig {
         this.manifestKeyPrefix = null;
         this.sharedCommonsEntryName = null;
         this.sharedCommonsEntryFile = null;
+        this.cacheGroups = {};
         this.providedVariables = {};
         this.configuredFilenames = {};
         this.aliases = {};
@@ -509,6 +511,8 @@ class WebpackConfig {
     }
 
     createSharedEntry(name, file) {
+        logger.deprecation('Encore.createSharedEntry() is deprecated and will be removed in a future version, please use Encore.splitEntryChunks() or Encore.addCacheGroup() instead.');
+
         if (this.shouldSplitEntryChunks) {
             throw new Error('Using splitEntryChunks() and createSharedEntry() together is not supported. Use one of these strategies only to optimize your build.');
         }
@@ -526,6 +530,36 @@ class WebpackConfig {
         this.sharedCommonsEntryFile = file;
 
         this.addEntry(name, file);
+    }
+
+    addCacheGroup(name, options) {
+        if (typeof name !== 'string') {
+            throw new Error('Argument 1 to addCacheGroup() must be a string.');
+        }
+
+        if (typeof options !== 'object') {
+            throw new Error('Argument 2 to addCacheGroup() must be an object.');
+        }
+
+        if (!options['test'] && !options['node_modules']) {
+            throw new Error('Either the "test" option or the "node_modules" option of addCacheGroup() must be set');
+        }
+
+        if (options['node_modules']) {
+            if (!Array.isArray(options['node_modules'])) {
+                throw new Error('The "node_modules" option of addCacheGroup() must be an array');
+            }
+
+            options.test = new RegExp(`[\\\\/]node_modules[\\\\/](${
+                options['node_modules']
+                    .map(regexpEscaper)
+                    .join('|')
+            })[\\\\/]`);
+
+            delete options['node_modules'];
+        }
+
+        this.cacheGroups[name] = options;
     }
 
     copyFiles(configs = []) {
--- a/lib/config/validator.js
+++ b/lib/config/validator.js
@@ -29,6 +29,8 @@ class Validator {
         this._validateDevServer();
 
         this._validateSharedEntryName();
+
+        this._validateCacheGroupNames();
     }
 
     _validateBasic() {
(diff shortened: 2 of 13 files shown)


lib/WebpackConfig.js
```diff
@@ -500,6 +502,8 @@ class WebpackConfig {
     }
 
     createSharedEntry(name, file) {
+        logger.deprecation('Encore.createSharedEntry() is deprecated and will be removed in a future version, please use Encore.addCacheGroup() instead.');
```
weaverryan:
```suggestion
        logger.deprecation('Encore.createSharedEntry() is deprecated and will be removed in a future version, please use Encore.splitEntryChunks() or Encore.addCacheGroup() instead.');
```

WDYT?

lib/WebpackConfig.js
```diff
@@ -519,6 +523,36 @@ class WebpackConfig {
...
+        if (options['node_modules']) {
+            if (!Array.isArray(options['node_modules'])) {
+                throw new Error('The "node_modules" option of addCacheGroup() must be an array');
+            }
+
+            options.test = new RegExp(`[\\\\/]node_modules[\\\\/](${
+                options['node_modules']
+                    .map(regexpEscaper)
+                    .join('|')
+            })[\\\\/]`);
```
weaverryan: Wow :)

lib/config/validator.js
```diff
@@ -75,6 +77,14 @@ class Validator {
             logger.warning(`Passing "${this.webpackConfig.sharedCommonsEntryName}" to createSharedEntry() is not recommended, as it will override the built-in cache group by this name.`);
         }
     }
+
+    _validateCacheGroupNames() {
+        for (const groupName of Object.keys(this.webpackConfig.cacheGroups)) {
+            if (['vendors', 'default'].includes(groupName)) {
```
weaverryan: I think it might be called `defaultVendors` internally? https://github.com/webpack/webpack/blob/f5e69fcc9b4142aadf8588e96f5e80d6b7cc5320/lib/config/defaults.js#L563-L574

lib/WebpackConfig.js

lib/config/validator.js
Lyrkan: Hmm, I put `vendors` because we currently have a validation on `createSharedEntry()` including that group: https://github.com/symfony/webpack-encore/blob/bdf8ba46cc447c84ed2f4daf80c16122be191189/lib/config/validator.js#L73-L77

But you seem to be correct :)

lib/config/validator.js
```diff
@@ -75,6 +77,14 @@ class Validator {
             logger.warning(`Passing "${this.webpackConfig.sharedCommonsEntryName}" to createSharedEntry() is not recommended, as it will override the built-in cache group by this name.`);
         }
     }
+
+    _validateCacheGroupNames() {
+        for (const groupName of Object.keys(this.webpackConfig.cacheGroups)) {
+            if (['defaultVendors', 'default'].includes(groupName)) {
```
stof: should we also include `vendors`, which is the name in webpack 4 ?

stof: should we also warn if you pass `this.webpackConfig.sharedCommonsEntryName` ?

Lyrkan: > should we also include vendors, which is the name in webpack 4 ?

Oh, so that's why we previously used `vendors` in `_validateSharedEntryName()`!
I couldn't find a reference to it in the doc anymore since the following page already got updated with `defaultVendors`: https://webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks

I think I'll add it and we'll remove it when we upgrade to Webpack 5.

> should we also warn if you pass `this.webpackConfig.sharedCommonsEntryName` ?

We could but I'm not sure it'll be really useful... people will probably either:

* start a new project and directly use `addCacheGroup()` since `createSharedEntry()` throws a deprecation warning
* see the deprecation warning and try to replace their `createSharedEntry()` calls by `addCacheGroup()`

Doesn't cost much to add a warning though.

weaverryan:
Thank you @Lyrkan!

XWB:
@Lyrkan @weaverryan Can you guys give an example how I should migrate this:

```
Encore
    .createSharedEntry('app', './assets/js/app.js')
;
```

To `Encore.addCacheGroup()`?

The [documentation](https://symfony.com/doc/current/frontend/encore/shared-entry.html) still refers to `createSharedEntry`.

Lyrkan:
@XWB If you can, use `Encore.splitEntryChunks()` as explained [here](https://symfony.com/doc/current/frontend/encore/split-chunks.html). It will automatically handle common code and split your files accordingly.

This requires reading the generated `entrypoints.json` file to know which file(s) you need to include into your pages but:

* if you can use the Webpack Encore Bundle that's also done automatically
* if you can't it should still be relatively easy to read it and generate `script`/`link` tags based on it

If you can't use `splitEntryChunks()` that'll depend on what the content of your `app.js` is:

* if it's only a bunch of vendor imports you can create a cache groupe and list them in the `node_modules` option (note that it will only add those package to the group, not sub-dependencies)
* if you also have custom code you can use the `test` option and make it match that files + its dependencies (you can use a function instead of a RegExp is that's easier).

Note that `addCacheGroup()` does not do much apart from adding [cache groups to Webpack](https://webpack.js.org/plugins/split-chunks-plugin/#split-chunks-example-1) with some default options (see the doc [here](https://github.com/symfony/webpack-encore/blob/ce8d221dc7160f57d0e96be9bd750e560a31c650/index.js#L479-L519)) and shortcuts (well, only one currently: the `node_modules` option).

XWB:
Thank you @Lyrkan :)

Issue #685, opened 2020-01-11
Webpack encore requires "babel-eslint" even if not used


david-wilson-mcn:
What i'm trying to use is encore, typescript, react, and linting typescript with using the '@typescript-eslint/parser' parser (since tslint is being deprecated). I got it all working, but my question is why do I still need the 'babel-eslint' package if i'm not using the babel eslint parser? Encore complains that I need it if I enable 'enableEslintLoader'. Here is the relevant parts of my webpack.config.js:
```
    .enableTypeScriptLoader()
    .enableForkedTypeScriptTypesChecking()
    .enableEslintLoader(options => {
        options.parser = '@typescript-eslint/parser'
    })
    .configureLoaderRule('eslint', loader => {
        loader.test = /\.(jsx?|tsx?)$/
    })
    .enableReactPreset()
```
Here is the relevant parts of my .eslintrc.js file:
```
    parser: "@typescript-eslint/parser",
    plugins: [
        "@typescript-eslint",
        "react"
    ],
    extends: [
        "eslint:recommended",
        "plugin:@typescript-eslint/eslint-recommended",
        "plugin:@typescript-eslint/recommended",
        "plugin:react/recommended"
    ],
    parserOptions: {
        ecmaVersion: 2018,
        sourceType: 'module',
        ecmaFeatures: {
            jsx: true
        }
    }
```
If I don't have the package and run 'yarn dev-server' I get:
```
Running webpack-dev-server ...

  WARNING   Be careful when using Encore.configureLoaderRule(), this is a low-level method that can potentially break Encore and Webpack when not used carefully.
  Error: Install eslint-loader & babel-eslint to use enableEslintLoader()
    yarn add eslint-loader@^2.1.2 babel-eslint@^10.0.1 --dev

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command
```

Kocal:
Hi,

This is because Encore already configure ESLint a bit by setting [`parser` to `babel-eslint`](https://github.com/symfony/webpack-encore/blob/master/lib/loaders/eslint.js#L26) and so [`babel-eslint`](https://github.com/symfony/webpack-encore/blob/master/lib/features.js#L106) is required when using `enableEslintLoader`.

We already had an issue with because of this `babel-eslint` already configured (#574) and I've opened a RFC (#657) to make Encore stop configuring ESLint and let the end-user have a full control on ESLint configuration.

[...]

Issue #659, opened 2019-10-21
Parsing Error: Unexpected token typescript inside Vue files


masterofnothing11:
Hello, first of all I would like to state I know this is not the right place to make such question but I have tried other means but got no response. This leaves me no choice other than to come to the repository itself. 

I'm getting the following error when watching for files rebuild from ESLint

```
/var/www/assets/js/app/pages/ErpFlex.vue
  140:10  error  Parsing error: Unexpected token

   9 |   @Getter getTotal: number;
  10 | 
> 11 |   private resendOrder: any = null;
     |           ^
  12 |   private loadingOrders: boolean = true;
  13 |   private callout: object = {
  14 |     show: false,

```

In my webpack configuration I have added `typeScriptLoader`:

```
.enableTypeScriptLoader(function (typeScriptConfigOptions) {
        typeScriptConfigOptions.transpileOnly = true;
        typeScriptConfigOptions.configFile = path.resolve(__dirname, 'tsconfig.json');
    })
```

My `tsconfig.json` file looks like:
```
{
  "exclude": [
    "vendor",
    "/node_modules/"
  ],
  "include": [
    "**/*.ts",
    "**/*.vue"
  ],
  "compilerOptions": {
    "target": "es6",
    "allowJs": true,                       /* Allow javascript files to be compiled. */
    "sourceMap": true,                     /* Generates corresponding '.map' file. */
    "strict": true,                           /* Enable all strict type-checking options. */
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true
  }
}

```

Probably I'm looking at something pretty stupid from my part here. Any help would be much appreciated

Lyrkan:
Hi @masterofnothing11,

You'll need to configure ESLint to use `typescript-eslint`: https://github.com/typescript-eslint/typescript-eslint#how-do-i-configure-my-project-to-use-typescript-eslint

This can be done by either using the options callback of `Encore.enableEslintLoader()` or through an `.eslintrc.*` file (note that in the latter case you'll currently have to `delete options.parser` in the `enableEslintLoader`'s callback to be able to override the parser from an external file)

masterofnothing11:
@Lyrkan Thank you for the answer. 
I have already done that with the following:

```
.enableEslintLoader(eslLinterLoaderOptions => {
        eslLinterLoaderOptions.configPath = path.resolve(__dirname, '.eslintrc.js');
[...]

Issue #657, opened 2019-10-18
RFC: Use .eslintrc.* files for ESLint configuration


Kocal:
This issue is a following of the discussion from #574 and with @weaverryan on Slack.

### The problem

Currently, Encore configure ESLint (eslint-loader) to [use `babel-eslint` parser](https://github.com/symfony/webpack-encore/blob/master/lib/loaders/eslint.js#L26). 
This is nice, because now ESLint uses Babel to parse files, and it's easy and transparent for the end-user.

However, this should not be configured by Encore itself. Here what the [`eslint-loader` documentation](https://github.com/webpack-contrib/eslint-loader#options) says:

> Note that the config option you provide will be passed to the `CLIEngine`. This is a different set of options than what you'd specify in `package.json` or `.eslintrc`. 

Some options (like `parser`) are common to the two sets of options. However, I think that some of those options should not be passed `eslint-loader` options, but only in `.eslintrc`/`package.json` files, where the end-user configurations should belongs.

Normally forcing `babel-eslint` should not be a problem because it allows ESLint to parse _more files_, but it is when configuring ESLint to lint Vue files (#574) because [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue) use [`eslint-vue-parser`](https://github.com/mysticatea/vue-eslint-parser) for `parser` option, not `babel-eslint`, which then is configured in `parserOptions` object. See [this link](https://eslint.vuejs.org/user-guide/#what-is-the-use-the-latest-vue-eslint-parser-error) for more information.
It is also a problem when you are trying to lint TypeScript code which require `@typescript-eslint/parser`, see #685.	

And since options passed to `eslint-loader` overrides the external configuration (`.eslintrc.*` files), then we have to do this to let ESLint lint Vue files properly:
```js
Encore
  .enableEslintLoader(options => {
    delete options.parser;
  }, { 
    lintVue: true,
  })
```

This is clearly not ideal. Also, we can think _But if `lintVue` is true, then Encore can delete `options.parser` itself?_ but no because we are again interfering with ESLint configuration that should be configured by the end-user. 
What will happens if someone create a new ESLint plugin for Vue files but which required `babel-eslint` parser? (I don't think it's possible, but why not 😄).

We also think it's a better solution to configure ESLint outside Encore, with `.eslintrc.*` files, because it can be used by other tools like PhpStorm/VS Code which provide ESLint integration. Having an IDE ESLint integration is not possible when configuring ESLint with Encore. 

[...]

Issue #656, opened 2019-10-18
Use the latest vue-eslint-parser


masterofnothing11:
First of all I would like to state that I'm fully aware this is a question to stackoverflow but since the community around Encore is not that big I see no options other than to come to the repository itself. 

Enviroment:

- Docker version 19.03.3
- Linux Mint 19.1 Tessa

Basically I'm getting the famous error ` error  Use the latest vue-eslint-parser. See also https://vuejs.github.io/eslint-plugin-vue/user-guide/#what-is-the-use-the-latest-vue-eslint-parser-error  vue/html-indent`. I have tried all the solutions that I could find on Google but nothing seems to work. 

`.eslintrc.js`

```
module.exports = {
    "parser": "vue-eslint-parser",
    "parserOptions": {
        parser: "babel-eslint",
        "allowImportExportEverywhere": true,
        "ecmaFeatures": {
            "legacyDecorators": true,
            "modules": true
        },
        "ecmaVersion": 6,
        "sourceType": "module",
    },
    "extends": [
        "plugin:vue/base",
        "airbnb-base"
    ],
    "rules": {
        ...
        "vue/html-indent": [2, 4],
        ...
    },
    "env": {
        browser: true,
        es6: true,
        node: true
    },
};
```

`webpack.config.js`
```
...
.enableEslintLoader(eslLinterLoaderOptions => {
        eslLinterLoaderOptions.configPath = path.resolve(__dirname, '.eslintrc.js');
        eslLinterLoaderOptions.ignorePath = path.resolve(__dirname, '.eslintignore');
        eslLinterLoaderOptions.cache = false;
    })
...
```

`package.json`
```
...
  "devDependencies": {
    "@symfony/webpack-encore": "^0.27.0",
    "babel-eslint": "^10.0.1",
    "copy-webpack-plugin": "^4.5.3",
    "eslint": "^6.4.0",
    "eslint-plugin-vue": "^5.2.2",
    "ts-loader": "^5.3.0",
    "tslint": "^5.16.0",
    "typescript": "^3.4.5"
  },
...
```

Every time I run the command `yarn encore --watch` ESLint does it thing and check the files and the error mentioned above is thrown in almost every file.

I have tried removing the `node_modules/` folder and also upgrading ESLint. 

Any help would be highly appreciated

Kocal:
Hi, it's because Encore force `parser` option to `babel-eslint` and it's not compatible with `eslint-plugin`-vue.

[...]

PR #687, opened 2020-01-12
Remove ESLint user-related config

Kocal:
Hi :hand:

This PR is a proposal for #657 implementation and should fix issues encountered in #473, #574, #656, #659, and #685.

As discussed in #657, it would be better if Encore didn't configure ESLint itself. It should be done by the end user in an configuration file (e.g.: `.eslintrc.js`)

ESLint's `parser` option is not configured by default anymore, and `babel-eslint` requirement has been removed. We can considering this as a breaking change, end users should now configure `parser: 'babel-eslint` themselves.

Method `Encore.enableEslintLoader('extends-name')` has been deprecated and undocumented, in order to prevent end users to configure ESLint through Encore.

A nice message is also displayed when no ESLint configuration is found:
Capture d’écran de 2020-01-12 11-15-09
I couldn't use `FriendlyErrorsPlugin` for this because error is not coming from Webpack. If someone has a better idea... :confused: 

**Pros:**
- end users have now full control hover ESLint configuration **by default**
- no need to `delete options.parser` when trying to use [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue) or [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser)
- IDEs will be able to integrate ESLint (if they support it)

**Cons:**
- end users should now configure `parser` option to `babel-eslint` themselves 
- end users will have to move their config to external config file, but it's ok

What do you think?
Thanks.

**EDIT:** tests failures are unrelated I think.

### DIFF
--- a/fixtures/js/eslint-es2018.js
+++ b/fixtures/js/eslint-es2018.js
@@ -1,0 +1,2 @@
+  const a = { x: 0, y: 1, z: 2 };
+  const { x, ...b } = a;
--- a/index.js
+++ b/index.js
@@ -1070,18 +1070,14 @@ class Encore {
      * // enables the eslint loaded using the default eslint configuration.
      * Encore.enableEslintLoader();
      *
-     * // Optionally, you can pass in the configuration eslint should extend.
-     * Encore.enableEslintLoader('airbnb');
-     *
      * // You can also pass in an object of options
      * // that will be passed on to the eslint-loader
      * Encore.enableEslintLoader({
-     *     extends: 'airbnb',
      *     emitWarning: false
      * });
      *
      * // For a more advanced usage you can pass in a callback
-     * // https://github.com/MoOx/eslint-loader#options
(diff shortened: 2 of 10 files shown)


Lyrkan:
Thanks @Kocal, I think that we **really** need this :)

> I couldn't use FriendlyErrorsPlugin for this because error is not coming from Webpack. If someone has a better idea... 😕

Not a big issue imo, the message looks fine.

> EDIT: tests failures are unrelated I think.

Yeah, the ones on AppVeyor have been there for a while now (and I still have no clue what causes them...), and the ones on Travis are related to TypeScript (a bit weird that they only happen on that job though).

Kocal:
Hum wait what, it looks like I've forgot to push some commits... I will push them tonight. :sweat_smile:

weaverryan:
Thank you @Kocal!

Issue #691, opened 2020-02-07
Document how to use TypeScript with Babel


Kocal:
Just a reminder to write a documentation for using TypeScript with Babel (and not with `ts-loader`).

See those links for more info:
- https://iamturns.com/typescript-babel/
- https://github.com/microsoft/TypeScript-Babel-Starter
- https://devblogs.microsoft.com/typescript/typescript-and-babel-7/

Kocal:
Closed since #694 has been merged.

PR #694, opened 2020-02-07
Add Encore.enableBabelTypeScriptPreset() to "compile" TypeScript with Babel

Kocal:
See #691, I thought it can be interesting to have a test here.

Using Babel to "compile" TypeScript is faster than using `ts-loader` or `tsc` directly, because in fact, it literally remove types annotations. 

To continue to check types, you have to run `tsc --emitDeclarationOnly` manually (or in a CI). But this is not part of the PR.

To migrate an already existing TypeScript app, you just have to configure `babel-loader` to run over `.tsx?` file like this:
```diff
Encore
-  .enableTypeScriptLoader()
+  .configureLoaderRule('javascript', loader => {
+    loader.test = /.(j|t)sx?$/; // let Babel to run over .tsx? files too
+  })
```

Install some dependencies: `yarn add --dev @babel/preset-typescript @babel/plugin-proposal-class-properties`.

And modify your Babel configuration:
```diff
{
    "presets": [
        "@babel/env",
+        "@babel/typescript"
    ],
+    "plugins": [
+        "@babel/proposal-class-properties"
+    ]
}
```

Maybe I can update `Encore.configureBabel()` and add an option to runs over TypeScript files too... like I did in #574, something like this:
```js
Encore
  .configureBabel(null, {
    typescript: true
  })
```

I've also changed the legacy import/export (`import a = require('...')` to `import a from '...'`). Because it's the legacy way (ES6 imports are very fine) and the Babel TypeScript was not compatible with them:
Capture d’écran de 2020-02-07 22-06-11
 

**EDIT :** Added `Encore.enableBabelTypeScriptPreset()` that do all the job for us! :)

```js
// simple usage
Encore.enableBabelTypeScriptPreset();

// configure TypeScript preset (https://babeljs.io/docs/en/babel-preset-typescript#options)
Encore.enableBabelTypeScriptPreset({
  isTSX: true; 
})
```

`Encore.enableBabelTypeScriptPreset()` can not be used aside `Encore.enableTypeScriptLoader()` or `Encore.enableForkedTypeScriptTypesChecking()`.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -109,6 +109,7 @@ class WebpackConfig {
         this.useEslintLoader = false;
         this.useTypeScriptLoader = false;
         this.useForkedTypeScriptTypeChecking = false;
+        this.useBabelTypeScriptPreset = false;
         this.useWebpackNotifier = false;
         this.useHandlebarsLoader = false;
 
@@ -647,6 +648,10 @@ class WebpackConfig {
     }
 
     enableTypeScriptLoader(callback = () => {}) {
+        if (this.useBabelTypeScriptPreset) {
+            throw new Error('Encore.enableTypeScriptLoader() can not be called when Encore.enableBabelTypeScriptPreset() has been called.');
+        }
+
         this.useTypeScriptLoader = true;
 
         if (typeof callback !== 'function') {
@@ -657,6 +662,9 @@ class WebpackConfig {
     }
 
     enableForkedTypeScriptTypesChecking(forkedTypeScriptTypesCheckOptionsCallback = () => {}) {
+        if (this.useBabelTypeScriptPreset) {
+            throw new Error('Encore.enableBabelTypeScriptPreset() can not be called when Encore.enableForkedTypeScriptTypesChecking() has been called.');
+        }
 
         if (typeof forkedTypeScriptTypesCheckOptionsCallback !== 'function') {
             throw new Error('Argument 1 to enableForkedTypeScriptTypesChecking() must be a callback function.');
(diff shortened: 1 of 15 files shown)


Kocal:
I've added a small TypeScript-specific code, and tests pass fine.
It should be ok now :+1:

Kocal:
I don't know if it's a _better_ way to _compile_ TypeScript with Babel, but that's an interesting alternative. 

But note that you lose TypeScript _power_ at compilation since every types are removed during transpilation. You need to run `tsc` manually to check your `.ts` files (it can be done in CI for example), **we should document it**.

However, `Encore.enableBabelTypeScriptPreset()` sounds nice if it can make our life easier 😄 

I will try to work on it tonight :)

Kocal:
@weaverryan I've added support to `Encore.enableBabelTypeScriptPreset()`

lib/WebpackConfig.js
```diff
@@ -657,6 +663,9 @@ class WebpackConfig {
     }
 
     enableForkedTypeScriptTypesChecking(forkedTypeScriptTypesCheckOptionsCallback = () => {}) {
+        if (this.useBabelTypeScriptPreset) {
+            throw new Error('Encore.enableBabelTypeScriptPreset() can not be called when Encore.enableForkedTypeScriptTypesChecking() has been called.');
```
Lyrkan: Shouldn't that be `Encore.enableForkedTypeScriptTypesChecking() can not be called when Encore.enableBabelTypeScriptPreset() has been called.`?

Not a big difference but since the exception will be thrown on the `enableForkedTypeScriptTypesChecking()` call it makes more sense to say that it's the "wrong" one.

Kocal: Ah yes, it was a real mess to deal with those error messages :D 

It has been fixed, thanks!

Kocal:
Hi @Lyrkan, thanks for your review.
I fixed the error message and added two tests in `config-generator.js`.

weaverryan:
This is really, really nice work. Thank you for an awesome new feature @Kocal!

jennevdmeer:
I ran into this trying to help someone on slack. #691 mentions to create a PR/issue for symfony-docs. Has said documentation ever been done? I could not find anything about it.

weaverryan:
I am guessing that's correct!

PR #697, opened 2020-02-13
Fix source maps being generated by default in dev

### LINKED ISSUE #695: Use of eval and Content Security Policy
I am using Webpack Encore for a very minimalistic usage of JS and CSS and noticed the JS when using dev-server (`encore dev-server`) contains eval everywhere, therefore it doesn't work with any reasonable Content Security Policy, where eval is the first and most important thing to be prohibited.

It does work fine with `encore production`, and one could just not use any Content Security Policy for development, yet that may lead to breaks in production because the CSP was not tested / realistically set in development, so ideally no eval is ever used anywhere. I managed to solve the usage of eval by changing the value of `devtool` like shown in my configuration below:

```js

Encore
    .setOutputPath('public/assets/')
    .setPublicPath('/assets')
    .addEntry('app', './assets/js/app.js')
    .splitEntryChunks()
    .enableSingleRuntimeChunk()
    .cleanupOutputBeforeBuild()
;

process.env.NODE_ENV = process.env.NODE_ENV || 'development';

Encore.configureWatchOptions(watchOptions => {
    watchOptions.poll = 250; // check for changes every 250 ms
});

encoreConfig = Encore.getWebpackConfig();

// This prevents the use of eval and makes
// webpack in development work with a Content-Security-Policy
encoreConfig.devtool = "source-map";

module.exports = encoreConfig;
```

[I found the solution in a webpack issue about the usage of eval in webpack](https://github.com/webpack/webpack/issues/5627#issuecomment-389492939), although there were multiple solutions there depending on what kind of packages people used, so finding that solution was a bit random.

I thought I would start a discussion here about making Webpack Encore work with a reasonable Content Security Policy by default, so the friction to implement a CSP becomes lower, and/or to add documentation about how one can configure Encore to work without eval and with a CSP. I am also not sure if my above approach is the only way of removing eval, or if `devtool` can be set through the Encore object. In general I think it would be nice to factor in the usage of CSP, and I could help with documentation if needed.

Lyrkan:
When `config.devtool` is not set (which is the current behavior with `enableSourceMaps(false)` or when the function is not called at all) the `eval` source maps type is used in development mode.

This PR sets the default value to `false` to make sure that no source map is being generated in those cases (fixes #695).

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -70,7 +70,8 @@ class ConfigGenerator {
             },
             plugins: this.buildPluginsConfig(),
             optimization: this.buildOptimizationConfig(),
-            watchOptions: this.buildWatchOptionsConfig()
+            watchOptions: this.buildWatchOptionsConfig(),
+            devtool: false,
         };
 
         if (this.webpackConfig.useSourceMaps) {
--- a/test/config-generator.js
+++ b/test/config-generator.js
@@ -125,7 +125,7 @@ describe('The config-generator function', () => {
             config.useSourceMaps = false;
 
             const actualConfig = configGenerator(config);
-            expect(actualConfig.devtool).to.be.undefined;
+            expect(actualConfig.devtool).to.be.false;
 
             expect(JSON.stringify(actualConfig.module.rules)).to.not.contain('?sourceMap');
         });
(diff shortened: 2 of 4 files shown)


weaverryan:
Thank you @Lyrkan!

PR #710, opened 2020-03-16
bump: style-loader version 1.X

Grafikart:
The webpack encore version style loader is too old and inject CSS using `<link>` instead of `<style>`. This solution loads style asynchronously cause lots of problem with JavaScript.

With this upgrade style-loader doesn't accept sourcemap anymore. SourceMap support is calculated from previous loaders (https://github.com/webpack-contrib/style-loader#source-maps)

### DIFF
--- a/lib/loaders/css-extract.js
+++ b/lib/loaders/css-extract.js
@@ -26,9 +26,6 @@ module.exports = {
             // style-loader instead.
             return [{
                 loader: 'style-loader',
-                options: {
-                    sourceMap: webpackConfig.useSourceMaps,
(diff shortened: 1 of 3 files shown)


Lyrkan:
> This makes sense to me. @Lyrkan do you see any issues?

@weaverryan Can't see any, it's probably one of the things we never updated ^^'
Breaking changes are listed there: https://github.com/webpack-contrib/style-loader/releases/tag/v1.0.0

weaverryan:
Thank you Jonathan!

PR #713, opened 2020-03-20
Fix failing 'Lowest versions' tests

Lyrkan:
I think it suddenly failed because a `@types/*` package required by another dependency got updated and was not compatible with our lowest allowed version of Typescript anymore.


weaverryan:
YOU ROCK! <3

PR #715, opened 2020-03-21
Added the possibility to configure the StyleLoader via the method Enc…

tooltonix:
…ore.configureStyleLoader()


index.js
```diff
@@ -965,6 +965,27 @@ class Encore {
         return this;
     }
 
+    /**
+     * Configure the css-loader.
```
Lyrkan:
```suggestion
     * Configure the style-loader.
```

index.js
```diff
@@ -965,6 +965,27 @@ class Encore {
...
+     *
+     * https://github.com/webpack-contrib/style-loader#Options
+     *
+     * ```
+     * Encore.configureStyleLoader(function(config) {
+     *     // change the config
+     *     // config.injectType: 'styleTag'
```
Lyrkan:
```suggestion
     *     // config.injectType = 'styleTag';
```

lib/loaders/css-extract.js
```diff
@@ -22,13 +23,15 @@ module.exports = {
...
         if (!webpackConfig.extractCss) {
+
+            const options = {};
+
             // If the CSS extraction is disabled, use the
             // style-loader instead.
             return [{
                 loader: 'style-loader',
-                options: {
-                    sourceMap: webpackConfig.useSourceMaps,
```
Lyrkan: Doesn't our version still requires the `sourceMap` option?
It was removed in `style-loader@1.0.0` and until #710 is merged we're using `style-loader@^0.21.0`.

index.js
tooltonix: I have corrected that now

index.js
tooltonix: I have corrected that now

lib/loaders/css-extract.js
tooltonix: I have included sourcemaps in the default options

Kocal:
Hi :wave: 

> Also I'm wondering if this should be added as a separate method or directly as a `disableCssExtraction()` argument (since it is the only method that enables the loader)

Hum, I am bit divided on this point but I think a separate method is better.

If we had use a `disableCssExtraction()` argument:
- yes `style-loader` is only used if CSS extraction is disabled, but it feels weird to me to configure a loader in a method which is not named `configure...Loader()`
- what if one day we use `style-loader` somewhere else? (but I can't imagine cases where it would be possible :stuck_out_tongue:)

But if we have a `configureStyleLoader`:
- that's nice, a method for configuring a loader
- but the loader is not active by default, only when the CSS extraction is disabled. We should warn the user about this

tooltonix:
Due to internal requirements, we use a different style loader, which we install per yarn, so that the options of the previous style loader are no longer applicable. In the latest version 1.1.3 of the StyleLoader, the loader automatically inject source maps when previous loader emit them.  

It is correct that the method disableCssExtraction() activates the StyleLoader, but I would override the options with a separate method configureStyleLoader(), as it is more consistent for me in context and name.

lib/loaders/css-extract.js
```diff
@@ -22,13 +23,17 @@ module.exports = {
      */
     prependLoaders(webpackConfig, loaders) {
         if (!webpackConfig.extractCss) {
+
+            const options = {
+                sourceMap: webpackConfig.useSourceMaps,
+            };
+
```
weaverryan: NOW this can be removed - #710 is merged. Sorry for the back and forth.

lib/loaders/css-extract.js
```diff
@@ -22,13 +23,17 @@ module.exports = {
      */
     prependLoaders(webpackConfig, loaders) {
         if (!webpackConfig.extractCss) {
+
+            const options = {
+                sourceMap: webpackConfig.useSourceMaps,
```
Kocal: #710 has been merged, you can remove this _again_ :D

index.js
```diff
@@ -965,6 +965,27 @@ class Encore {
         return this;
     }
 
+    /**
+     * Configure the style-loader.
+     *
+     * https://github.com/webpack-contrib/style-loader#Options
+     *
```
weaverryan: Can we add:

The style-loader is used only if you also call `Encore. disableCssExtraction()`.

index.js
```diff
@@ -965,6 +965,27 @@ class Encore {
...
+     * Encore.configureStyleLoader(function(config) {
+     *     // change the config
+     *     // config.injectType = 'styleTag';
+     * });
+     * ```
+     *
+     * @param {function} callback
+     * @returns {Encore}
+     */
+    configureStyleLoader(callback) {
```
weaverryan: Can you move this method next to `disableCssExtraction()`? Maybe we can mention `configureStyleLoader()` in the docs for `disableCssExtraction()` - I'm trying to make it easy for people to see that these 2 functions are related.

index.js
```diff
@@ -965,6 +965,27 @@ class Encore {
...
 
+    /**
+     * Configure the style-loader.
+     *
+     * https://github.com/webpack-contrib/style-loader#Options
+     *
+     * ```
+     * Encore.configureStyleLoader(function(config) {
+     *     // change the config
+     *     // config.injectType = 'styleTag';
```
weaverryan: This might not be the best example because this is the default value, right?

lib/loaders/css-extract.js
tooltonix: I've done that.

lib/loaders/css-extract.js
tooltonix: I've done that.

index.js
tooltonix: I've done that.

index.js
tooltonix: I've done that.

index.js
tooltonix: I've done that.

tooltonix:
> Hey @tooltonix!
> 
> This looks good to me - and I agree with the 2 separate methods, but I'm happy we had the conversation - I think it will help us with some docs to show that these methods are connected.
> 
> Only minor comments. But we _do_ need a functional test for this. See `functional.js` - there is already a case that includes `disableCssExtraction()`. We could duplicate that test case, maybe set the `attributes` option to `TESTING_ATTRIBUTES` and then assert that this string is present inside `main.js`.
> 
> Cheers!

Hi @weaverryan 

I'm sorry, but I haven't written any tests in Mocha yet. But I tried anyway and the test was successful, as long as I used the correct options of the style loader in version 1.1.3. I hope that the test is sufficient. Otherwise I ask for appropriate suggestions.

weaverryan:
Very excellent work! Thanks so much for this Patrick!

PR #718, opened 2020-04-01
Include the .pcss extension for PostCSS files

opdavies:
This PR adds support for the `.pcss` file extension for PostCSS files.

I've got this to work by overriding the CSS Loader `test` within my webpack.config.js using the `.configureLoaderRule()` as documented in https://www.oliverdavies.uk/articles/using-the-pcss-extension-with-webpack-encore.

But it would be nice to have this supported by default to make things simpler and not having to see the `Be careful when using Encore.configureLoaderRule()` message on each build.

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -245,6 +245,7 @@ class ConfigGenerator {
         // files.
         const cssExtensions = ['css'];
         if (this.webpackConfig.usePostCssLoader) {
+            cssExtensions.push('pcss');
             cssExtensions.push('postcss');
         }
 
--- a/test/config-generator.js
+++ b/test/config-generator.js
@@ -1340,7 +1340,7 @@ describe('The config-generator function', () => {
                 findRule(/\.(css)$/, actualConfig.module.rules);
             }).not.to.throw();
             expect(function() {
-                findRule(/\.(css|postcss)$/, actualConfig.module.rules);
+                findRule(/\.(css|pcss|postcss)$/, actualConfig.module.rules);
             }).to.throw();
         });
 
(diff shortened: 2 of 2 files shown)


opdavies:
@weaverryan Any advice on this? The tests are passing here, but failing locally if I run `npm run test` and don't update the versioned filenames.

```
1) Functional tests using webpack
   Basic scenarios.
     copyFiles() allows to copy files and folders
       Copy with versioning enabled:

  Uncaught AssertionError: expected '/Users/opdavies/Code/tmp/webpack-encore/test_tmp/oil16n/www/build' to have the same files as [ Array(6) ]
  + expected - actual

   [
     "entrypoints.json"
  -  "main.88a14d18.js"
  +  "main.ceeddab6.js"
     "manifest.json"
  -  "runtime.0c796a0d.js"
  +  "runtime.d2591ff7.js"
     "symfony_logo.ea1ca6f7.png"
     "symfony_logo_alt.f27119c2.png"
   ]
```

Lyrkan:
> The tests are passing here, but failing locally if I run npm run test and don't update the versioned filenames.

Did you retrieve the dependencies using npm instead of Yarn?
The project only has a `yarn.lock` file, so if you don't install them through Yarn you won't get the right versions.

opdavies:
Possibly. :)

I'll delete `node_modules`, re-download using Yarn and re-try.

opdavies:
I've re-installed with Yarn and ran `yarn run test` instead, and everything passes.

Thanks @Lyrkan!

weaverryan:
Thank you @opdavies!

PR #725, opened 2020-04-08
Add more configuration to Terser

romainneutron:
This is imported from create-react-app configuration.
Please note that this is a copy from a MIT source, so it should be compatible, see https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L221-L260

### DIFF
--- a/lib/plugins/terser.js
+++ b/lib/plugins/terser.js
@@ -22,6 +22,42 @@ module.exports = function(webpackConfig) {
         sourceMap: webpackConfig.useSourceMaps,
         cache: true,
         parallel: true,
+        terserOptions: {
+            parse: {
+                // We want terser to parse ecma 8 code. However, we don't want it
+                // to apply any minification steps that turns valid ecma 5 code
+                // into invalid ecma 5 code. This is why the 'compress' and 'output'
+                // sections only apply transformations that are ecma 5 safe
+                // https://github.com/facebook/create-react-app/pull/4234
+                ecma: 8,
+            },
+            compress: {
+                ecma: 5,
+                warnings: false,
+                // Disabled because of an issue with Uglify breaking seemingly valid code:
+                // https://github.com/facebook/create-react-app/issues/2376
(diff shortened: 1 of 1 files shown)


Lyrkan:
Hi @romainneutron,

Are we sure that we should use the same default values than `create-react-app` ?

The thing is, since `create-react-app` doesn't allow users to modify the Webpack config they provide they have to be extra careful about what they put in it to avoid as much edge cases as they can (such as making sure some optimizations are disabled because in some rare situations they could be an issue).

Some values seem to already be the current default ones in Terser:

* [`parse.ecma` defaults to `2017`](https://github.com/terser/terser#parse-options) (so ES8)
* [`compress.ecma` defaults to `5`](https://github.com/terser/terser#compress-options)
* [`compress.warning` defaults to `false`](https://github.com/terser/terser#compress-options)
* [`keep_classnames` defaults to `undefined`](https://github.com/terser/terser#minify-options) which based on the related comment disables the feature
* [`keep_fnames` defaults to `false`](https://github.com/terser/terser#minify-options)
* [`output.ecma` defaults to `5`](https://github.com/terser/terser#output-options)

And I'm not sure about some other changes:

* [`output.comments`](https://github.com/terser/terser#output-options) should, in my opinion, stay `true` by default
* `compress.comparisons`, `compress.inline` and `mangle.safari10` seem to be workarounds related to old issues that may have been fixed since then

Lyrkan:
@romainneutron Note that I'm not 100% against those kind of changes, so thank you for opening this PR, but we really have to be able to justify them.

For instance setting `ascii_only` to `true` seem to have some pros (it avoids issues related to emojis) but also some cons (it can increase build size for people that didn't have those issues).

romainneutron:
@Lyrkan Thanks for your message. No worries, I've closed my PR because after reading your comments, it makes sense that I was not understanding completely the philosophy behind encore. It fits me.

PR #726, opened 2020-04-08
Adapt fonts and images filenames given the useVersioning configu property

romainneutron:


Lyrkan:
Hi @romainneutron,

The current behavior is actually intended.

The issue with removing the hash is that if you import two files with the same name from different folders they'll end up overwriting each others during the build.

We also thought about adding the `[path]` placeholder to prevent that but it also had some cons (such as really ugly paths if you imported an image from your `node_modules`).

You can check the comments on this PR for more details: https://github.com/symfony/webpack-encore/pull/103#issuecomment-317135031

PR #732, opened 2020-04-17
feat: add ESLint 6 support

Kocal:
Someone on Slack had an issue with Encore and ESLint 6:
Capture d’écran de 2020-04-17 22-12-29

`CLIEngine#config` does not exists on ESLint 6 anymore, since [6ae21a4b](https://github.com/eslint/eslint/commit/6ae21a4bfe5a1566f787fbad798182a524b96d28#diff-3daec82d5218d64d8f2fed4d21f8e6ccL472-L486). 
Now we can use [`CLIEngine#getConfigForFile`](https://eslint.org/docs/developer-guide/nodejs-api#cliengine-getconfigforfile) to get the ESLint configuration.

Encore now supports both ESLint 5 and 6.

cc @a-menshchikov :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "chai-fs": "^1.0.0",
     "chai-subset": "^1.6.0",
     "core-js": "^3.0.0",
-    "eslint": "^5.15.2",
+    "eslint": "^5.15.2 || ^6.0.0",
     "eslint-loader": "^2.1.2",
     "eslint-plugin-header": "^1.0.0",
     "eslint-plugin-import": "^2.8.0",
--- a/lib/loaders/eslint.js
+++ b/lib/loaders/eslint.js
@@ -12,6 +12,7 @@
 const WebpackConfig = require('../WebpackConfig'); //eslint-disable-line no-unused-vars
 const loaderFeatures = require('../features');
 const applyOptionsCallback = require('../utils/apply-options-callback');
+const logger = require('../logger');
 
 function isMissingConfigError(e) {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "chai-fs": "^1.0.0",
     "chai-subset": "^1.6.0",
     "core-js": "^3.0.0",
-    "eslint": "^5.15.2",
+    "eslint": ">=5.15.2",
```
weaverryan, line 69: Should we do a “5 || 6” type of logic here? We aren’t necessarily compatible with some future version 7.

Kocal: That's true, let's go for it.

a-menshchikov, line 69: It seems as potential source of problems on 7.0 release. Maybe `^5.12.2 || ^6.0` will be more reliable?

Lyrkan: I think it would be better to keep `>=5.15.2`:

This is a dev dependency and we don't enforce `eslint` version so what we put here doesn't really matter except for our tests. It would help spot an incompatibility once v7 is released.

Kocal: Yep :+1:

weaverryan:
@Lyrkan and I are discussing the version. What I can’t remember is if the version here will affect the “yarn add” that’s displayed if eslint is missing. That was my main motivator for allowing 5 or 6 - I want to make sure that it doesn’t give you a command that would allow version 7 to be installed when it comes out (until we can verify we’re compatible). But I’m not at a computer at the moment to check :)

weaverryan:
Thank you @Kocal!

PR #733, opened 2020-04-17
build(deps-dev): update eslint-loader to ^3

Kocal:
eslint-loader ^3 is the last version to [support both ESLint 5 and 6](https://github.com/webpack-contrib/eslint-loader/blob/v3.0.4/package.json#L40) (related #732).

Given the [changelog for v3.0.0](https://github.com/webpack-contrib/eslint-loader/blob/master/CHANGELOG.md#300-2019-08-24), the only breaking changes are:
- drop support for Node < 8.9.0 (we did it too https://github.com/symfony/webpack-encore/pull/731)
- minimum supported webpack version is 4: not a problem
- minimum supported eslint version is 5: not a problem

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "chai-subset": "^1.6.0",
     "core-js": "^3.0.0",
     "eslint": "^5.15.2 || ^6.0.0",
-    "eslint-loader": "^2.1.2",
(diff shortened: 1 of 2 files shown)


weaverryan:
Thank you again @Kocal!

PR #736, opened 2020-04-19
feat: upgrade sass-loader support to ^8

### LINKED ISSUE #654: Support for sass-loader 8.0.0
**OS :** WSL Ubuntu 18.04
**webpack-encore :** 0.28.0

Since I upgraded my ` package.json`, **sass-loader** was updated to 8.0.0, and I get the following warning running `yarn encore dev` :

```
Running webpack ...

  WARNING   Webpack Encore requires version ^7.0.1 of sass-loader. Your version 8.0.0 is too new. The related feature *may* still work properly. If you have issues, try downgrading the library, or upgrading Encore.
```

My `webpack.config.js` :
```js

Encore
    .setPublicPath('/build')

    // some javascript entries
    // ...

    .enableSingleRuntimeChunk()

    .enableSassLoader((options) => {
        options.sourceMap = true;
        options.sassOptions = {
            outputStyle: 'compressed',
            sourceComments: !Encore.isProduction(),
        };
    }, {})

    .autoProvidejQuery()
    .configureBabel(function (babelConfig) {
        babelConfig.plugins.push('@babel/plugin-proposal-class-properties');
    })
    .enableReactPreset()
    .copyFiles({
        from: './assets/images',
        to: 'images/[path][name].[ext]',
    })
    .copyFiles({
        from: './assets/svg',
        to: 'svg/[path][name].[ext]',
    })
;

```

Could be integrated to PR #645

Kocal:
Close #654 

Breaking changes from [`sass-loader` changelog](https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#800-2019-08-29): 
- minimum required `webpack` version is 4.36.0: done in 4411c45, however [a test is failing with lowest dependencies](https://travis-ci.org/github/symfony/webpack-encore/jobs/676799821#L11322)
- minimum required `node.js` version is 8.9.0: not an issue since we don't support Node 8 anymore (#731)
- sass options has been moved in `sassOptions` option: ok

@Lyrkan do you have an idea oh what to do with the failing test? Use  `DISABLE_UNSTABLE_CHECKS` on it?

Thanks!

### DIFF
--- a/lib/loaders/sass.js
+++ b/lib/loaders/sass.js
@@ -42,8 +42,10 @@ module.exports = {
         const config = Object.assign({}, {
             // needed by the resolve-url-loader
             sourceMap: (true === webpackConfig.sassOptions.resolveUrlLoader) || webpackConfig.useSourceMaps,
-            // CSS minification is handled with mini-css-extract-plugin
-            outputStyle: 'expanded'
+            sassOptions: {
+                // CSS minification is handled with mini-css-extract-plugin
+                outputStyle: 'expanded'
+            }
         });
 
         sassLoaders.push({
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "style-loader": "^1.1.3",
     "terser-webpack-plugin": "^1.1.0",
     "tmp": "^0.0.33",
-    "webpack": "^4.20.0",
+    "webpack": "^4.36.0",
     "webpack-cli": "^3.0.0",
     "webpack-dev-server": "^3.1.14",
     "webpack-manifest-plugin": "^2.0.2",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "style-loader": "^1.1.3",
     "terser-webpack-plugin": "^1.1.0",
     "tmp": "^0.0.33",
-    "webpack": "^4.36.0",
+    "webpack": "^4.36.1",
     "webpack-cli": "^3.0.0",
     "webpack-dev-server": "^3.1.14",
     "webpack-manifest-plugin": "^2.0.2",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "preact": "^8.2.1",
     "preact-compat": "^3.17.0",
     "sass": "^1.17.0",
-    "sass-loader": "^7.0.1",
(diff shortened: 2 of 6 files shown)


Lyrkan:
> @Lyrkan do you have an idea oh what to do with the failing test? Use DISABLE_UNSTABLE_CHECKS on it?

Weird fail... since it is on the "Lowest versions" job you can either:

* find the dependency that causes it and bump the min version so it matches the one we have in `yarn.lock`
* use `DISABLE_UNSTABLE_CHECKS` to bypass that check entirely

In my opinion it would probably be better to avoid `DISABLE_UNSTABLE_CHECK` if we can (especially if it's just a matter of updating something).

weaverryan:
@Kocal about the failure, interestingly enough, on #746 (Vue3 support) I see the same failure. But not in "lowest", in that PR, I see it on the normal build. It seems like Webpack may have tweaked something internally that is causing our order to change in `entrypoints.json`, but I'm not sure....

Kocal:
Hum, that's weird...

Does it mean that I don't have to fix this issue there, and we can ignore CI failures to review and merge this PR?

Because ATM, the only reason I keep this PR as draft is because of the CI failure :sweat_smile:

weaverryan:
I found the issue. Just change the order in the test so that the test passes. In Webpack 4.22.0, a "deterministic ordering bug" was apparently fixed - https://github.com/webpack/webpack/releases/tag/v4.22.0 - I believe that caused Webpack to start outputting things in a different order. I don't think it's actually a problem - just something that's different before and after this version. Because this PR bumps Webpack higher than 4.22, it shouldn't be problem. Though... I'm puzzled as to why this wouldn't cause the tests to fail across all builds... since all builds now use higher than 4.22. So maybe I'm still missing a piece of the puzzle. I DO know that, on my Vue3 PR, I determined that Webpack 4.22.0 was the version that "toggled" the different `0.js` and `.js` ordering.

Kocal:
PR rebased and conflicts fixed.

stof:
@weaverryan as it was not deterministic before, it is possible that it would fail only randomly...

weaverryan:
Closing in favor of #758 - I got the weird test failure figured out over there :)

Issue #670, opened 2019-11-19
Babel not working for IE11


eberhapa:
Hi,

Do I have to do something special to make classes work in ie11 because there are still classes in my js build file and therefore my js isn't working in ie11. 

`class A extends d{constructor(...e) ........`

In the package.json I'm targeting:
`"browserslist": [
    "last 2 versions",
    "IE 11",
    "ios >= 8"
  ],` 

I'm using "@symfony/webpack-encore": "^0.28.0".

webpack.config.js:
```
Encore
  .enableSingleRuntimeChunk()

  // the project directory where all compiled assets will be stored
  .setOutputPath('build/')

  // the public path used by the web server to access the previous directory
  .setPublicPath('/wp-content/themes/enfold-child/build')

  .addEntry('main', [
    './src/js/main.js'
  ])
  .addStyleEntry('style', './src/sass/style.scss')
  .addStyleEntry('custom-editor-style', './src/sass/custom-editor-style.scss')

  // allow legacy applications to use $/jQuery as a global variable
  .autoProvidejQuery()

  // allow sass/scss files to be processed
  .enablePostCssLoader()
  .enableSassLoader()

  .copyFiles({
    from: './src/images',
    pattern: /\.(png|jpg|jpeg|svg)$/,
    to: 'images/[path][name].[ext]'
  })

  // enable source maps during development
  .enableSourceMaps(!Encore.isProduction())

  // empty the outputPath dir before each build
  .cleanupOutputBeforeBuild()
;
```

Thanks in advance,
Patrick

stof:
Can you share your webpack.config.js file too ?

stof:
also, do you have a `.babelrc` file in your project ?

eberhapa:
Hi stof,
No I don't have a .babelrc file because "Babel is automatically configured for all .js and .jsx files via the babel-loader with sensible defaults (e.g. with the @babel/preset-env and @babel/preset-react if requested)." - https://symfony.com/doc/current/frontend/encore/babel.html

Do I need one?

I updated my first comment with the config.

Thanks,
Patrick

Lyrkan:
Hey @eberhapa,

Two other things to check there:

[...]

Issue #558, opened 2019-04-09
Can't modify babel rule to exclude bootstrap in node_module


ger86:
Hi,

Last year I found a solution to allow Babel to process files in `node_modules` using something like this:

```
const webpackConfig = Encore.getWebpackConfig();
const babelLoader = webpackConfig.module.rules.find(rule => { 
  if (rule.use && rule.use[0]) {
    const firstUse = rule.use[0];
    return firstUse.loader === 'babel-loader';
  } 
  return false;
});
babelLoader.exclude = /node_modules\/(?!bootstrap\/).*/;
module.exports = webpackConfig;
```

But in the recent versions, when I run `yarn run encore production` I get the files from Bootstrap without being transpiled, which not happens in my regular files.

I am using the file babelrc to configure Babel:

```
{
    "presets": [
      [
        "@babel/preset-env",
        {
          "debug": true,
          "useBuiltIns": "usage",
          "corejs": 3,
          "targets": "> 0.25%, not dead"
        }
      ]
    ]
  }
```

The weird thing is that if I configure Babel through webpack.config.js:

 ```
   .configureBabel(function(babelConfig) {
    }, {
        include_node_modules: ['bootstrap']
    })
```

it works perfect. So I don't know the difference between the two solutions. Is there any implementation detail under this case that causes the first solution not valid (as I said, last year it works well)?

**Comment by ger86** [comment]
Something more... if a copy the babel config from file babel.js to .babelrc:

```
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {},
        "modules": false,
        "forceAllTransforms": true,
        "useBuiltIns": "entry"
      }
    ]
  ],
  "plugins": ["@babel/plugin-syntax-dynamic-import"]
}
```

It doesn't work neither. I have tried to replicate the logic of option `include_node_modules` in my webpack.config.js:


```
const webpackConfig = Encore.getWebpackConfig();
const babelLoader = webpackConfig.module.rules.find(rule => { 
  if (rule.use && rule.use[0]) {
    const firstUse = rule.use[0];
    return firstUse.loader === 'babel-loader';
  } 
  return false;
});
babelLoader.exclude = (filePath) => {
  if (!/(node_modules|bower_components)/.test(filePath)) {
      return false;
  }
  const whitelistedModules = ['bootstrap'].map(
      module => path.join('node_modules', module) + path.sep
  );

  for (const modulePath of whitelistedModules) {
[...]

PR #738, opened 2020-04-24
Fix babel config file detection

jdreesen:
Since v7 [Babel supports two types of config formats](https://babeljs.io/docs/en/config-files):
- Project-wide configuration
  - `babel.config` files, with the different extensions (`json`, `js`, `cjs`, `mjs`)
- File-relative configuration
  - `.babelrc` files, with the different extensions (none, `json`, `js`, `cjs`, `mjs`)
  - `package.json` files with a "babel" key

Encore, however, only checks for the existence of **file-relative** config files.

This is a problem if you want to compile packages from the `node_modules` folder (via `.configureBabel(null, {includeNodeModules: ['...']})`) because **project-wide** configuration files are needed for this.

This means that if you only have a `.babelrc.js` file, the package from the `node_modules` folder **will** be compiled by Babel, but **without** the settings defined in the `.babelrc.js` file.

If you now rename the `.babelrc.js` to `babel.config.js`, Encore fails to detect it and applies its default Babel config, which results in everything being complied without the settings defined in `babel.config.js`.

> Note: Configuring babel through Encore is not an option for me, because I want to make use of Babels [Config Function API](https://babeljs.io/docs/en/config-files#config-function-api) which is not available in Encore afaik.
And apart from that I prefer to have the configuration in dedicated files.

So my solution for now is to keep an empty `.babelrc.js` to make Encore think there is a Babel config file, and have an additional `babel.config.js` file with the real config. This results in both the application code and node modules being compiled with the desired settings.

---

But of course I dug a little deeper into Encore and Babel to fix the problem and the fix is using the `hasFilesystemConfig()` method instead of checking the `babelrc` property (the wording in the comment is a bit outdated, I already created a PR for that ;)):
https://github.com/babel/babel/blob/af669297efd775016725ee39318cdb391cf00a21/packages/babel-core/src/config/partial.js#L191-L200

I added some tests, too. I'm not sure if it's worth it to add tests for the different config file extensions, though.

I tried to improve the error messages as well. Tell me if you want another wording or different file examples and I'll change it.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -396,7 +396,7 @@ class WebpackConfig {
             }
 
             if (this.doesBabelRcFileExist()) {
-                logger.warning('The "callback" argument of configureBabel() will not be used because your app already provides an external Babel configuration (a ".babelrc" file, ".babelrc.js" file or "babel" key in "package.json"). Use null as a first argument to remove that warning.');
+                logger.warning('The "callback" argument of configureBabel() will not be used because your app already provides an external Babel configuration (e.g. a ".babelrc" or "babelrc.config.js" file or "babel" key in "package.json"). Use null as a first argument to remove that warning.');
             }
         }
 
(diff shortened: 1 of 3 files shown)


jdreesen:
This may be related to #558 and https://github.com/symfony/webpack-encore/issues/670#issuecomment-563043103.

weaverryan:
Thank you Jacob!

ankurk91:
We could use [cosmicconfig](https://www.npmjs.com/package/cosmiconfig) here, example:
 
```js
const cosmiconfig = require('cosmiconfig');

const userConfigExists = () => {
  return !!cosmiconfig('babel', {
    searchPlaces: [
      'package.json',
      '.babelrc',
      '.babelrc.js',
      'babel.config.js'
    ]
  }).searchSync()
};
```

This package can used to search for `postcss` configs as well.

Kocal:
That's a quite solution, but I think it's better to let the package find its configuration itself. This way we don't have to maintain a list of search places.

Issue #753, opened 2020-05-07
Tests frameworks are outdated


stof:
| Package | Used version | Latest version |
| --- | --- | --- |
| chai | 3.5.0 | 4.2.0 |
| chai-fs | 1.0.0 | 2.0.0 |
| mocha | 6.2.3 | 7.1.2 |
| sinon | 2.4.1 | 9.0.2 |

Issue #751, opened 2020-05-07
clean-webpack-plugin is totally outdated


stof:
Encore uses the 0.1.19 release, while the latest one is 3.0.0.

Issue #749, opened 2020-05-07
fork-ts-checker-webpack-plugin is totally outdated


stof:
Webpack uses fork-ts-checker-webpack-plugin 0.4.x for its tests (I haven't checked whether we enforce the version at runtime). The current version is 4.1.3

Issue #742, opened 2020-04-29
mini-css-extract-plugin is totally outdated


stof:
We currently force installing at most mini-css-extract-plugin@0.4.2, which is from 2018. Latest version is 0.9.0.

Lyrkan:
I already tried to update it in https://github.com/symfony/webpack-encore/pull/564 but for some reason it came with inconsistent hashes in filenames.

It seems to be fixed when used with Webpack 5 so it will probably be updated once https://github.com/symfony/webpack-encore/pull/645 is released.

stof:
this has been updated in the meantime

PR #754, opened 2020-05-07
Upgrade dependencies

stof:
This upgrades a bunch of dependencies to their latest version.

Things I intentionally kept out of this PR:

- test frameworks (mocha, sinon and chai), as upgrading them is more risky (when I don't touch them, I can rely on the existing testsuite to check the upgrade). These should be handled separately (especially given that they are massively outdated, requiring many major version bumps and so potentially many changes in the tests). See #753 
- webpack loaders and plugins, as such upgrades can create BC breaks for Encore itself (and so they deserve separate PRs to test them and make them visible). See #751, #749, #742 for some of them


weaverryan:
Thank you Stof - this is a very pragmatic, first step - I appreciate it!

PR #757, opened 2020-05-09
Update 0.29.0 changelog regarding CommonJS imports following the file-loader update

### LINKED ISSUE #743: dynamic image not found in react component
Hello, 
Here I try to import and use an image in a react component, but this one is never found : 

GET http://localhost:8000/[object%20Module] 404 (Not Found)

I try to access to defaultLogo.png which is in this tree:
https://prnt.sc/s92g2r

Here is my part of the code which is supposed to call and use it: 
`const pathToLogo = require('../../images/user_logo/${userData.logo_name}');

`

And here's my webpack encore config : 
`var Encore = require('@symfony/webpack-encore');
if (!Encore.isRuntimeEnvironmentConfigured()) {
    Encore.configureRuntimeEnvironment(process.env.NODE_ENV || 'dev');
}
Encore
    .setPublicPath('/build')
    .addEntry('app', './assets/js/app.jsx')
    .splitEntryChunks()
    .enableSingleRuntimeChunk()
    .cleanupOutputBeforeBuild()
    .enableBuildNotifications()
    .configureBabelPresetEnv((config) => {
        config.useBuiltIns = 'usage';
        config.corejs = 3;
    })
    .enableSassLoader()
    .enableReactPreset()
    .copyFiles({
        from: './assets/images/',
        to: 'images/[path][name].[hash:8].[ext]',
        pattern: /\.(png|jpg|jpeg|svg)$/
    })   
;
module.exports = Encore.getWebpackConfig(); `

Knowing that I'm working with "dev-server", but even "dev" or "build" doesn't work... 
Any ideas? Do I misunderstand the syntax? I confess I have some trouble with Webpack encore... 

Thanks a lot

Lyrkan:
As noted in https://github.com/symfony/webpack-encore/issues/743 the file-loader update we did in 0.29.0 has a breaking change when files are imported using the CommonJS syntax.

Since the loader is now exporting ES modules by default, importing them with `require(...)` retrieves the whole module and not only the default export.

We'll also have to update the Github changelog after merging this (closes #743).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@
 
    * `file-loader` upgraded from 1.1 to 6  - #731 thanks to @weaverryan.
      [CHANGELOG](https://github.com/webpack-contrib/file-loader/blob/master/CHANGELOG.md#600-2020-03-17)
-     Unless you're doing custom configuration, it's unlikely you're affected.
+     If you are importing images/fonts using the CommonJS syntax (`require('foo.png')`),
+     you will now need to explicitely retrieve the default export (`require('foo.png').default`)
+     in order to get the path of the file.
+     Imports done using the `import` keyword should not be affected.
 
    * `url-loader` upgraded from allowing version 1 or 2 to allowing version 4 - #731 thanks to @weaverryan.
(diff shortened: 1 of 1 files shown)


weaverryan:
Thank you @Lyrkan! I'll update the release right now

PR #746, opened 2020-05-05
Added Vue3 support

weaverryan:
Hi!

Vue3 is now beta. I think Encore should support Vue2 and Vue 3... at least for awhile. The changes needed to make Webpack work with v3 are not that many:

1) Make sure the user installs `vue/compiler-sfc` instead of `vue-template-compiler`
2) require `VueLoaderPlugin` in a slightly different way
3) Add a slightly different "alias" for vue

You can choose the version with a new `version` option to `enableVue()`. Otherwise, it will configure itself correctly if you have vue2 or vue3 installed, or it will (finally) recommend the vue2 install instructions. When v3 comes out, we would change the default recommendation to be v3.

But... getting the code (and test suite) to work with both vue2 and vue3 is quite tricky :).

PR #762, opened 2020-05-09
fix: preact compat

Grafikart:
With the realease of the version 10 of preact, preact compat was migrated inside preact library removing the need for preact-compat.

https://preactjs.com/guide/v10/switching-to-preact/

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -105,8 +105,8 @@ class ConfigGenerator {
         }
 
         if (this.webpackConfig.usePreact && this.webpackConfig.preactOptions.preactCompat) {
-            config.resolve.alias['react'] = 'preact-compat';
-            config.resolve.alias['react-dom'] = 'preact-compat';
+            config.resolve.alias['react'] = 'preact/compat';
+            config.resolve.alias['react-dom'] = 'preact/compat';
         }
 
         Object.assign(config.resolve.alias, this.webpackConfig.aliases);
--- a/test/config-generator.js
+++ b/test/config-generator.js
@@ -493,7 +493,7 @@ describe('The config-generator function', () => {
                 'foo': 'bar',
                 'vue$': 'new-vue$',
                 'react-dom': 'new-react-dom',
-                'react': 'preact-compat' // Keeps predefined aliases that are not overwritten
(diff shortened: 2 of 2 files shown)


weaverryan:
Thanks for this @Grafikart! One test is failing, with an easy failure.

But I'm a little concerned: this would mean now that Encore *only* works with preact 10, correct? If so, then we need to update the `devDependency` for Preact to version 10 - the tests are currently using version 8. And that makes me wonder something else: if you've updated the code for preact 10, but the tests are using version 8, why aren't more significant tests failing? It makes me wonder if the preact functional test - which builds this directory https://github.com/symfony/webpack-encore/tree/master/fixtures/preact - is not "good enough". In theory, that test should be failing right now because of your patch but because the tests are still using Preact 8 right?

Cheers!

Lyrkan:
> if you've updated the code for preact 10, but the tests are using version 8, why aren't more significant tests failing?

I don't think we've functional tests using `preact-compat` (which is a compatibility layer to use "real" React libraries from your preact app), the one you linked directly uses `preact`. I guess we could add some but it could be a bit tricky to know they are working properly since we also have `react`/`react-dom` in our dev dependencies.

I agree that we should bump the version in our dev dependencies though: if we don't people will be asked to install v8 and then get the wrong aliases for `preact-compat` if they enable that feature.

Grafikart:
> this would mean now that Encore only works with preact 10, correct? 

You are correct. I want input on how you want to handle this situation. We could parse the user package.json (hard way) or we could add a new option within enablePreactPreset.

```
enablePreactPreset({preactCompat: 10})
```

In my opinion the build should target the currentVersion. If someone chooses to work with an older version he can setup the aliases manually.

weaverryan:
I would say that we should either:

A) Only allow version 10 (I have no idea if 8 & 9 have LTS or something like that - I'm not familiar with preact)

B) Allow 8 or 10 and check their installed version to choose which alias to add. We do something similar for Vue 3 in #746. In that, we also have an option to force v2 or v3, but I did that mostly because v2 will be around for an LTS for a long time and because Vue is such a widely-used library.

Btw, my preference would be to only support 10. But if, for some reason, many people will continue to use 8 for a long time, then we could allow both.

Rodrigo001-de:

PR #763, opened 2020-05-10
Removing vue2 alias to use the full build

### LINKED ISSUE #674: RFC: Defaulting to vue.runtime.esm.js instead of vue.esm.js
Someone on Slack had a CSP issue with Vue.js caused by the fact we alias `vue$` to the full Vue.js build (`vue.esm.js`) that compiles templates on the client instead of using the runtime-only build (`vue.runtime.esm.js`).

As explained in Vue.js documentation (https://vuejs.org/v2/guide/installation.html#CSP-environments):

> Some environments, such as Google Chrome Apps, enforce Content Security Policy (CSP), which prohibits the use of new Function() for evaluating expressions. The full build depends on this feature to compile templates, so is unusable in these environments.
>
> On the other hand, the runtime-only build is fully CSP-compliant. When using the runtime-only build with Webpack + vue-loader or Browserify + vueify, your templates will be precompiled into render functions which work perfectly in CSP environments.

They also recommend using the runtime-only build:

> Since the runtime-only builds are roughly 30% lighter-weight than their full-build counterparts, you should use it whenever you can. If you still wish to use the full build instead, you need to configure an alias in your bundler:

Based on those two sections I can't see any reason not to switch to `vue.runtime.esm.js`, but maybe I'm missing something?

weaverryan:
Fixes #674

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## 0.30.0
 
+ * [BC BREAK] The Vue "build" was changed from `vue.esm.js` (full build) to
+   `vue.runtime.esm.js` (runtime build). With the runtime build, there are
+   two things that you cannot do:
+   
+   A) You cannot pass a string to `template`:
+   
+   ```js
+   new Vue({
+     template: '<div>{{ hi }}</div>'
+   })
+   ```
+   
+   B) You cannot mount to a DOM element and use its HTML as your template:
+
+   ```js
+    new Vue({
+      el: '#app', // where <div id="app"> contains your Vue template
+    });
+   ```
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 
 ## 0.30.0
 
+ * [BC BREAK] The Vue "build" was changed from `vue.esm.js` (full build) to
+   `vue.runtime.esm.js` (runtime build). With the runtime build, there are
+   two things that you cannot do:
+   
+   A) You cannot pass a string to `template`:
+   
+   ```js
```
stof, line 11: this code snippet looks incomplete, as it is empty.

PR #765, opened 2020-05-10
Fixing babel.config.js filename in message

weaverryan:
Introduced in #738 - @jdreesen I think this was just a typo? Can you confirm that my PR is right?

Thanks!


jdreesen:
Oh, yes, that's a typo!

PR #769, opened 2020-05-12
Reverting change to only package vue runtime loader

weaverryan:
Using the runtime build by default makes it harder for beginners and it's hard to know the fix. Specifically, if you follow the "intro docs" to Vue, you will likely try this:

```js
var app = new Vue({
  el: '#app',
  data: {
    message: 'Hello Vue!'
  }
})
```

which will not work (and the error will be tough for beginners to spot). As a compromise, on a production build, we print a recommendation so that people can discover the option to use the smaller build.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1151,7 +1151,7 @@ class Encore {
      *     // set optional Encore-specific options, for instance:
      *
      *     // set to false to *only* include the smaller "runtime"
-     *     // build, which can't compile templates at runtime, but it
+     *     // build, which can't compile templates at runtime, but is
      *     // CSP compliant/
      *     // set explicitly to true to silence the recommendation
      *     runtimeCompilerBuild: false
--- a/index.js
+++ b/index.js
@@ -1150,6 +1150,12 @@ class Encore {
      * Encore.enableVueLoader(() => {}, {
      *     // set optional Encore-specific options, for instance:
      *
+     *     // set to false to *only* include the smaller "runtime"
+     *     // build, which can't compile templates at runtime, but it
+     *     // CSP compliant/
+     *     // set explicitly to true to silence the recommendation
+     *     runtimeCompilerBuild: false
+     *
      *     // use version 2 or 3 to force your Vue version
      *     // otherwise, Encore will detect it automatically
      *     version: 2
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -137,6 +137,7 @@ class WebpackConfig {
         this.vueOptions = {
             useJsx: false,
             version: null,
+            runtimeCompilerBuild: null
         };
         this.eslintOptions = {
             lintVue: false,
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -43,6 +43,7 @@ const PluginPriorities = require('./plugins/plugin-priorities');
 const applyOptionsCallback = require('./utils/apply-options-callback');
 const sharedEntryTmpName = require('./utils/sharedEntryTmpName');
 const copyEntryTmpName = require('./utils/copyEntryTmpName');
+const getVueVersion = require('./utils/get-vue-version');
 const tmp = require('tmp');
 const fs = require('fs');
 const path = require('path');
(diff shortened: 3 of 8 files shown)


Lyrkan:
Displaying the recommendation during prod builds only may be too late I think. If someone starts working on an app only using `encore dev`/`encore dev-server` they could use things that won't be compatible with the runtime-only build and then need to do a lot of refactoring in order to be able to use it once they learn about the alias. Also, having a CSP compliant build by default could prevent other issues that are hard to diagnose for beginners.

Couldn't we detect the error thrown by Vue and make it easier to understand instead (with an example of how to add the alias)?

Lyrkan:
Also, the failing test looks like the one I had to fix recently when working on Webpack 5 : https://github.com/symfony/webpack-encore/pull/645/files#r422717443

weaverryan:
> Couldn't we detect the error thrown by Vue and make it easier to understand instead (with an example of how to add the alias)?

I don't think so? You see the error at runtime. For example, in Vue 3 (this is the source .ts, but you get the general idea of how this looks in the final dist file), you get the error simply because the `compile()` function is set to one that throws an error: https://github.com/vuejs/vue-next/blob/e954ba21f04f0ef848c687233fcb849d75e4153f/packages/vue/src/runtime.ts#L8

I don't think you can detect that at build time, well, at least not easily.

We could move the "recommendation" to show on *any* build... I just thought it might be annoying... though being annoying is not the worst thing here. CSP is great, but it's not our main use-case as afaik, it's only a problem in environments like Chrome apps.

stof:
> We could move the "recommendation" to show on _any_ build... I just thought it might be annoying... though being annoying is not the worst thing here. CSP is great, but it's not our main use-case as afaik, it's only a problem in environments like Chrome apps.

Or any site where you deploy CSP.

lib/config-generator.js
```diff
@@ -100,6 +102,24 @@ class ConfigGenerator {
...
+                logger.recommendation(`If you do not need to compile Vue templates at runtime, pass ${chalk.green('{ runtimeCompilerBuild: false }')} as the 3rd argument to ${chalk.green('enableVueLoader()')} for a smaller build.`);
+            }
+
+            const vueVersion = getVueVersion(this.webpackConfig);
+            switch (vueVersion) {
+                case 2:
+                    config.resolve.alias['vue$'] = 'vue/dist/vue.esm.js';
+                    break;
+                case 3:
+                    config.resolve.alias['vue'] = 'vue/dist/vue.esm-bundler.js';
```
stof, line 116: I think we also need to alias only `vue$` here, to avoid breaking cases where they import `vue/dist/*` explicitly.

lib/WebpackConfig.js
```diff
@@ -137,6 +137,7 @@ class WebpackConfig {
         this.vueOptions = {
             useJsx: false,
             version: null,
+            runtimeCompilerBuild: true
```
stof, line 140: What about making this option `boolean|undefined`, and showing the warning only if undefined ? That would allow people to configure explicitly that they need the full build, getting rid of the warning.

weaverryan:
Ok, compromise time! With help from Stof:

1) This PR still defaults to the "full" build. This helps beginners.
2) We now print a recommendation on ALL builds, AND that recommendation includes some info about CSP. So if CSP is not something you've thought about, we've just pinged you about it ;)

> To create a smaller (and CSP-compliant) build, see https://symfony.com/doc/current/frontend/encore/vuejs.html#runtime-compiler-build

That points to a docs PR I just made: https://github.com/symfony/symfony-docs/pull/13664/files

3) You can silence the warning (if you really DO want a runtime build and don't like seeing our nice blue message on each build) by explicitly setting `runtimeCompilerBuild: true`.

lib/utils/manifest-key-prefix-helper.js
```diff
@@ -20,6 +20,10 @@ const WebpackConfig = require('../WebpackConfig'); //eslint-disable-line no-unus
 module.exports = function(webpackConfig) {
     let manifestPrefix = webpackConfig.manifestKeyPrefix;
     if (null === manifestPrefix) {
+        if (null === webpackConfig.publicPath) {
+            throw new Error('publicPath is not set on WebpackConfig');
```
stof, line 24: This change looks unrelated

weaverryan: It totally is :). When I didn't fully set up a new test case, I hit a bad "calling replace on null" error (the next line). So, I covered that edge case with a proper error. It won't affect real users - if they somehow were hitting this before, I've just made their error more clear :)

index.js
```diff
@@ -1150,6 +1150,12 @@ class Encore {
      * Encore.enableVueLoader(() => {}, {
      *     // set optional Encore-specific options, for instance:
      *
+     *     // set to false to *only* include the smaller "runtime"
+     *     // build, which can't compile templates at runtime, but it
```
Lyrkan, line 1154:
```suggestion
     *     // build, which can't compile templates at runtime, but is
```

PR #770, opened 2020-05-13
build(deps): remove explicit dependency to "terser-webpack-plugin"

Kocal:
Webpack already [depends of `terser-webpack-plugin`](https://github.com/webpack/webpack/blob/v4.36.0/package.json#L28), so I don't think this is revelant to require here too.

WDYT?

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "resolve-url-loader": "^3.0.1",
     "semver": "^7.3.2",
     "style-loader": "^1.1.3",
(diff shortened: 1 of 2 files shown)


stof:
https://github.com/symfony/webpack-encore/blob/f9c3fc6fb047741449238d85e46622b24876f284/lib/plugins/terser.js is requiring that package directly. So we need that dependency. npm does not guarantee you that you are able to access transitive dependencies.

Issue #785, opened 2020-06-05
Support babel option includeNodeModules with pnpm


leonexcc:
The option includeNodeModules of bable doesn't work with [pnpm](https://pnpm.js.org/), since filePath will contain the real path after resolving the link (for example .../node_modules/.pnpm/plyr@3.6.2/node_modules/plyr/...)

After replacing
```
const whitelistedModules = options[optionKey].map(
     module => path.join('node_modules', module) + path.sep
);
```

with 

```
const whitelistedModules = options[optionKey].map(
    module => path.dirname(require.resolve(module + '/package.json'))
);
```

it works as expected.

Also the mapping is currently executed for each file again. This could probably done once outside of the callback.

stof:
Doing `require.resolve` when processing the config will have several issues:

- it restricts that to packages that can be resolved from the root, while the current approach works for any transitive deps too
- it won't work for packages using the new `exports` config (available in Node 13+), as `package.json` might not be an allowed export of the package

also, I don't think the path `.../node_modules/.pnpm/plyr@3.6.2/node_modules/plyr/...` is an issue. This path has `node_modules/plyr/` in it.

leonexcc:
You are right. My issue was that my modules in includeNodeModules were not transpiled and i thought this was the reason. But i used a .babelrc instead of a babel.config.json and the .babelrc is ignored for node modules. Maybe this could be documented somewhere next to the includeNodeModules option.

stof:
Documenting that makes sense.
Or maybe even detecting this case automatically in Encore. Is this possible @weaverryan @Lyrkan ?

Lyrkan:
> Or maybe even detecting this case automatically in Encore. Is this possible @weaverryan @Lyrkan ?

I think we could... we currently check for both files (and `package.json` config) here:

https://github.com/symfony/webpack-encore/blob/a7d76ee0661ddeac35e241e1f48962dbf6123906/lib/config/parse-runtime.js#L88-L104

But before #738 we only had a check for the existence of `.babelrc` files using:

```js
runtimeConfig.babelRcFileExists = (typeof partialConfig.babelrc === 'string');
```

Maybe we could have both of them instead (that will require to refactor some other files to use `babelConfigExists` instead of `babelRcFileExists`):

```js
runtimeConfig.babelConfigExists = partialConfig.hasFilesystemConfig();
[...]

PR #788, opened 2020-06-10
Update the JSDoc of configureBabel to warn about .babelrc and node_modules

stof:
Babel supports 2 different types of config files:

- `.babelrc` stops at package boundaries
- `babel.config.*` applies to the whole project, including `node_modules`

Once node_modules is not excluded anymore in the babel-loader configuration, this distinction matters.

Refs #785

### DIFF
--- a/index.js
+++ b/index.js
@@ -931,14 +931,19 @@ class Encore {
      *              determines which files and folders should not be
      *              processed by Babel (https://webpack.js.org/configuration/module/#condition).
      *              Can be used even if you have an external Babel configuration
-     *              (a .babelrc file for instance)
+     *              (a babel.config.json file for instance)
+     *              Warning: .babelrc config files don't apply to node_modules. Use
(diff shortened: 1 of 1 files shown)


jdreesen:
`babel.config.*` files are supported in encore since `v0.30.0`: #738

stof:
@jdreesen sure. But this doc update is to warn about the case of #785: if you use a `.babelrc` file and you decide to apply babel on some node_modules packages, it won't do what you might expect, because it will run babel without your project config (and so probably with an empty config)

jdreesen:
Yes, you're absolutely right!

I just wanted to add that for the reference. Sorry for the confusion.

Rodrigo001-de:

Lyrkan:
Thanks @stof

We should still implement what I described in https://github.com/symfony/webpack-encore/issues/785#issuecomment-641945537, I'll open another issue about it.

PR #801, opened 2020-07-10
Upgrade assets-webpack-plugin to ^5.0.3

### LINKED ISSUE #806: assets-webpack-plugin must be upgraded to v5
We currently use the v3 of assets-webpack-plugin, which depends on a vulnerable version of lodash. This is fixed in their v5 release (there is no actual v4).
the BC break between v3 and v5 is dropping support for Node 8, which does not impact Encore as our minimum version is already Node 10.13.

Khartir:
This upgrades `assets-webpack-plugin` to 5.0.3.

This is a new major version of the plugin, however according to https://github.com/ztoben/assets-webpack-plugin/blob/HEAD/changelog.md that is only due to the plugin now requiring node 10.

Since encore already requires node 10 this is not a problem.

The older version of `assets-webpack-plugin` depends on an unsafe version of `lodash` that causes `npm audit` to fail (https://npmjs.com/advisories/1523)

Fixes #806

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@babel/core": "^7.7.0",
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
     "@babel/preset-env": "^7.4.0",
-    "assets-webpack-plugin": "^3.9.7",
+    "assets-webpack-plugin": "^5.0.3",
     "babel-loader": "^8.0.0",
     "chalk": "^4.0.0",
     "clean-webpack-plugin": "^3.0.0",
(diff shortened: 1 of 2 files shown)


stof:
@Lyrkan @weaverryan  what is the plan to get this merged and released ?

Lyrkan:
Let's merge it (thanks @Khartir).

I won't be able to create a new release though.

RentecTravis:
I would love for this to be released on npm so I have one less anxiety-producing security vulnerability reported every time I do an npm install/update/audit.

Kocal:
You can run `yarn add https://github.com/symfony/webpack-encore.git --dev` if you really need the latest version.

PierrickMartos:
Is it possible to have a new release?

Kocal:
See https://github.com/symfony/webpack-encore/pull/801#issuecomment-670359375

PierrickMartos:
Yes but sorry it's not a suitable fix, had a dev dep to force the latest version

stof:
Well, the upgrade is not released yet. AFAIK, only @weaverryan can perform the npm release right now. And I don't know whether he is in vacations right now.

Khartir:
For everybody else following this issue: there has been a release containing this fix

PR #804, opened 2020-07-14
Bump less-loader to version 6

### LINKED ISSUE #803: less-loader 6.2.0 is too new
We're running Webpack Encore with the latest version of the less-loader package in our project. This results in a warning:

```
Webpack Encore requires version ^4.1.0 of less-loader. Your version 6.2.0 is too new. The related feature *may* still work properly. If you have issues, try downgrading the library, or upgrading Encore.
```

The suggested version is more than two years old. Also, the current version seems to work fine, at least in our project.

I'd like to get rid of the warning without downgrading to an ancient release. Are there any known issues with less-loader > 4 that need to be resolved?

derrabus:
Fixes #803.

Please see the linked issue for details & discussion.

Suggested ChangeLog entry:

```markdown
 * [DEPENDENCY UPGRADE] `less-loader` was upgraded from version 4 to 6.
   If you passed any less-specific options to the loader (like `plugins` or
   `rewriteUrls`), you will need to wrap them into an object and set them as
   `lessOptions`. Please have a look at the project's
   [CHANGELOG.md](https://github.com/webpack-contrib/less-loader/blob/v6.2.0/CHANGELOG.md)
   for details and more breaking changes between versions 4 and 6.
```


Lyrkan:
Thanks @derrabus

VincentLanglet:
[7.0](https://github.com/webpack-contrib/less-loader/releases/tag/v7.0.0) is released.

There is not a lot of BC-break. Can we bump the version the same way ?

derrabus:
@VincentLanglet Makes sense to me. Feel free to create a PR. 😃

Issue #808, opened 2020-07-20
Images in >0.30 in manifest have hashed path


darrylhein:
When I update to >=0.30 (works fine in 0.29.1) the keys for images in the manifest are the hashed versions instead of the original paths. (I suspect fonts will be the same.) Prior to 0.30 I get:

    "images/icons.svg": "/build/images/icons.677433a0.svg",

After 0.30 I get:

   "images/icons.677433a0.svg": "/build/images/icons.677433a0.svg",

...which of course doesn't work :(

I'm import the image in my JS files.

I'm wondering if this is related to #771 though I'm still have the issue with 0.30.2.

I can reproduce the issue when updating webpack-encore (and sass-loader) on https://github.com/xmmedia/starter_wordpress

Let me know what other information you need.

stof:
@darrylhein can you share the diff of your yarn.lock file for this upgrade (https://www.npmjs.com/package/yarn-lock-diff can probably be used to generate a summary of the upgrade) ?
This issue might be related to a upgrade of one of a dependency rather than a change in Encore itself.

Lyrkan:
As discussed on Slack with @4c0n, it seems to be related to https://github.com/danethurber/webpack-manifest-plugin/issues/209

We updated the `file-loader` in [v0.29.0](https://github.com/symfony/webpack-encore/releases/tag/v0.29.0) which came with its `esModule` option set to `true` by default (so it should be broken in v0.29 as well).

A quick workaround would be to change that value to `false` using `configureUrlLoader()` (since we don't have a `configureFileLoader()`):

```js
Encore.configureUrlLoader({
  images: {
    limit: 0, // Avoids files from being inlined
    esModule: false
  }
});
```

4c0n:
@Lyrkan Thanks for looking into it so fast :smile: I'm just using copyFiles()  in the webpack.config.js file to get the right result for now.

darrylhein:
@stof Here's the "diff". I only changed the encore version in the package.json and then did `yarn install` in the hope to not update anything else unrelated.

![yarn.lock diff](https://user-images.githubusercontent.com/376976/88315739-ecba0080-ccd3-11ea-8e16-445d8962647c.png)

I've tried @Lyrkan's suggestion, but didn't fix things and it appears to have removed the image files from the manifest all together. I could just copy the files, but I'd like them to be run through the svgo-loader.

Thanks for the help so far.

Lyrkan:
[...]

PR #810, opened 2020-07-23
Bug/incorrect manifest keys are produced

4c0n:
When importing a svg through an entry, the incorrect manifest key is generated.

This PR contains a failing test case to illustrate the problem.

It seems very related to https://github.com/symfony/webpack-encore/issues/808, however I also seem to have this problem when downgrading to a lower version.


stof:
When downgrading, are you keeping the same versions of dependencies or downgrading as well ? Maybe the upgrade mentioned in #808 was involving more packages than Encore

4c0n:
@stof I also had to downgrade the `sass-loader` package, but that's it. @Lyrkan mentioned it probably came to be in version 0.29.0
https://github.com/symfony/webpack-encore/issues/808#issuecomment-663083975

PR #817, opened 2020-08-03
Fix the documentation about autoProvidejQuery

stof:
The implementation actually defines 3 provided variables, not only 2.


Lyrkan:
Thanks @stof

Issue #816, opened 2020-08-03
[RFC] Ship typescript types for the public APIs


stof:
Several IDEs (including PHPStorm/WebStorm and VS Code) are using the typescript type definitions to enhance their autocompletion for Javascript as well.
While investigating #815, I made a small experiment: I generated a index.d.ts file in webpack-encore to have type definitions for the Encore public API (which is defined and documented in index.js). And this solved the type support for PHPStorm 2020.2.

Given that the type declaration can be fully generated from the existing JSDoc, I suggest that we bundle it in the package.
For the record, here is the command I ran (at the root of the package): `tsc --declaration --allowJs --emitDeclarationOnly index.js`.

To make this production-ready, a few changes would be necessary:
- use a proper semver constraint for the typescript devDependency rather than `>=2.9` to be able to use `tsc` in our publication process (no impact on projects using encore as that constraint is not enforced for enabling typescript features)
- improve JSDoc types for arrays by specifying the value type (Typescript does not like unspecified Array types).
- (optionally) improve the JSDoc for callbacks to document their actual signature rather than using `{function}` (but this will be hard to go fully there: configuration callbacks are dealing with options of loaders, but we cannot refer to the type definitions shipped in these loaders for any optional dependency)
- (optionally) improve the JSDoc for webpack-encore options to document the available options (for arguments where we mix loader options and Encore options, that's not possible for the same reason than above, but we have many methods with a separate argument for Encore options)
- add a `prepublishOnly` script doing that type generation before publishing the releases to npm

All the `improve JSDoc` steps could actually be done even without shipping typescript type declarations, as they would still proper better type definitions. And actually, I have part of them done locally as part of my experiment.

What do you think @Lyrkan @weaverryan ?

My own vote would be in favor of including them, given that we don't need to maintain the `.d.ts` file manually.

stof:
Note that even if the IDE decides to take the `.d.ts` file as the autocompletion source, all the explanations written in the JSDoc are preserved, as `tsc` will copy the JSDoc to the type declaration.

stof:
@Lyrkan @weaverryan any opinion on this ?

chapterjason:
> Note that even if the IDE decides to take the `.d.ts` file as the autocompletion source, all the explanations written in the JSDoc are preserved, as `tsc` will copy the JSDoc to the type declaration.

[...]

PR #818, opened 2020-08-03
Add more precise types in the JSDoc

stof:
These are the type improvements coming from my investigation in #816 (and then pushed until the end rather than stopping at a few methods for the experiment)

### DIFF
--- a/lib/utils/package-up.js
+++ b/lib/utils/package-up.js
@@ -16,13 +16,18 @@ const { fileURLToPath } = require('url');
 /**
  * Inlined version of the package "package-up" (ESM only).
  *
- * @param {string} cwd The directory to start searching from.
+ * @param {object} options
+ * @param {string} options.cwd The directory to start searching from.
  * @returns {string|undefined} The path to the nearest package.json file or undefined if not found.
  */
 module.exports = function({ cwd }) {
     return findUpSync('package.json', { cwd, type: 'file' });
 };
 
+/**
+ * @param {string|URL} urlOrPath
+ * @returns {string}
+ */
 function toPath(urlOrPath) {
     return urlOrPath instanceof URL ? fileURLToPath(urlOrPath) : urlOrPath;
 }
@@ -32,7 +37,7 @@ function toPath(urlOrPath) {
  *
  * @param {string} name The name of the file to find
  * @param {object} options
- * @param {string} options.cwd The directory to start searching from.
+ * @param {string=} options.cwd The directory to start searching from.
  * @returns {string|undefined} The path to the file found or undefined if not found.
  */
 function findUpSync(name, { cwd = process.cwd() } = {}) {
--- a/index.js
+++ b/index.js
@@ -1655,7 +1655,7 @@ class Encore {
      * )
      * ```
      *
-     * Be aware than using this method will also reset the current
+     * Be aware that using this method will also reset the current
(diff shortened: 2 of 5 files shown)


stof:
The failing tests seem unrelated to my PR. I'm changing only comments

stof:
hmm, looks like the jsdoc parser of eslint does not like the object types defining properties. And given that the `valid-jsdoc` rule is deprecated and the corresponding parser is unmaintained, it is unlikely to be fixed.

@Lyrkan @weaverryan should we disable that rule ?

stof:
Or maybe we should migrate to https://github.com/gajus/eslint-plugin-jsdoc

Lyrkan (commented):
Sorry for not looking at this sooner, those are some great changes :)

> @Lyrkan @weaverryan should we disable that rule ?
> Or maybe we should migrate to gajus/eslint-plugin-jsdoc

We could try `gajus/eslint-plugin-jsdoc` but I wouldn't mind disabling `valid-jsdoc` entirely either tbh.

weaverryan:
My preference would be to remove our `valid-jsdoc` and try https://github.com/gajus/eslint-plugin-jsdoc - it looks high quality, and i do kinda like our very-complete (and validated with eslint) jsdoc.

stof:
@Lyrkan @weaverryan I made the migration to eslint-plugin-jsdoc (and I fixed all the extra things it reports, which were not detected by `valid-jsdoc`)

stof:
The failure for the job `Lowest versions of the dependencies` is unrelated to my change.  that's also broken in the main branch.

Kocal:
Hey, what about using TypeScript to check types? 
No transpilation, just checking types on .js files with `allowJs`/`checkJs` options, and `tsc --noEmit`?

stof:
@Kocal using typescript to check types on our own codebase could be a next step after that. But it would require having types defined properly for all our code, while this PR was focusing on what impacts our public API.
The goal was to only downstream projects to use typescript to check types on their `webpack.config.js` file.

I should definitely revive that PR as it would be a nice addition.

Kocal:
@stof you're right, that makes sense. 

> I should definitely revive that PR as it would be a nice addition.

Totally! :)

lib/utils/package-up.js
```diff
@@ -16,13 +16,17 @@ const { fileURLToPath } = require('url');
 /**
  * Inlined version of the package "package-up" (ESM only).
  *
- * @param {string} cwd The directory to start searching from.
+ * @param {{cwd: string}} options The directory to start searching from.
```
Kocal, line 19: Shouldn't we keep the previous syntax? Here the description "The directory (...)" refers to the `options` object, and not to `cwd` right?

Or we can remove the description, since it's super internal, I don't really have a strong opinion on this.

Kocal (approved):
Wow 🤩 

That's will definitely increase DX for end-users and for us, just a minor comment tho

lib/utils/package-up.js
stof: The previous syntax describes a `cwd` argument which does not exist (there is a `cwd` property in the first argument instead).

stof: I updated the JSDoc to use the alternative syntax allowing to put a description on properties (this was already used later in the file)

Issue #830, opened 2020-08-28
copyFiles() should not require to use addEntry() or addStyleEntry()


pszalko:
I have the use case where I use two separate webpack configs.
First one is for whole application, second for vendor assets only.

In the second config I use only `copyFiles()` and I have the following error:
```
 Error: No entries found! You must call addEntry() or addStyleEntry() at least once - otherwise... there is nothing to webpack!
```

Whole webpack.config.js is as below:

```
const webpackConfig = Encore.getWebpackConfig(); // build config from settings given earlier
Encore.reset();

// vendor assets:
Encore
    // directory where compiled assets will be stored
    .setOutputPath('public/build/vendor/')
    // public path used by the web server to access the output path
    .setPublicPath('/build/vendor')
    // copy vendor files
    .copyFiles([
    {
        from: './assets/vendor/keen_v2.0.0/theme/demo1/src/media',
        to: 'keen_v2/media/[path][name].[hash:8].[ext]',
    },
    {
        from: './assets/vendor/swagger',
        to: 'swagger/[path][name].[hash:8].[ext]'
    }
])

let vendorConfig = Encore.getWebpackConfig();
module.exports = [webpackConfig, vendorConfig];
```

I think the internal validator should allow to use `copyFiles()` only, without any `addEntry()` nor `addStyleEntry()`, because `copyFiles()` creates one entry in webpack config internally.

Current workaround is to add dummy entry and then delete it:

```
Encore
    // ...
    .addEntry('hack', 'will-be-removed')
    // copy vendor files
    .copyFiles(/* ... */);

let vendorConfig = Encore.getWebpackConfig();
delete vendorConfig.entry.hack;

// ...
```

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #831, opened 2020-08-28
Validator should allow copyFiles() without other entries.

pszalko:
Fix for #830

### DIFF
--- a/test/config/validator.js
+++ b/test/config/validator.js
@@ -34,6 +34,19 @@ describe('The validator function', () => {
         }).to.throw('No entries found!');
     });
 
+    it('should accept use with copyFiles() only', () => {
+        const config = createConfig();
+        config.setOutputPath('/tmp');
+        config.setPublicPath('/tmp');
+        config.copyFiles({ from: './' });
+
+        expect(() => {
+            validator(config);
+        }).not.throw();
+
+        expect(Object.keys(config.copyFilesConfigs).length).to.equal(1);
+    });
+
     it('throws an error if there is no output path', () => {
         const config = createConfig();
         config.publicPath = '/';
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -152,6 +152,14 @@ describe('Public API', () => {
             expect(returnedValue).to.equal(api);
         });
 
+        it('should be allowed to use without addEntry() or addStyleEntry()', () => {
+            api.setOutputPath('/tmp')
+                .setPublicPath('/tmp')
+                .copyFiles({ from: './' });
+            const webpackConfig = api.getWebpackConfig();
+            expect(Object.keys(webpackConfig.entry).length).to.equal(1);
+        });
```
Lyrkan, line 161: This test should probably be in https://github.com/symfony/webpack-encore/blob/master/test/config/validator.js since that's where we have the other "no entries found" checks.

Lyrkan (changes requested):
Hi @pszalko,

Thanks for the PR, I think it would be a nice thing to have :)

I only have a small comment regarding the test you added.

test/index.js
pszalko: Hi @Lyrkan, thanks for suggestion. I moved the code to `test/config/validator.js` file. I hope the PR is good to merge now :)

test/config/validator.js
```diff
@@ -34,6 +34,19 @@ describe('The validator function', () => {
...
 
+    it('should accept use with copyFiles() only', () => {
+        const config = createConfig();
+        config.setOutputPath('/tmp');
+        config.setPublicPath('/tmp');
+        config.copyFiles({ from: './' });
+
+        !expect(() => {
+            validator(config);
+        }).not.throw();
```
Lyrkan, line 45: I'm not sure that exclamation mark should be there:

```suggestion
        expect(() => { validator(config); }).to.not.throw();
```

Lyrkan (commented):
One last really minor comment before approving :)

test/config/validator.js
pszalko: I don't know why the exclamation mark is there at all :) Thanks for spotting it.

weaverryan:
Thank you @pszalko!

PR #848, opened 2020-10-21
Update resolve-url-loader to fix prototype pollution

Khartir:
See https://github.com/bholloway/resolve-url-loader/issues/170 and https://npmjs.com/advisories/1573


Khartir:
I'm unsure if the errors in the test are caused by my changes. I get the same errors when running the test with master.

weaverryan:
Thank you @Khartir!

PR #854, opened 2020-11-18
Updates postcss loader to v4

railto:
postcss-loader is a full major version behind, this updates it to v4


msheakoski:
Hopefully this gets merged soon! It supports PostCSS v8 while remaining backwards compatible with v7. It is holding back an upgrade to Tailwind CSS v2 because it only works with PostCSS v8.

weaverryan:
Thank you Mark!

PR #859, opened 2020-12-03
Implement Stimulus bridge configurator

tgalopin:
Implement Stimulus bridge configurator for Symfony UX


fabpot:
Thank you @tgalopin.

lib/plugins/stimulus-bridge.js
```diff
@@ -0,0 +1,27 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+const WebpackConfig = require('../WebpackConfig'); //eslint-disable-line no-unused-vars
+const createPlugin = require('@symfony/stimulus-bridge/webpack-helper');
```
stof, line 13: this breaks, because you import this file even when the plugin is not enabled but `@symfony/stimulus-bridge` is not a dependency of `webpack-encore`.

And this also misses the dependency checks

tgalopin: Thanks, I'm having a look!

Kocal: I've opened https://github.com/symfony/webpack-encore/pull/863

PR #862, opened 2020-12-03
chore: sync yarn.lock after #859 updates

Kocal:
The PR #859 has modified the `package.json` but forgot to update the `yarn.lock`, making all CI builds failing (because we use `yarn --frozen-lockfile`).

cc @fabpot @weaverryan :pray:


weaverryan:
Thank you @Kocal!

PR #866, opened 2020-12-03
Add missing call to done() in the Stimulus functional test

Lyrkan:
One of the new tests is currently missing a call to `done()` which breaks the output of the next tests:

image


weaverryan:
Good catch @Lyrkan!

PR #870, opened 2020-12-03
Prefer sass over node-sass

weaverryan:
This will change the recommendation to `sass` instead of `node-sass` which is deprecated and harder to deal with. I haven't had any issues or differences working with `sass` yet vs `node-sass`.

I also upgraded the lock file in general, it makes it easier to see what packages are legitimately out of date and doesn't affect end users.

### DIFF
--- a/lib/features.js
+++ b/lib/features.js
@@ -21,7 +21,7 @@ const features = {
         packages: [
             { name: 'sass-loader', enforce_version: true },
             // allow node-sass or sass to be installed
-            [{ name: 'node-sass' }, { name: 'sass' }]
+            [{ name: 'sass' }, { name: 'node-sass' }]
         ],
         description: 'load Sass files'
     },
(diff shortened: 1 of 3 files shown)


Kocal:
> I haven't had any issues or differences working with sass yet vs node-sass.

I faced one issue which is related to vue-loader, if you use [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors) and still want to style a deeper element, you can use `>>>` (CSS) or `/deep/` (node-sass) or `::v-deep` (dart sass). `/deep` won't work on Dart Sass because of syntax error.

stof:
yeah, that's because the `/deep/` proposal has been rejected in the CSS spec and so has been removed from the sass syntax as well in dart-sass (ruby-sass was warning about it being deprecated and libsass had a todo about implementing the deprecation IIRC)

Issue #880, opened 2020-12-31
Webpack 5 Support & 1.0 meta issue


weaverryan:
Hi!

This issue tracks things that we want to get done before releasing a 1.0 of Encore with Webpack 5 support:

* [X] A) Initial Webpack 5 support #645 
* [X] B) Using old watchOptions regexp & removing Node 13 support #878
* [X] C) Collect changes (mostly from #645) and add them to CHANGELOG #883
* [X] D) Integrate [Asset Modules](https://webpack.js.org/guides/asset-modules/) for images, fonts, etc #883 
* [~] ~~E) potentially use WebpackManifestPlugin also for entrypoints.json~~ Will consider later
* [x] F) Check into our `stats` config - should we "hijack" this less and allow for more normal output?
* [X] G) add method to enable new persistent caching - #884
* [x] H) Decide on a backwards-compatibility (BC) promise related to Webpack dependencies (e.g. if we bump `@babel/core` from v7 to v8, does that require a new major version of Encore? OR what if we drop support for `sass-loader` v9 (which is a library version that the user controls) - does that require a new major version of Encore?
* [X] I) Ideally, make all the tests that rely in hashes more forgiving - don't skip them entirely, but match for "fuzzines" - e.g. that we expect a file matching a pattern (faux example: `main.[alpha length 8].js`). - #883 
* [~] ~~J) Make a decision about `friendly-errors-webpack-plugin` which appears abandoned https://github.com/geowarin/friendly-errors-webpack-plugin/issues/123~~ we will keep for now, it works fine, but the warning when installing Encore is annoying :/ 
* [x] L) Merge new Stimulus-bridge support #888
* [x] K) #889 - bumped `@babel/preset-env`

versedi:
>  H) Decide on a backwards-compatibility (BC) promise related to Webpack dependencies (e.g. if we bump @babel/core from v7 to v8, does that require a new major version of Encore? OR what if we drop support for sass-loader v9 (which is a library version that the user controls) - does that require a new major version of Encore?

IMHO it shouldn't rely on the major/minor update of the dependency's version, but instead on the fact if the dependency update breaks anything inside encore - eg. configuration syntax changes or behaviour of something changes.

weaverryan:
About dependencies:

[...]

PR #883, opened 2021-01-09
Updating images and fonts to use Webpack 5 Asset modules

weaverryan:
This is one more big piece for proper Webpack 5 support - see #880

This also updates the CHANGELOG both for this PR and for the original in #645 

And, this adds a new assertion method that helps check directory contents in functional.js in a way that is more resilient to hash changes.

For the most part... this just worked! See the CHANGELOG. Because this will release on 1.0, I have decided to remove a few methods instead of deprecating them ).

Cheers!


lib/webpack/copy-files-loader.js
```diff
@@ -10,7 +10,7 @@
 'use strict';
 
 const LoaderDependency = require('webpack/lib/dependencies/LoaderDependency');
-const fileLoader = require('file-loader');
+const fileLoader = require('file-loader'); // eslint-disable-line node/no-unpublished-require
```
Lyrkan, line 13: I think that should be done in the `loader` function.

test/helpers/assert.js
```diff
@@ -168,6 +192,57 @@ class Assert {
 
         expect(JSON.stringify(actualData, null, 2)).to.equal(JSON.stringify(expectedData, null, 2));
     }
+
+    /**
+     * Verifies that the directory contains the array of files.
+     *
+     * The expectedFiles can contain a [hash:8] syntax in case
+     * the file is versioned - e.g. main.[hash:8].js, which would
+     * match a real file like main.abcd1234.js.
```
Lyrkan, line 201: 👍

lib/WebpackConfig.js
```diff
@@ -97,8 +97,18 @@ class WebpackConfig {
         this.useSourceMaps = false;
         this.cleanupOutput = false;
         this.extractCss = true;
-        this.useImagesLoader = true;
-        this.useFontsLoader = true;
+        this.imageRuleOptions = {
+            type: 'asset',
+            maxSize: null,
```
Lyrkan, line 102: I'm wondering if we shouldn't override the default value for maxSize.

Based on the documentation everything under 8kb will be inlined... which isn't necessarily a bad thing but it will probably confuse some people/break things (for instance if someone used to import images in order to trigger a copy).

weaverryan: > which isn't necessarily a bad thing but it will probably confuse some people/break things

I think this will also cause some confusion. On the other hand, if you have:

```js
import image from './images.foo.png';

img.src = image;
```

I think this will work regardless of whether it was inlined or copied. Unless we defaulted this to 0 (or used `asset/resource` instead of `assets` by default), this will surprise people. 

So, I'm conflicted: on the one hand, the inlining of 8kb and lower is probably better for performance... but will cause more confusion.

weaverryan: I'm going to go the "more conservative" route and use `asset/resource` by default

PR #887, opened 2021-01-20
Bump less 4 and less loader 7

VincentLanglet:
I tried to do like your PR https://github.com/symfony/webpack-encore/pull/804 @derrabus


VincentLanglet:
The lower test is already failing on main.

weaverryan:
Thanks Vincent!

PR #888, opened 2021-01-22
updating stimulus-bridge plugin to work with proposed new loader

weaverryan:
Partner to https://github.com/symfony/stimulus-bridge/pull/14

* [x] Wait for partner PR to be merged


lib/features.js
```diff
@@ -152,7 +152,7 @@ const features = {
         method: 'enableStimulusBridge()',
         packages: [
             { name: '@symfony/stimulus-bridge' },
```
stof, line 154: shouldn't we enforce the version here too (even more than for `stimulus`) ?

lib/features.js
```diff
@@ -152,7 +152,7 @@ const features = {
...
-            { name: 'stimulus' }
+            { name: 'stimulus', enforce_version: true }
```
stof, line 155: the commit message talks about bumping stimulus version, but that's not what you bumped

weaverryan: yup - this was a mistake - it should be on the other one - thanks for catching!

PR #900, opened 2021-01-28
Adding ifProd() and ifDev() methods

weaverryan:
Hi!

This occurred to me when I was trying to add `Encore.enableBuildCache()` to the official recipe. I was thinking that it might be best to enable it *except* when doing a production build. Currently, that's only possible by breaking the chaining:

```js
Encore
    // .alotOfCalls()
;

if (!Encore.isProduction()) {
    Encore.enableBuildCache({
        config: [__filename],
    })
}
```

We have always handled making this easier on a method-by-method basis, which is still probably nice in many cases (e.g. .`.enableVersioning(Encore.isProduction())` reads really nice. But these new methods make it possible to keep chaining with anything else we don't support in this way:

```js
Encore
    // .alotOfCalls()

    .ifDev((Encore) => Encore.enableBuildCache({ config: [__filename] }))
;
```

Cheers!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1597,6 +1597,46 @@ class Encore {
     }
 
     /**
+     * Use to conditionally configure or enable features only in production.
+     *
+     * ```
+     * Encore
+     *     // ...
+     *     .ifProduction((Encore) => Encore.enableVersioning()))
+     * ```
+     *
+     * @param {function} callback
+     * @return {Encore}
+     */
+    ifProduction(callback) {
+        if (this.isProduction()) {
+            callback(this);
+        }
+
+        return this;
+    }
+
+    /**
+     * Use to conditionally configure or enable features for dev only.
+     *
+     * ```
+     * Encore
+     *     // ...
+     *     .ifDev((Encore) => Encore.enableBuildCache({ config: [__filename] }))
+     * ```
+     *
+     * @param {function} callback
+     * @return {Encore}
+     */
+    ifDev(callback) {
+        if (!this.isProduction()) {
+            callback(this);
+        }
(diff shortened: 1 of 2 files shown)


Kocal:
What about `isDevServeur`?

weaverryan:
> What about ifDevServer, since we have isDevServer?

I thought about that... and this does introduce some inconsistency. The new method `isDev()` is true if you are `isDev()` OR `isDevServer()`... which I think is just much more useful. We could name the new method `ifNotProduction()`... but that might be less clear.

Kocal:
Hum alright, I see your point!

Kocal:
Also, I've been thinking about this for a long time but, WDYT about a more generic method like `when`, that would be usable like this:

```js
Encore
  .when(true, Encore => Encore.foo())
  .when(() => true, Encore.bar())
  .when(Encore => Encore.isDev(), Encore.baz())
```

A usecase I have on many projects at work is to add `BundleAnalyzerPlugin` plugin when argument `--analyze` is passed:
```js
// Before

// ...
if (process.argv.includes('--analyze')) {
  Encore.addPlugin(new BundleAnalyzerPlugin({ analyzerHost: '0.0.0.0' }));
}

// After
Encore
  // ...
  .when(process.argv.includes('--analyze'), Encore => Encore.addPlugin(new BundleAnalyzerPlugin({ analyzerHost: '0.0.0.0' }));

```

Lyrkan:
> Also, I've been thinking about this for a long time but, WDYT about a more generic method like when, that would be usable like this:

I really like that idea.

Issue #893, opened 2021-01-27
Update to Encore 1.0 dev-server failed


kl3sk:
Hello,

I try updating my project with the new `Encore 1.0`, but after trying to update my config I ran on a fail build

Here is the error when I start `npm run dev-server`
```
[webpack-cli] Invalid configuration object. Object has been initialized using a configuration object that does not match the API schema.
 - configuration has an unknown property 'disableHostCheck'. These properties are valid:
   object { bonjour?, client?, compress?, dev?, firewall?, headers?, historyApiFallback?, host?, hot?, http2?, https?, injectClient?, injectHot?, liveReload?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, openPage?, overlay?, port?, proxy?, public?, setupExitSignals?, static?, stdin?, transportMode?, useLocalIp? }
```

Here is the npm log:

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/bin/npm', 'run', 'dev-server' ]
2 info using npm@6.14.8
3 info using node@v14.13.1
4 verbose run-script [ 'predev-server', 'dev-server', 'postdev-server' ]
5 info lifecycle @~predev-server: @
6 info lifecycle @~dev-server: @
7 verbose lifecycle @~dev-server: unsafe-perm in lifecycle true
8 verbose lifecycle @~dev-server: PATH: /usr/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin:/home/klesk/<project>/node_modules/.bin:/home/klesk/.npm-global/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
9 verbose lifecycle @~dev-server: CWD: /home/klesk/<project>
10 silly lifecycle @~dev-server: Args: [ '-c', 'encore dev-server' ]
11 silly lifecycle @~dev-server: Returned: code: 2  signal: null
12 info lifecycle @~dev-server: Failed to exec dev-server script
13 verbose stack Error: @ dev-server: `encore dev-server`
13 verbose stack Exit status 2
13 verbose stack     at EventEmitter.<anonymous> (/usr/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:332:16)
13 verbose stack     at EventEmitter.emit (events.js:314:20)
[...]

PR #901, opened 2021-01-29
Working around missing manifest.json bug

weaverryan:
This fixes the 2nd part of #893 - https://github.com/symfony/webpack-encore/issues/893#issuecomment-769488763

On Webpack 5, there is some miscommunication between the CleanWebpackPlugin and the WebpackManifestPlugin when using `dev-server`.

The `WebpackManifestPlugin` *does* emit the physical `manifest.json` file, but CleanWebpackPlugin then deletes it. CleanWebpackPlugin should just remove files *before* we start, and not delete anything that's emitted. I'm not sure which plugin has the bug. But, the work around is quite clean.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1421,11 +1421,11 @@ class Encore {
      *     // common values: asset, asset/resource, asset/inline
      *     // Using "asset" will allow smaller images to be "inlined"
      *     // instead of copied.
-     *     // javascript/auto caan be used to disable asset images (see next example)
+     *     // javascript/auto can be used to disable asset images (see next example)
      *     type: 'asset/resource',
      *
      *     // applicable when for "type: asset": files smaller than this
-     *     // size will be "inlined" into CSS, larer files will be extracted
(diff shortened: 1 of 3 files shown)


mradhi:
I think we need to get rid of the **"https" argument** as specified on [webpack/webpack-dev-server /CHANGELOG.md#400-beta0-2020-11-27](https://github.com/symfony/webpack-encore/blob/main/lib/config/parse-runtime.js#L44), the `https` option is now an object, so I think we have a conflict, especially when we try to configure the https using the `webpack.config.js` file as below

```
Encore.configureDevServerOptions(options => {
        options.https = {
            pfx: '/path/to/.symfony/certs/default.p12'
        }
})
```

Then we should not run the command `yarn encore dev-server --https `as it will override the defined `https` option prepared in the config file, and if we **remove** that option from the command, the output file inside `public/build/entrypoints.json` file will contains invalid URLs **without the https protocol**

PR #912, opened 2021-02-03
Allow node 10.19

nicolas-grekas:
This was bumped in https://github.com/symfony/webpack-encore/pull/645
Ubuntu LTS [ships node 10.19](https://packages.ubuntu.com/focal/nodejs) by default.
If we can, let's save ppl (aka me) from installing yet another ppa.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/symfony/webpack-encore/issues"
   },
   "engines": {
-    "node": "^10.22.1 || ^12.0.0 || >=14.0.0"
+    "node": "^10.19.0 || ^12.0.0 || >=14.0.0"
   },
   "homepage": "https://github.com/symfony/webpack-encore",
   "dependencies": {


jdreesen:
Note that Node 10 EOLs at 2021-04-30: https://nodejs.org/en/about/releases/

nicolas-grekas:
That's the EOL date from node maintainers. There are other ppl that take over maintenance after them, eg from Ubuntu LTS or REHL.

tgalopin:
An indirect dependency of Encore (webpack-manifest-plugin) doesn't support 10.19 anymore, I guess that's the reason behind the bump.

Kocal:
I don't think using nodejs (or php, python, ...) binaries from Ubuntu (or other any OS) official repositories is a good idea. 
Using a versions managing tool, like [nvm](https://github.com/nvm-sh/nvm), would allow you to use any Node.js versions you want (latest 10, latest 12, etc..).

Also, I'm not a big fan for lowering the version range "just for you", because of your OS constraints. IMO, this is not a problem Encore should have to deal with.

And as @tgalopin said, [webpack-manifest-plugin](https://github.com/shellscape/webpack-manifest-plugin/blob/master/package.json#L12] requires at least Node 10.22.1.

nicolas-grekas:
Thanks for spotting the dep that triggered the bump. Here is a PR to fix it too: https://github.com/shellscape/webpack-manifest-plugin/pull/247

We don't have to wait for this other PR to merge this one.

Bumping when there is no technical reason to do so is doing DRM and forcing ppl (not just me this time) to spend more time doing extra work. This is definitely not friendly towards any community.

weaverryan:
Thanks Nicolas!

nicolas-grekas:
Unfortunately https://github.com/shellscape/webpack-manifest-plugin/pull/247 has been rejected :'(

Kocal:
Should we revert this PR?

PR #917, opened 2021-02-05
Re-working dev-server and https detection

### LINKED ISSUE #903: When using dev-server https option manifests and paths in CSS files are still http:// instead of https://
I have just spent a few hours trying to make dev-server serve files via SSL with .pfx specified (as per Symfony recommendation and personal preference), but I found no way around it. It looks to me like a bug in `webpack-dev-server`, but am not sure where to look for.

* Before (encore < 1.0.0)

I had set up in `packages.json`:
```
"scripts": {
    "dev-server": "encore dev-server --compress --https --pfx=\"$HOME/.symfony/certs/default.p12\" --disable-host-check",
}
```

* After (encore 1.0.x)

Most options were easily migrated to `configureDevServerOptions`, but:
- when I set `options.https: { pfx: process.env.HOME + '/.symfony/certs/default.p12' }` and no `--https` in command line, dev-server serves files with correct certificate. But `manifest.json`, `entrypoints.json` and paths in CSS contains paths with `http://`, thus development app served via https cannot load any asset.
- when I set `options.https: { pfx: process.env.HOME + '/.symfony/certs/default.p12' }` or no `options.https` at all, but pass `--https` in command line, dev-server serves files with self signing certificate (ignoring `pfx` configuration), but creates correct manifest.json, entrypoints.json and paths in CSS.

Second options could be a short term solution for me (accepting new certificate once in a while is not that bad), but it would be nice to have working `pfx` option with manifest plugin. I see no way to pass pfx in command-line, which could probably solve the issue.

weaverryan:
Hi!

Fixes #903

If you pass `--https` at the command line, that overrides your `devServer.https` config. In webpack-dev-server v3, that wasn't a problem. But in v4, the certificate config was moved under `devServer.https`, which can now be an object. This meant that if you passed `--https` at the command line (which Encore previously required) but also set `devServer.https = { ... }`, that would be "re-set" back to `devServer.https = true` and your config would be lost.

The fix is to not require the `--https` flag and look at it or the user's config to determine if the dev-server is running in https.

Cheers!

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -57,6 +57,20 @@ class ConfigGenerator {
     }
 
     getWebpackConfig() {
+        const devServerConfig = this.webpackConfig.useDevServer() ? this.buildDevServerConfig() : null;
+        /*
+         * An unfortunate situation where we need to configure the final runtime
+         * config later in the process. The problem is that devServer https can
+         * be activated with either a --https flag or by setting the devServer.https
+         * config to an object or true. So, only at this moment can we determine
+         * if https has been activated by either method.
+         */
+        if (this.webpackConfig.useDevServer() && (devServerConfig.https || this.webpackConfig.runtimeConfig.devServerHttps)) {
+            this.webpackConfig.runtimeConfig.devServerFinalIsHttps = true;
+        } else {
+            this.webpackConfig.runtimeConfig.devServerFinalIsHttps = false;
+        }
+
         const config = {
             context: this.webpackConfig.getContext(),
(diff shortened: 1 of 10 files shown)


lib/config-generator.js
```diff
@@ -86,7 +100,7 @@ class ConfigGenerator {
         }
 
         if (this.webpackConfig.useDevServer()) {
```
Kocal, line 102: What about this, since we alraedy check `this.webpackConfig.useDevServer()` before?

```suggestion
        if (null !== devServerConfig) {
```

PR #940, opened 2021-02-24
Clarify missing css file error message

Gadgetdude:
As a new user I came across a very confusing error message which took me way too long to debug. I was trying to set up the new webpack-encore configuration to use the `twig-loader`. I followed all the instructions, but I kept getting the following error:
```
./node_modules/twig/twig.js contains a reference to the file path.
This file can not be found, please check it for typos or update it if the file got moved.
```

Since "file path" is a very commonly used phrased, I did not realise that it meant that the file (or dependency) `path` was missing. Searching online also didn't help me out.

I'm not sure what the best solution is (or if it's an accepted edge case), but having it highlighted could save some new users a lot of time and frustration when starting out. I decided on simple quotes since that works regardless of colours are enabled in the terminal.

### DIFF
--- a/lib/friendly-errors/formatters/missing-css-file.js
+++ b/lib/friendly-errors/formatters/missing-css-file.js
@@ -22,7 +22,7 @@ function formatErrors(errors) {
         chalk.red('Module build failed: Module not found:')
     );
     for (let error of errors) {
-        messages.push(`${error.file} contains a reference to the file ${error.ref}.`);
+        messages.push(`"${error.file}" contains a reference to the file "${error.ref}".`);
         messages.push('This file can not be found, please check it for typos or update it if the file got moved.');
         messages.push('');
     }
--- a/lib/friendly-errors/formatters/missing-loader.js
+++ b/lib/friendly-errors/formatters/missing-loader.js
@@ -50,7 +50,7 @@ function formatErrors(errors) {
         if (error.loaderName) {
             messages.push(`${chalk.bgGreen.black('', 'FIX', '')} To ${loaderFeatures.getFeatureDescription(error.loaderName)}:`);
         } else {
-            messages.push(`${chalk.bgGreen.black('', 'FIX', '')} To load ${error.file}:`);
+            messages.push(`${chalk.bgGreen.black('', 'FIX', '')} To load "${error.file}":`);
         }
 
         let index = 0;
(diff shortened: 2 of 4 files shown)


lib/friendly-errors/formatters/missing-css-file.js
```diff
@@ -22,7 +22,7 @@ function formatErrors(errors) {
         chalk.red('Module build failed: Module not found:')
     );
     for (let error of errors) {
-        messages.push(`${error.file} contains a reference to the file ${error.ref}.`);
+        messages.push(`${error.file} contains a reference to the file "${error.ref}".`);
```
stof, line 25: should we wrap `error.file` in quotes too for consistency ?

Gadgetdude: Sounds good. I will do a quick check in the other files in the `lib/friendly-errors` directory as well.

Gadgetdude: After some research I noticed that in `lib/friendly-errors/transformers/missing-css-file.js` (among other files) a single quote is used for referencing to paths and modules. I also found another path reference in another formatter and updated it.

stof: I'd rather normalize to quotes rather than apostrophes, as that's about an English message, not a code snippet.

"single quote" vs "double quote" is a language abuse coming from programming languages describing things in a weird way. In Unicode, `"` is named `quotation mark` and `'` is named `apostrophe`. Using `'` as a quotation mark is misusing it.

Also, this would be consistent with the formatting used in error messages in Symfony itself.

Gadgetdude:
@stof: Thanks for the review. I've updated the code accordingly. Could you check the changes? I've left the discussion open for you to comment on or to resolve/close.

lib/friendly-errors/formatters/missing-css-file.js
Gadgetdude: You are absolutely correct, these are English sentences and should be treated as such. I've refactored all to use quotation marks as you suggested.

I skipped the file I based it off before (`lib/friendly-errors/transformers/missing-css-file.js`). If I understand the file correctly it transforms an error message that originates outside of this repository, so updating it to use quotation marks would break things.

weaverryan:
Wow, this is wonderful! Even I had to stare for a minute at the error message you received and check the code... I was also confused! Thank you so much for, at the very least, saving others the pain in the future. Cheers Rolf!

PR #960, opened 2021-04-06
fix: don't override process variable in DefinePlugin configuration

drazik:
Following the warning on https://webpack.js.org/plugins/define-plugin/:

> When defining values for process prefer
> `'process.env.NODE_ENV': JSON.stringify('production')` over
> `process: { env: { NODE_ENV: JSON.stringify('production') } }`.
> Using the latter will overwrite the process object which can break
> compatibility with some modules that expect other values on the
> process object to be defined.

This fixed a conflict warning I had when using [`dotenv-webpack` plugin](https://github.com/mrsteele/dotenv-webpack):

```
DefinePlugin
Conflicting values for 'process.env'
```

### DIFF
--- a/lib/plugins/define.js
+++ b/lib/plugins/define.js
@@ -21,9 +21,9 @@ const applyOptionsCallback = require('../utils/apply-options-callback');
  */
 module.exports = function(plugins, webpackConfig) {
     const definePluginOptions = {
-        'process.env': {
-            NODE_ENV: webpackConfig.isProduction() ? '"production"' : '"development"'
-        }
+        'process.env.NODE_ENV': webpackConfig.isProduction()
+            ? '"production"'
+            : '"development"',
     };
 
     plugins.push({
--- a/test/config-generator.js
+++ b/test/config-generator.js
@@ -225,7 +225,7 @@ describe('The config-generator function', () => {
             const actualConfig = configGenerator(config);
 
             const definePlugin = findPlugin(webpack.DefinePlugin, actualConfig.plugins);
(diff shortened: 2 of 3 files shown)


weaverryan (commented):
Hey @drazik!

This seems to make sense - thank you. However, there are some tests that are failing on Travis. with an error about not being able to read NODE_ENV from undefined. Can you take a look?

Thanks!

Kocal:
Should be this considered as a breaking change? Or at least being documented in a changelog?

At work we use the following code to inject `SENTRY_DSN` and `SENTRY_ENVIRONMENT` env vars in the bundle (from a `.env` file or from `process.env`):
```js
Encore
  .configureDefinePlugin((options) => {
    const dotenvEnv = dotenv.config();
    if (dotenvEnv.error) {
      throw dotenvEnv.error;
    }

    const env = { ...dotenvEnv.parsed, ...process.env };

    ['SENTRY_DSN', 'SENTRY_ENVIRONMENT'].forEach((envVar) => {
      options['process.env'][envVar] = JSON.stringify(env[envVar]);
    });
  })
```

And I guess it won't works until we change `options['process.env'][envVar]` to `options['process.env.' + envVar]` right?
WDYT?

drazik:
@weaverryan I fixed most of the test. One of them is still failing but I don't understand if it's related to my changes or not.

@Kocal you are totally right, this is a breaking change for people that customize the DefinePlugin like you do. And you are also right about the changes it requires in your case.

weaverryan:
Due to the breaking change (good catch), I'd love to stuff this into a new major release instead of a minor.

weaverryan:
Thank you @drazik!

PR #963, opened 2021-04-15
feat: add Encore.when()

Kocal:
Replace #900 by adding a more generic and flexible API than `Encore.isDev()`/`Encore.isProd()`, see https://github.com/symfony/webpack-encore/pull/900#issuecomment-769380187.

`Encore.when()` takes two parameters:
- `condition`: can be a callback (where the current instance of Encore is passed as first parameter) or a boolean. If this results to `true`, the parameter `callback` is called
- `callback`: executed when parameter `condition` results to `true`, it takes the current instance of Encore as first parameter

WDYT? Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1597,6 +1597,29 @@ class Encore {
     }
 
     /**
+     * Use to conditionally configure or enable features only in when the first parameter results to "true".
+     *
+     * ```
+     * Encore
+     *     // passing a callback
+     *     .when((Encore) => Encore.isProduction(), (Encore) => Encore.enableVersioning())
+     *     // passing a boolean
+     *     .when(process.argv.includes('--analyze'), (Encore) => Encore.addPlugin(new BundleAnalyzerPlugin()))
+     * ```
+     *
+     * @param {(function(Encore): boolean) | boolean} condition
+     * @param {function(Encore): void} callback
+     * @return {Encore}
+     */
+    when(condition, callback) {
+        if (typeof condition === 'function' && condition(this) || typeof condition === 'boolean' && condition) {
+            callback(this);
+        }
+
+        return this;
+    }
+
+    /**
      * Use this at the bottom of your webpack.config.js file:
      *
      * ```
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1596,6 +1596,29 @@ class Encore {
...
+     * ```
+     * Encore
+     *     // passing a callback
+     *     .when((Encore) => Encore.isProduction(), (Encore) => Encore.enableVersioning())
+     *     // passing a boolean
+     *     .when(process.argv.includes('--analyze'), (Encore) => Encore.addPlugin(new BundleAnalyzerPlugin()))
+     * ```
+     *
+     * @param {(function(Encore): boolean) | boolean} condition
+     * @param {function} callback
```
stof, line 1611: should be `function(Encore): void`

stof (changes requested):

index.js
```diff
@@ -1596,6 +1596,29 @@ class Encore {
         return webpackConfig.isDevServer();
     }
 
+    /**
+     * Use to conditionally configure or enable features only in when the first parameter results to "true".
```
weaverryan, line 1600:
```suggestion
     * Use to conditionally configure or enable features only when the first parameter results to "true".
```

index.js
```diff
@@ -1596,6 +1596,29 @@ class Encore {
...
+     *
+     * ```
+     * Encore
+     *     // passing a callback
+     *     .when((Encore) => Encore.isProduction(), (Encore) => Encore.enableVersioning())
+     *     // passing a boolean
+     *     .when(process.argv.includes('--analyze'), (Encore) => Encore.addPlugin(new BundleAnalyzerPlugin()))
```
weaverryan, line 1607: Cool examples!

weaverryan (approved):

weaverryan:
This is very cool - thank you @Kocal!

index.js
weaverryan: Updated at sha: d0741010efcd47212bc632bb8ffa2021e5c724a5

Issue #936, opened 2021-02-19
Manifest key problem with "build cache" + copyFiles with non-fresh cache


weaverryan:
Hi!

I just hit a bug when you combine `Encore.enableBuildCache` with `Encore.copyFiles()`. When you build fresh (no cache, easy to replicate by removing `node_modules/.cache/`), the `manifest.json` contains entries like this:

```json
"build/images/chart.png": "/build/images/chart.db1f878a.png",
```

But if you build again when there is already cache available (easy to replicate by running `yarn watch` and modifying any file), then you get:

```json
"build/images/chart.db1f878a.png": "/build/images/chart.db1f878a.png",
```

Notice they key has the version. 

Btw, my guess is that this is a bug with webpack-manifest-plugin & the build cache, not actually with `copyFiles()`. It's just that `copyFiles()` is the only place in the system that uses `file-loader`... which is probably related.

Cheers!

weaverryan:
I was correct about this being a bug with cache + webpack-manifest-plugin, and I can see why - the `file-loader` is never called... and that library relies on a hook in `file-loader` to collect the filenames.

Interestingly, https://github.com/webdeveric/webpack-assets-manifest does not have this problem. They DO have a different bug related to cache, however - https://github.com/webdeveric/webpack-assets-manifest/issues/131

Anyways, I'll look there for inspiration to see what hooks they're using for file-loader.

kl3sk:
FYI I ran into the problem, disabling `enableBuildCache` resole my problems.

This is how it was configured:

```js
    .enableBuildCache({
        // object of "buildDependencies"
        // https://webpack.js.org/configuration/other-options/#cachebuilddependencies
        // __filename means that changes to webpack.config.js should invalidate the cache
        config: [__filename],
    })

```

bobvandevijver:
@weaverryan I just ran into this issue as well... Did you by any chance find a way to solve this (or another place to file an issue)?

weaverryan:
I haven’t had time to look into it yet - I only know that the https://github.com/webdeveric/webpack-assets-manifest library somehow doesn’t have this problem. But diving into these complex kind to understand what they’re doing differently is... quite tricky.

bobvandevijver:
[...]

PR #979, opened 2021-05-18
#936: Fix manifest key problem when using copy files

bobvandevijver:
So, the solution for #936 was not complex at all, but creating a failing test was. 

I needed to update the test scripts in the `package.json`, as webpack simply doesn't write it's cache until the mocha  process ends. I did not find an alternative way to trigger the cache files to be written. 

So, I separated the persistent cache tests into a dedicated directory/file, which is ignored during the normal test run. I added commands to run the persistent cache test both with and without previous cache (using the `CLEAR_PERSISTENT_CACHE` environment variable), and updated the original test command to run the normal test once, and the persistent cache tests twice (so the second run is with a valid cache). 

I also needed to update the temp directory generation for the persistent cache, as it need to be deterministic over multiple cache runs. This is done with an extra parameter when creating the webpack configuration. This also makes sure that the webpack persistent cache is not shared over multiple tests (as long as you use unique keys obviously.

Just let me know if something needs to be done different!

### DIFF
--- a/lib/webpack-manifest-plugin/hooks.js
+++ b/lib/webpack-manifest-plugin/hooks.js
@@ -129,12 +129,16 @@ const normalModuleLoaderHook = ({ moduleAssets }, loaderContext, module) => {
   const { emitFile } = loaderContext;
 
   // eslint-disable-next-line no-param-reassign
-  loaderContext.emitFile = (file, content, sourceMap) => {
+  loaderContext.emitFile = (file, content, sourceMap, assetInfo) => {
+    const info = Object.assign({}, assetInfo);
+
     if (module.userRequest && !moduleAssets[file]) {
-      Object.assign(moduleAssets, { [file]: join(dirname(file), basename(module.userRequest)) });
+      info.sourceFilename = join(dirname(file), basename(module.userRequest));
+
+      Object.assign(moduleAssets, { [file]: info.sourceFilename });
     }
 
-    return emitFile.call(module, file, content, sourceMap);
+    return emitFile.call(module, file, content, sourceMap, info);
   };
 };
 
--- a/package.json
+++ b/package.json
@@ -4,9 +4,11 @@
   "description": "Webpack Encore is a simpler way to integrate Webpack into your application",
   "main": "index.js",
   "scripts": {
-    "test": "mocha --reporter spec test --recursive",
+    "test": "yarn run test:main && yarn run test:persistent-cache",
+    "test:main": "mocha --reporter spec test --recursive --ignore test/persistent-cache/*",
+    "test:persistent-cache": "node run-persistent-tests",
     "lint": "eslint lib test index.js .eslintrc.js",
-    "travis:lint": "npm run lint"
+    "travis:lint": "yarn run lint"
   },
   "bin": {
     "encore": "bin/encore.js"
--- a/run-persistent-tests.js
+++ b/run-persistent-tests.js
@@ -1,0 +1,18 @@
+/*
+ * This file is part of the Symfony Webpack Encore package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+const { execSync } = require('child_process');
+const { emptyTmpDir } = require('./test/helpers/setup');
+
+emptyTmpDir();
+for (let i = 0; i < 2; i++) {
+    execSync('mocha --reporter spec test/persistent-cache --recursive', { stdio: 'inherit' });
+}
--- a/test/functional.js
+++ b/test/functional.js
@@ -753,23 +753,6 @@ describe('Functional tests using webpack', function() {
             });
         });
 
-        it('Persistent caching does not cause problems', (done) => {
-            const config = createWebpackConfig('www/build', 'dev');
-            config.setPublicPath('/build');
-            config.addEntry('main', './js/code_splitting');
-            config.enableBuildCache({ config: [__filename] });
-
-            testSetup.runWebpack(config, (webpackAssert) => {
-                // sanity check
-                webpackAssert.assertManifestPath(
-                    'build/main.js',
-                    '/build/main.js'
-                );
-
-                done();
-            });
-        });
-
         describe('addCacheGroup()', () => {
             it('addCacheGroup() to extract a vendor into its own chunk', (done) => {
                 const config = createWebpackConfig('www/build', 'dev');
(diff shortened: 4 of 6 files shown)


lib/webpack-manifest-plugin/hooks.js
```diff
@@ -128,13 +128,18 @@ const emitHook = function emit(
 const normalModuleLoaderHook = ({ moduleAssets }, loaderContext, module) => {
   const { emitFile } = loaderContext;
 
+  // assetInfo parameter was added in Webpack 4.40.0
```
stof, line 131: We don't need this comment

package.json
```diff
@@ -4,7 +4,11 @@
   "description": "Webpack Encore is a simpler way to integrate Webpack into your application",
   "main": "index.js",
   "scripts": {
-    "test": "mocha --reporter spec test --recursive",
+    "test": "run-s test:*",
+    "test:normal": "mocha --reporter spec test --recursive --ignore test/persistent-cache/*",
+    "test:persistent-cache-clear": "cross-env CLEAR_PERSISTENT_CACHE=true npm run test:persistent-cache:base",
+    "test:persistent-cache": "npm run test:persistent-cache:base",
```
stof, line 10: I'd use `yarn run` rather than `npm run` given that we use yarn

package.json
```diff
@@ -4,7 +4,11 @@
...
+    "test:persistent-cache:base": "mocha --reporter spec test/persistent-cache --recursive",
```
stof, line 11: isn't this one also run directly by the `test` script ?

lib/webpack-manifest-plugin/hooks.js
bobvandevijver: I actually added as it is also stated in https://github.com/webdeveric/webpack-assets-manifest/blob/master/src/WebpackAssetsManifest.js#L735, and the fact that I couldn't find any real documentation about this parameter anywhere...

It does clarify why the `Object.assign` is used on line 134.

package.json
bobvandevijver: I totally agree, I only used npm because it was already used in the `travis:lint` command. Should I adjust them all?

package.json
bobvandevijver: It isn't, that would require the [globstar `**`](https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/run-s.md#glob-like-pattern-matching-for-script-names) instead of the used `*`

package.json
weaverryan: So we want to run `test/persistent-cache` twice? Once with `CLEAR_PERSISTENT_CACHE` and once without? Testing it both ways catches different behavior?

Would it be possible to, in `test/persistent-cache/functional.js`, loop over each test and run it twice each time? Or, purposely execute Webpack 2x for each test case: 

A) clear the cache
B) Execute webpack once (assert a few things are ok from a fresh cache)
C) Execute webpack again (assert a few things are ok from a NOT fresh cache)

I would love to get rid of the `cross-env`

package.json
```diff
@@ -4,7 +4,11 @@
   "description": "Webpack Encore is a simpler way to integrate Webpack into your application",
   "main": "index.js",
   "scripts": {
-    "test": "mocha --reporter spec test --recursive",
+    "test": "run-s test:*",
```
weaverryan, line 7: Why `run-s` vs `yarn test:normal && yarn:test-persistent-cache-clear && `, etc?

weaverryan (commented):
This is awesome! I'm SO happy someone was able to find the fix - thank you! I just have a few minor comments about the test organization... and hopefully slimming down (removing) the new dev dependencies.

Thanks!

package.json
bobvandevijver: Mainly because it is just simpler (to read) and easier to extend with new scripts in the same namespace in the future. And partly personal preference. 
If it needs to be adjusted, just say the word!

package.json
bobvandevijver: Actually, I tried that, and it didn't work :( 

The webpack cache is only written to disk (and thus available for a new invocation) when all tests have run (the mocha invocation is done) and the node process is exiting. I'm not sure why this is the case, but I couldn't figure out a way to work around this limitation. With my first iteration I actually tried running the webpack invocation twice by adding a `runTwice` parameter to the [`runWebpack`](https://github.com/symfony/webpack-encore/blob/972c8827500f3095dbaff06895a47ac0621bf28e/test/helpers/setup.js#L62) helper and so on. This was the only way I could get it to work reliably.

The order of the tests and cache creation are important, due to the persistent cache: originally, the cache files are always cleared when [the functional tests are started](https://github.com/symfony/webpack-encore/blob/972c8827500f3095dbaff06895a47ac0621bf28e/test/helpers/setup.js#L69). This is unwanted behaviour for the persistent cache tests (remember, I need to finish the node process to flush the webpack cache to disk), which is why I moved those tests into a separate test file. However, clearing the cache is still important for the first invocation, as we want that to be run without cache, which is why I added the ENV switch and the clear separation in scripts for transparency.

So, the final order becomes something like this:
  - `yarn run test`
    - `test:normal` is started
      - Whole cache folder (`test_tmp`) is cleared, which includes the webpack cache
      - All tests, except for the persistent cache tests, are run
      - No webpack cache is written, as there are no tests with persistent cache in this run
    - `test:persistent-cache-clear` is started
      - Whole cache folder (`test_tmp`) is cleared, which includes the webpack cache
      - Persistent cache tests are run, but without cache
      - Webpack cache is flushed to disk
    - `test:persistent-cache` is started
      - No cache is cleared, so the webpack cache now is available
      - Persistent cache tests are run, with cache
      - Updated cache is flushed to disk

Possible solutions:
  - We could rely on the cache clear from the earlier run normal tests: it works, but it not very transparent.
  - Find another crossplatform way of signaling the test that it should clear the cache on invocation. Running another script would also be fine with me by the way. 
  - Maybe the creation of a dedicated test file could help here, something like below. This removed the need of both new dev dependencies, but I'm not sure if this plays fair with all tooling.
```js
const { execSync } = require('child_process');
const { emptyTmpDir } = require('./test/helpers/setup');

execSync('mocha --reporter spec test --recursive --ignore test/persistent-cache/*', { stdio: 'inherit' });

emptyTmpDir();
for (let i = 0; i < 2; i++) {
    execSync('mocha --reporter spec test/persistent-cache --recursive', { stdio: 'inherit' });
}
```

bobvandevijver:
Alright, everything has been adjusted according to the feedback, so it should be ready for merge 👍🏻

bobvandevijver:
@weaverryan Is there anything left to do for me before this can be merged and released?

weaverryan:
Thank you Bob! This is truly an exciting PR :D :D

PR #983, opened 2021-05-20
Allow less-loader v9

bobvandevijver:
I'm not sure what the policy on node compatibility is, but the latest major of `less-loader` only seems to drop Node 10 support ([release notes](https://github.com/webpack-contrib/less-loader/blob/master/CHANGELOG.md)).

Webpack Encore still supports Node 10, but that should still be possible if you use one of the other allowed less loaders.

### DIFF
--- a/.github/workflows/node-windows.yml
+++ b/.github/workflows/node-windows.yml
@@ -19,7 +19,10 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node }}
+    # needed so that Node 10 downloads a compatible less-loader version
+    - name: Clear lock file
+      run: del yarn.lock
     - name: yarn install
-      run: yarn install
+      run: yarn --ignore-engines
     - name: run tests
       run: yarn test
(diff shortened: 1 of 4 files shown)


weaverryan:
You're correct - there's no problem with allowing less-loader 9, as we still allow versions that support Node 10.

However, it's definitely made a mess for the tests. I've just pushed what will hopefully be a workaround.

bobvandevijver:
Well, Node 10 is no longer supported since the beginning of this month, so we could consider dropping support for it altogether if that makes it easier 😀

weaverryan:
I’m concerned about ubuntu 20.04 LTS, which I believe comes with node 10. And to bump to 12 just for this reason would be unfortunate :)

weaverryan:
Plus, I think my last commit fixed things. Appveyor is broken still, but I’m not sure we need to anymore, as we have windows tested by GH actions.

bobvandevijver:
@weaverryan Is there anything left to do for me before this can be merged and released?

weaverryan:
Thank you Bob!

PR #999, opened 2021-06-03
Allow sass-loader ^12.0.0, close #996

### LINKED ISSUE #996: Support sass-loader v12
sass-loader v12 is currently not supported. 

This has the same discussion about Node 10 as #993, as this release also drops it.

Kocal:
The only breaking change is that the minimum supported version of Node.js is 12.13.0,
see https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#1200-2021-06-01.


BackEndTea (approved):

weaverryan:
Thanks @Kocal!

PR #1000, opened 2021-06-03
Allow ts-loader ^9.0.0, close #993

### LINKED ISSUE #993: Support ts-loader 9
ts-loader v9 is currently not supported.

Kocal:
Like #999, the only breaking change is that the minimum supported version of Node.js is 12.0.0,
see https://github.com/TypeStrong/ts-loader/blob/main/CHANGELOG.md#v900

**EDIT:** PR #1000 :exploding_head: 

**EDIT 2:** Looks like something broke. It happens since 9.0.0 (exact version) but I don't know what happens. 

**EDIT 3:** Added TypeScript types for `.vue`, and it's working fine. Not sure if it's really a BC break, I expect most of people using Vue 3 + TypeScript already using a `vue-shims.d.ts` file... :confused:

### DIFF
--- a/fixtures/vuejs3-typescript/types/shims-vue.d.ts
+++ b/fixtures/vuejs3-typescript/types/shims-vue.d.ts
@@ -1,0 +1,6 @@
+declare module '*.vue' {
+  import { defineComponent } from 'vue';
+
+  const component: ReturnType<typeof defineComponent>;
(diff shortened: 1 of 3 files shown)


weaverryan:
Congrats on #1000 @Kocal! And thank you as always :)

PR #1009, opened 2021-06-22
Allow less-loader 10

bobvandevijver:
Looks like the removed a [less context option](https://github.com/webpack-contrib/less-loader/releases/tag/v10.0.0), but I could find any place where it is used in this package.


weaverryan:
And thank you again! :)

PR #1028, opened 2021-08-30
Upgraded resolve-url-loader to remove all PostCSS 7 dependents

TropicalRaisel:
Mixing PostCSS 7 & 8 is a [bad idea](https://github.com/cssnano/cssnano/issues/1185#issuecomment-908237779).


package.json
```diff
@@ -42,7 +42,7 @@
     "mini-css-extract-plugin": "^1.5.0",
     "pkg-up": "^3.1.0",
     "pretty-error": "^3.0.3",
-    "resolve-url-loader": "^3.1.2",
+    "resolve-url-loader": "^4.0.0",
```
weaverryan, line 45: But v4 of resolve-url-loader still depends on postcss 7: https://github.com/bholloway/resolve-url-loader/blob/v4-maintenance/packages/resolve-url-loader/package.json#L41

Are you sure that this is fixing your issue?

TropicalRaisel: Ah I didn't notice that. As to whether the issue is actually solved once everything gets updated to PostCSS 8 I'm not sure. I'm not [entirely sure `resolve-url-loader` is the problem](https://github.com/cssnano/cssnano/issues/1185#issuecomment-908492947).

PR #1034, opened 2021-08-31
[ci] move unix tests to Github Actions

jrushlow:
- replace travis ci w/ github actions
- consolidates unix / windows tests
- adds testing against `Node v16`
- CI Runs every Sunday at Midnight for `Stable`, `Low`, & `High` tests

Stable tests run against Node `10`, `12`, `14`, & `16`
Low / High dependency tests run against Node `10` & `16`
All tests run on Windows & Linux

Failing `lowest` tests related to Typescript `v3.6.3` - to be fixed under separate PR
Node `10` support to be dropped in separate PR


.github/workflows/high-depends.yml
```diff
@@ -0,0 +1,43 @@
+name: High Dependencies
+on:
+    push:
+    pull_request:
+    schedule:
+        -   cron: '0 */12 * * *'
```
weaverryan, line 6: Running these twice a day seems excessive - what about once a week?

.github/workflows/stable-tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Locked Dependencies
+on:
+    push:
+    pull_request:
+    schedule:
+        -   cron: '0 */12 * * *'
```
stof, line 6: running them twice a day in addition to running them on every push seems excessive

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,26 @@
+name: ESLint
+on:
+    push:
+    pull_request:
+    schedule:
+        - cron: '0 */12 * * *'
```
stof, line 6: Running eslint on a schedule is useless. New eslint violations won't ever be introduced without a push.

jrushlow: Good point - my head was over in MakerBundle where we run twice a day to catch upstream releases that break things

weaverryan:
Thanks Jesse!

PR #1037, opened 2021-09-01
[dependencies] drop node 10 support

jrushlow:
Node 10 is beyond it's EOL.

- waiting on #1034


stof:
Note that due to the way the npm ecosystem works, this requires being released as a major version (as the node compatibility is not taken into account when selecting a version to install, it only warns during the installation step)

weaverryan:
I'd like to keep this - https://github.com/symfony/webpack-encore/pull/912 - unless it really starts causing us problems :)

PR #1056, opened 2021-10-28
Support AVIF images

benbankes:


benbankes:
It doesn't feel like these test failures are a result of my change.  I am using this change successfully with avif images.

weaverryan:
Sorry for the delay! My guess is also that the test failures were unrelated, and have quite possibly been fixed upstream since then (`main` is green right now). I've just rebased your work onto #1074. 

Cheers!

PR #1058, opened 2021-11-04
Fix deprecated public option failure

### LINKED ISSUE #1044: [regression] dev-server --public option failure
Hi, 

The option `encore dev-server --public` used to work to specify which host to use in the manifest.json symfony assets file.

But in the current version (1.6.1), this option fails:

```shell
ban@na:/var/www/html$ ./node_modules/.bin/encore dev-server --public http://192.168.1.22:8000

[webpack-cli] Error: Unknown option '--public'
[webpack-cli] Run 'webpack --help' to see available commands and options
```

Looking at webpack-encore code, apparently this is expected to be still working (at least it’s tested):

https://github.com/symfony/webpack-encore/blob/8c2752afd4d63d0e72c757b96af6ae2c2d3c85cf/test/config/parse-runtime.js#L96-L101

### Versions

| Package | Version |
|-----|-----|
|webpack-encore| `1.6.1`|
|webpack-dev-server|`4.1.1`|

### Notes about our use case

We use webpack-encore in a docker-compose environment. With the encore dev-server running in its own container, and the symfony app in another.

This setup works very well (serving assets and livereload) as long as the browser shares the same host as the docker containers (0.0.0.0, or localhost is ok). But, we are testing with real devices on the local network (with ips like 192.168.0.12, or even virtualhosts), and currently it seems that we cannot specify the public host to put in the manifest.json.

Thanks :)

atesca09:
The --public option used to specify which host to use in the manifest.json.

That options was deprecated in favor of client.webSocketURL

Fixes #1044

### DIFF
--- a/lib/config/parse-runtime.js
+++ b/lib/config/parse-runtime.js
@@ -44,8 +44,8 @@ module.exports = function(argv, cwd) {
             runtimeConfig.devServerHttps = argv.https;
             runtimeConfig.devServerKeepPublicPath = argv.keepPublicPath || false;
 
-            if (typeof argv.public === 'string') {
-                runtimeConfig.devServerPublic = argv.public;
+            if (typeof argv['client-web-socket-url'] === 'string') {
+                runtimeConfig.devServerPublic = argv['client-web-socket-url'];
             }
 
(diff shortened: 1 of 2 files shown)


weaverryan:
This looks very nice - thank you @atesca09!

PR #1062, opened 2021-11-30
Allowing @hotwired/stimulus, allowing @symfony/stimulus-bridge 3, dropping v1

weaverryan:
Hi!

This does 3 things:

1) Allows `@hotwired/stimulus` when working with `enableStimulusBridge()`. More precisely, it just removes the requirement of Stimulus entirely... it's not technically needed in THIS library.

2) Add support for the upcoming `@symfony/stimulus-bridge` version 3

3) Drop support for `@symfony/stimulus-bridge` version 1


stof:
tests are broken, because the tests are written expecting v3 of the bridge (as they use stimulus 3) while the yarn.lock contains v2 of the bridge.

stof:
given that @symfony/stimulus-bridge 3 is not released yet, the CI is just skipping the tests now

weaverryan:
> given that @symfony/stimulus-bridge 3 is not released yet, the CI is just skipping the tests now

You're too fast Stof! I was typing the below message when your message popped up :).

I've just skipped the test for stimulus-bridge 2.0 (we COULD create a completely separate test for `stimulus` v2, but it would require duplicating a fixtures directory and adding another "fake" package to `package.json` - I don't think it's worth it). Locally, I temporarily forced the latest (eventually v3) version of stimulus-bridge and the test passes.

fixtures/vuejs3-typescript/App.vue
```diff
@@ -1,12 +1,13 @@
...
   <div id="app">
-    <img src="./assets/logo.png">
+    <img :src="logoPath">
     <hello></hello>
   </div>
 </template>
 
 <script lang="ts">
 import Hello from './components/Hello.vue'
+import logoPath from './assets/logo.png';
```
weaverryan, line 10: This and the `package.json` files are unrelated, but fix the tests. I'm not sure what's going on - it's super complex. It relates to this commit - https://github.com/vuejs/vue-loader/commit/21725a4ebc9c8d7f8a590d700017759327e21c2e - and vue-loader 16.7.0. Our testing setup is very vanilla, but for some reason, the `` no longer worked. I have no idea how to fix that - and it's so strange, that, for now, I'm going to assume it's some edge-case issue specific to our app or some bug that will be fixed upstream.

weaverryan:
This is ready to go now

PR #1070, opened 2022-01-08
New Encore method for adding multiple entries at once

shmolf:
Purpose: I dislike having to call `addEntry` multiple times, and would prefer the cleaner look of a single hash.

Example Usage:
```js
Encore.addEntries({
  main: './path/to/main.js',
  secondary: './path/to/another/file.js',
});
```

This supports validation that any keys provided in the Object literal do not yet exist as an Entrypoint or Style Entrypoint.

Tests have also been updated.

---

Another new method (`validateNameIsNewEntry`) was also added to extract and consolidate logic related to checking for duplicate entrypoint names.

Tests for this other function have also been added.

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -346,27 +346,27 @@ class WebpackConfig {
     }
 
     addEntry(name, src) {
-        if (this.entries.has(name)) {
-            throw new Error(`Duplicate name "${name}" passed to addEntry(): entries must be unique.`);
-        }
-
-        // also check for styleEntries duplicates
-        if (this.styleEntries.has(name)) {
-            throw new Error(`The "${name}" passed to addEntry conflicts with a name passed to addStyleEntry(). The entry names between addEntry() and addStyleEntry() must be unique.`);
-        }
+        this.validateNameIsNewEntry(name);
 
         this.entries.set(name, src);
     }
 
-    addStyleEntry(name, src) {
-        if (this.styleEntries.has(name)) {
-            throw new Error(`Duplicate name "${name}" passed to addStyleEntry(): entries must be unique.`);
+    /**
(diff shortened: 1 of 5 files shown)


lib/WebpackConfig.js
```diff
@@ -346,27 +346,32 @@ class WebpackConfig {
...
-        // also check for entries duplicates
-        if (this.entries.has(name)) {
-            throw new Error(`The "${name}" passed to addStyleEntry() conflicts with a name passed to addEntry(). The entry names between addEntry() and addStyleEntry() must be unique.`);
-        }
+        Object.keys(entries).forEach((key) => this.validateNameIsNewEntry(key));
+
+        this.entries = new Map([
+            ...this.entries,
+            ...Object.entries(entries),
+        ]);
```
Kocal, line 370: To me it would have been simpler to: 
- not update `addEntry` method
- don't introduce a new `validateNameIsNewEntry` method
- iterate over `entries` and call `this.addEntry()` 

WDYT?

shmolf: Extracting out logic into `validateNameIsNewEntry` still makes sense, for single responsibility.

Regarding whether `addEntries` should call `addEntry` in a loop, to set the data, or do a merge directly...
- by calling `addEntry` in a loop, a single method is responsible for updating `this.entries`, which is good

shmolf: @Kocal I've pushed up another commit, taking into some of your recommendations. Please resolve this thread, if you find the latest changes meet your concerns.

Kocal: It looks good to me, unfortunately I don't have the butto, "Resolve this thread" :grimacing: (maybe because my comment was not a review comment)

weaverryan (approved):
This is really, really nicely done! I don't know if personally need the feature, but it's subjective, and you implemented this beautifully! 👍

PR #1033, opened 2021-08-31
Allowing new major versions of outdated dependencies

weaverryan:
Fixes #1029

~~A bit more work to do. For now, I'm focused only on the main dependencies (not the `devDependencies`).~~

~~You could argue that this will require a new major version of webpack-encore. However, there is a precedence for allowing major dependency upgrades on minor versions in the past: https://github.com/symfony/webpack-encore/releases/tag/v1.2.0. Assuming none of the libraries have underlying changes that are likely to affect users, I'd like to do that again.~~ Edit: likely we should have a new major release.

## Dependencies

* ~`chalk`~: kept at v4 because v5 requires esm. v4 is stable.
* `clean-webpack-plugin`: `3->4` No breaking changes
* `css-loader`: `5->6`: Min node to 12.3.0,  other changes: [CHANGELOG](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#-breaking-changes)
* `css-minimizer-webpack-plugin`: `2->3` Min node to 12.3.0
* `loader-utils`: REMOVED
* mini-css-extract-plugin `1.5->2.2.1` Min node to 12.13.0, Webpack to 5.0.0, other changes: [CHANGELOG](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/CHANGELOG.md#-breaking-changes) 
- ~`pkg-up`~: Keeping version `3.1` because `4` requires esm.
- `pretty-error`: `3.0->4.0` Min node to 12.0 
- `resolve-url-loader` `3.0->5.0` Min node to 12.0, postcss ^8.0, remove `rework` engine,  other changes: [CHANGELOG](https://github.com/bholloway/resolve-url-loader/blob/v4-maintenance/packages/resolve-url-loader/CHANGELOG.md)
- `style-loader`: `2->3` Min node to 12.13, Min webpack to 5.0 other changes: [CHANGELOG](https://github.com/webpack-contrib/style-loader/blob/master/CHANGELOG.md#-breaking-changes)
- `yargs-parser`: `20.2->21` Min node to 12.0

## Dev Dependencies
- `eslint`: `7->8` Min node to `12.22`
- `eslint-webpack-plugin`: `2.5->3` min node to `12`. Min webpack `5`
- `fork-ts-checker-webpack-plugin`: `5->6` [CHANGELOG](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/releases/tag/v6.0.0)
[...]

PR #1078, opened 2022-01-21
Upgrade to resolve-url-loader v5

stof:
This is a slight BC break in case someone configures the resolveUrlLoaderOptions when enabling the sass-loader.

BC breaks in that loader between v3 and v5:
- node 12 is required ~~(fine as Encore already requires it)~~ not actually fine.
- the `engine` option is gone, keeping only the postcss-based engine. The `rework` engine was not recommended anymore in v3, officially deprecated in v4 and removed in v5. given that the `rework` library is essentially unmaintained and has lots of bugs with modern CSS, I strongly doubt that any webpack-encore users would have made the choice to use that old engine instead of the default one
- the `join` option changed the signature of the function. This is a very advanced customization option

This will avoid many warnings when installing dependencies:
- `rework` depends on several unmaintained packages that have been deprecated in the npm registry
- The `postcss` version used by this loader has been updated to postcss 8, which avoids the warning about postcss 7 being vulnerable to ReDoS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "mini-css-extract-plugin": "^1.5.0",
     "pkg-up": "^3.1.0",
     "pretty-error": "^3.0.3",
-    "resolve-url-loader": "^3.1.2",
+    "resolve-url-loader": "^5.0.0",
     "semver": "^7.3.2",
     "style-loader": "^2.0.0",
     "sync-rpc": "^1.3.6",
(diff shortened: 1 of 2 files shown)


stof:
Actually, I was wrong. Encore still supports node 10.

@weaverryan I strongly suggest doing a new major version of Encore bumping the min version to node 12 (or even node 14 as node 12 is EOL in April) instead of keeping support for the EOL node 10. More and more npm packages are bumping their min version.

stof:
btw, our CI uses `yarn install --ignore-engines` on node 10, which is a sign that we already do a mess there (which might hide warnings that projects using encore would see on node 10)

weaverryan:
I was thinking about dropping 10 and doing a major release also. I have another PR open that upgrades EVERYTHING that’s out of date, with the idea of making the most out of our major release.

weaverryan:
> btw, our CI uses yarn install --ignore-engines on node 10, which is a sign that we already do a mess there (which might hide warnings that projects using encore would see on node 10)

To match up with Ubuntu's LTS version, and its Node version, I was trying to hold onto Node 10 support as long as possible. But because of what you mentioned, it was already shaky.

stof:
@weaverryan I opened https://github.com/symfony/webpack-encore/issues/1079 to discuss that.

weaverryan:
Closed in favor of #1033, which is ready now

PR #1080, opened 2022-01-25
Upgrade resolve-url-loader

y4roc:
Security-Issue: Regular Expression Denial of Service in postcss

resolve-url-loader updated this dependency in 3.1.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "mini-css-extract-plugin": "^1.5.0",
     "pkg-up": "^3.1.0",
     "pretty-error": "^3.0.3",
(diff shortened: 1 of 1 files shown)


dragonito (approved):

j0k3r (approved):

weaverryan:
Sorry for the slow reply - this will be included in #1033. But, users can naturally upgrade their own projects already, and the vulnerability doesn't affect Encore anyways (it's only if you use `postcss` to parse CSS, which we don't): https://github.com/postcss/postcss/commit/8682b1e4e328432ba692bed52326e84439cec9e4#commitcomment-49809613

Cheers!

PR #1095, opened 2022-02-17
Revert removing public option

### LINKED ISSUE #1094: HMR when using Encore in virtual machine
Since webpack-dev-server 4.0.0, public option is not available.  
The PR #1058 replace public with client-web-socket-url to set url used to generate entry asset in entrypoints.json.  

But when we use client-web-socket-url option with a custom domain, url used for the websocket url is wrong.  The path /ws is missing and HMR doesn't work.

When I read the webpack-dev-sever [documentation](https://webpack.js.org/configuration/dev-server/#websocketurl) for the client-web-socket-url, I think it's not a good idea to use the webpack option client-web-socket-url to set the public url because the format is not the same.

I will send a PR which should solve the problem.

louismariegaborit:
Add --public option like an encore argument to specify public url that use for entry asset in entrypoints.json. 

I have a problem with this suggestion when I try to build webpack but the result is correct.
Anyone can help me ?

Capture d’écran 2022-02-17 à 20 43 09

Fixes #1094

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -33,6 +33,12 @@ process.argv.splice(2, 1);
 const encoreOnlyArguments = new Set(['--keep-public-path']);
 process.argv = process.argv.filter(arg => !encoreOnlyArguments.has(arg));
 
+// remove argument --public not supported by webpack/dev-server
+const indexPublicArgument = process.argv.indexOf('--public');
+if (indexPublicArgument !== -1) {
+    process.argv.splice(indexPublicArgument, 2);
+}
+
 if (!runtimeConfig.isValidCommand) {
     if (runtimeConfig.command) {
         console.log(chalk.bgRed.white(`Invalid command "${runtimeConfig.command}"`));
@@ -78,6 +84,7 @@ function showUsageInstructions() {
     console.log(`       - ${chalk.yellow('--host')} The hostname/ip address the webpack-dev-server will bind to`);
     console.log(`       - ${chalk.yellow('--port')} The port the webpack-dev-server will bind to`);
     console.log(`       - ${chalk.yellow('--keep-public-path')} Do not change the public path (it is usually prefixed by the dev server URL)`);
+    console.log(`       - ${chalk.yellow('--public')} The public url for entry asset in entrypoints.json`);
     console.log('       - Supports any webpack-dev-server options');
     console.log();
     console.log(`    ${chalk.green('production')} : runs webpack for production`);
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks for the PR. This stuff is very complex, and I am not an expert on it: I haven't had time to dig deeply enough.

But, we have a problem :). This PR (as you know) reverts #1058. So #1058 was purposely done to fix an issue... and now we're reversing it. It makes me think that we're not understanding the full story and that the correct solution might be some combination of the two.

WDYT?

louismariegaborit:
Yes, I will try to explain the problem that I encounter.  

The PR #1058 replace the public option with the client-web-socket-url option like it was described in the webpack-dev-server upgrade doc.  

If I run webpack-dev-server with `--client-web-socket-url http://app.domain:8080` the entrypoints.json file content is ok and assets are loaded but HMR not working.  
The url generated for the websocket is `ws://app.domain:8080/` and is not working.  

When I remove the option and I set the parameter `options.client={webSocketUrl: ws://app.domain:8080/ws}` in webpack-config.js file (note that /ws is important) HMR works but entrypoints.json is malformed because `--client-web-socket-url` was not used for encore and only webpack-dev-server.  
I was succeed to run correctly my projet with HMR and entrypoints.json by defining hard the runtimeConfig.devServerPublic var with my url http://app.domain:8080 (without /ws) in runtime.js file.  

The target of my PR is that --client-web-socket-url is a webpack-dev-server option which protocol start with auto://, ws:// or wss:// and in encore is override with an url start with http:// or https://.  
The idea is to let `--client-web-socket-url` to webpack-dev-server and integrate `--public` as an encore option like `--keep-public-path` to define the url used for entrypoints.json that is an internal mechanisme of encore.  

I hope I was clear...

I'm open for any other solutions to solve the usage of HMR with a custom domain.

shadowc:
I for once, back this PR. I think it's on the right track.

I would LOVE to ALSO see an option to configure this somewhere within the webpack encore config file as well.

Something like:

`.setPublicDevServerUrl(path: string)`

That would achieve the exact same result as providing the `--public` option

louismariegaborit:
Thanks @shadowc for your interest.

Add an option in webpack_config.js can be a solution but I don't think that is the best because the base url depends on the environment where you execute webpack-dev-server. 
An another developer on the same projet can decide to use another url for his environment.

@weaverryan an opinion ?

louismariegaborit:
I just fixed the error I had at the build.
It's ready for a review and a possible merge.
Always available to discuss the subject which seems quite blocking to me.

weaverryan:
Thank you @louismariegaborit! @shadowc would you mind creating a docs PR for this, including how it could be used in a Docker context?

PR #1118, opened 2022-05-13
Use cli param server-type to define devServer https mode

thegillou:
Check also the serverType cli param to define if the devServer is in https mode


weaverryan:
This makes sense - thanks Gilles!

PR #1133, opened 2022-07-06
Increasing dependencies

weaverryan:
This is (probably) the last PR before releasing Encore 3.0 - it bumps all out-of-date dependencies (except for a few internal libraries):

* `css-minimizer-webpack-plugin`  3.4->4.0 (just dropping Node 12 support)
* `less-loader` 10->11
* `postcss-loader` 6->7
* `sass-loader` 13 (drops support for 12)
* `stylus` 0.57->0.58
* `styus-loader` 6->7

Cheers!


stof:
@weaverryan any reason to lock `assets-webpack-plugin` to `7.0.*` instead of using the 7.1 version ?

PR #1135, opened 2022-07-17
Accept a symbolic link as target of copyFrom

alsciende:
There's no reason why the `from` option of `copyFrom` cannot be a link to a directory. By using `stat` instead of `lstat`, we are allowing just that.


alsciende:
Actually that doesn't work, files get copied under a bunch of directories named `_`. I'm just going to put my assets directly into `public`.

Issue #1154, opened 2022-10-01
Fix error with eslint due to variable initialization before checking, eslint was set


wesoledi:
In some projects, with symfony 6 with encore used, to work with sass+vue, without external eslint configuration, there may be error while building final dist/build files. The reason was, it could not init eslint, even it was disabled in webpack configuration (or more precisly: not enabled). Webpack-encore still wanted to run it, before checking if "useEslintPlugin" was set. 

To fix this, I moved variable initialization to the block, after checking this variable, because it is used only inside if-block anyway. So there is no need to init these before. 

Here is my compare: https://github.com/symfony/webpack-encore/commit/bc34f910dfa189016a6f7563927ea752479c0695

lu43n:
The same here. With fix [https://github.com/symfony/webpack-encore/commit/bc34f910dfa189016a6f7563927ea752479c0695](url) from @wesoledi everything works. Before fix I got errors:


```
yarn run v1.22.17
$ /usr/home/account/domains/domain.com/node_modules/.bin/encore dev
Running webpack ...

events.js:377
      throw er; // Unhandled 'error' event
      ^

Error: listen EPERM: operation not permitted 0.0.0.0:20868
    at Server.setupListenHandle [as _listen2] (net.js:1314:21)
    at listenInCluster (net.js:1379:12)
    at Server.listen (net.js:1465:7)
    at Object.<anonymous> (/usr/home/account/domains/domain.com/node_modules/sync-rpc/lib/worker.js:68:8)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)
    at internal/main/run_main_module.js:17:47
Emitted 'error' event on Server instance at:
    at emitErrorNT (net.js:1358:8)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  code: 'EPERM',
  errno: -1,
  syscall: 'listen',
  address: '0.0.0.0',
  port: 20868
}
[webpack-cli] Failed to load '/usr/home/account/domains/domain.com/webpack.config.js' config
[webpack-cli] Error: Timed out waiting for sync-rpc server to start (it should respond with "pong" when sent "ping"):

[...]

PR #1155, opened 2022-10-09
move const def under main condition - do not init until needed

wesoledi:
https://github.com/symfony/webpack-encore/issues/1154

### DIFF
--- a/lib/plugins/eslint.js
+++ b/lib/plugins/eslint.js
@@ -9,13 +9,6 @@
 
 'use strict';
 
-const forceSync = require('sync-rpc');
-const hasEslintConfiguration = forceSync(require.resolve('../utils/has-eslint-configuration'));
-const WebpackConfig = require('../WebpackConfig'); //eslint-disable-line no-unused-vars
-const applyOptionsCallback = require('../utils/apply-options-callback');
-const pluginFeatures = require('../features');
-const babelLoaderUtil = require('../loaders/babel');
-
 /**
  * Support for ESLint.
  *
@@ -25,6 +18,13 @@ const babelLoaderUtil = require('../loaders/babel');
  */
 module.exports = function(plugins, webpackConfig) {
     if (webpackConfig.useEslintPlugin) {
+        const forceSync = require('sync-rpc');
+        const hasEslintConfiguration = forceSync(require.resolve('../utils/has-eslint-configuration'));
+        const WebpackConfig = require('../WebpackConfig'); //eslint-disable-line no-unused-vars
+        const applyOptionsCallback = require('../utils/apply-options-callback');
+        const pluginFeatures = require('../features');
+        const babelLoaderUtil = require('../loaders/babel');
+
         pluginFeatures.ensurePackagesExistAndAreCorrectVersion('eslint_plugin');
 
         if (!hasEslintConfiguration(webpackConfig)) {


wesoledi:
Fix for error, while running 'npm run build' on some enviroments (with restriction on available ports). 

The error was: 
```
[webpack-cli] Failed to load '<path>/webpack.config.js' config
[webpack-cli] Error: Timed out waiting for sync-rpc server to start (it should respond with "pong" when sent "ping"):
```

More here: https://github.com/symfony/webpack-encore/issues/1154

Kocal (approved):

lib/plugins/eslint.js
```diff
@@ -25,6 +18,13 @@ const babelLoaderUtil = require('../loaders/babel');
  */
 module.exports = function(plugins, webpackConfig) {
     if (webpackConfig.useEslintPlugin) {
+        const forceSync = require('sync-rpc');
+        const hasEslintConfiguration = forceSync(require.resolve('../utils/has-eslint-configuration'));
```
weaverryan, line 22: Can we move ONLY these 2 lines and put the below 4 back on top? For clarity, I'd like to only move the items that are actually creating the problem.

weaverryan: good point - it is likely just the `hasEslintConfiguration` that is the problem since it is USING `sync-rpc`

wesoledi: Importing sync-rpc it **is** the main problem for this error. It tried to connect on initialization on import so it is mandatory, to fix this issue.  
Can we only move this one? Maybe, I didnt tried. Maybe you are right, sync-rpc can stay, but I moved all, because I believe, there is no point to init or import all of this things if we never use it. Waste of resources in my opinion. Import and use it only when we need to.

weaverryan: > Waste of resources in my opinion. Import and use it only when we need to.

It is a matter of code clarity. We normally expect `require` statements to be at the top of the file. If we start requiring them inline, it'll make the code harder to follow. That's a waste of human resources :). That's why we want to move only the one line needed. That one line actually should have always been moved down here because we are USING a library, not just doing a simple `require`. But the other `require` are standard and should be at the top.

stof: @wesoledi the error in the reported issue has a stack trace pointing to line 13, which means the line **calling** `forceSync()`, not the line importing it. So `require('sync-rpc')` is not the issue.

stof: Thus, if we migrate to ESM imports in the future (not planned yet, but it might happen at some point as that's the future of the ecosystem), imports are required to be at the top-level. So our existing convention of requiring deps at the top-level will make that migration easier.

michalhorak:
Please allow me to ask you to reconsider the initial request.
I just had some disturbing moments looking for a solution in all wrong places at first. 
2022-12-19 21_15_24-sync-rpc-out
The problem appeared irregularly. When it started to appear, it was always just the first run (bearable). Recently it became much more frequent. As I am sort of learning to use your wonderful software, I first attempted to exclude all possible misconfigurations on my side, removed node_modules, antivirus ... I did not at all know there is any use of eslint (which I do not use with encore) during packing.

After finally reading this thread I manually performed the request in local 'node_modules' and all is well.

IMHO:
Importing files which are not at all used seems as unnecessary delay to me. I was moreover at first confused and thought I have some misconfiguration in .eslintrc.json (which I use only with IDE), and that is somehow parsed by Encore and disturbes it. By various experiments I discovered it is not the case.

In case of transition to ES6 modules you can use [import()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) for the same purpose.
Thank you.

weaverryan:
> Please allow me to ask you to reconsider the initial request.

Oh, I definitely support this PR :). We just got tangled in some final details. I've finished it over in #1175. I'd love if you could confirm that this small change fixes your issue.

Closing this PR - as the original work + new tweaks is over on #1175.

Cheers!

Issue #1165, opened 2022-11-21
Dependency issues with Vue 2.7


dmaicher:
I have some issues with Vue 2.7 and the latest release of `@symfony/webpack-encore@4.1.1`

```
npm ERR! While resolving: @symfony/webpack-encore@4.1.1
npm ERR! Found: @vue/compiler-sfc@2.7.14
npm ERR! node_modules/@vue/compiler-sfc
npm ERR!   @vue/compiler-sfc@"^2.7.14" from the root project
npm ERR!   @vue/compiler-sfc@"2.7.14" from vue@2.7.14
npm ERR!   node_modules/vue
npm ERR!     vue@"^2.6.12" from the root project
npm ERR!     peer vue@"^2.0.0 || >=3.0.0" from @cumul.io/vue-cumulio-dashboard@2.0.13
npm ERR!     node_modules/@cumul.io/vue-cumulio-dashboard
npm ERR!       @cumul.io/vue-cumulio-dashboard@"^2.0.6" from the root project
npm ERR!     16 more (@customer-alliance/ca-ui-components, ...)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peerOptional @vue/compiler-sfc@"^3.0.2" from @symfony/webpack-encore@4.1.1
npm ERR! node_modules/@symfony/webpack-encore
``` 

Should this also allow `^2.7`?

https://github.com/symfony/webpack-encore/blob/main/package.json#L114

stof:
yes, it should

dmaicher:
For now I fixed it with 

```json
    "overrides": {
        "@symfony/webpack-encore": {
            "@vue/compiler-sfc": "^2.7 || ^3.0.2"
        }
    }
```

PR #1166, opened 2022-11-21
Match `@vue/compiler-sfc` requirement with `vue`

dmaicher:
Fixes https://github.com/symfony/webpack-encore/issues/1165

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "@symfony/stimulus-bridge": "^3.0.0",
     "@vue/babel-helper-vue-jsx-merge-props": "^1.0.0",
     "@vue/babel-preset-jsx": "^1.0.0",
-    "@vue/compiler-sfc": "^3.0.2",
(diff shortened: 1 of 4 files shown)


stof:
Why making it a draft PR ?

dmaicher:
Do I need to fix the yarn fails?

stof:
@Kocal any idea what those failures are about ? To me, it seems like yarn says it fails with errors without actually showing those errors.

dmaicher:
Yeah I cannot fix those.

I tried

```
yarn set version berry
yarn install
```

generates a huge yarn.lock diff and still fails

stof:
ah indeed. I missed that one of those symbols was a GHA group and not the Yarn prefix...

Kocal (approved):

weaverryan:
Thanks @dmaicher and everyone else!

Issue #1172, opened 2022-12-09
Yarn warning when updating webpack-cli to v5.0


UlrichHP:
Hello,

After updating `webpack-cli` to `v5.0.1` in a symfony project, yarn berry (v3.3.0) now show this warning during a `yarn install` in the resolution step:

```
➤ YN0060: │ ***@workspace:. provides webpack-cli (p146bb) with version 5.0.1, which doesn't satisfy what @symfony/webpack-encore and some of its descendants request
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
```

Is there something I can do about it ? Or will you update the bundle requirements ?

Thanks

stof:
Looking at the webpack-cli changelog, I think encore should work fine with v5. So it should be safe to update the peer dependency of webpack-encore to allow both 4.x and 5.x

UlrichHP:
Do you want me to do a PR updating the package.json webpack-cli peerDependency ?

PR #1173, opened 2022-12-14
chore: updated webpack-cli peerDependencies with v5

UlrichHP:
Following issue #1172 

In `package.json` : updated webpack-cli peerDependencies and reordered svelte in devDependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,16 +90,16 @@
     "strip-ansi": "^6.0.0",
     "stylus": "^0.58.1",
     "stylus-loader": "^7.0.0",
+    "svelte": "^3.50.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -98,8 +100,6 @@
     "webpack": "^5.72",
     "webpack-cli": "^4.9.1",
```
weaverryan, line 101: Also change it here. This is used in the tests - and we test both the upper and highest matching versions.

UlrichHP: I wasn't sure if it was needed here also.
Thanks, it's done.

weaverryan:
Thanks @UlrichHP!

PR #1188, opened 2023-02-22
Update docblock for configure options to fix filename output

bastnic:
generated is `flags.9a7017d3.png`, not `flags.9a7017d3png`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1400,7 +1400,7 @@ class Encore {
      * Encore.configureFilenames({
      *     js: '[name].[contenthash].js',
      *     css: '[name].[contenthash].css',
-     *     assets: 'assets/[name].[hash:8][ext]',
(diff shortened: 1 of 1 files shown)


bastnic:
nevermind, this is correct, but I don't understand why the dot is necessary in js/css but not for assets

stof:
This is because JS and CSS don't use the `[ext]` placeholder. So they have to provide not only the `.` but also the `js` chars to end with `.js`.

The `[ext]` placeholder includes the `.`

bastnic:
And I was confused about that because in this template, the [ext] does NOT contain the `.`

```js
  .copyFiles({
    from: './assets/images',
    to: `images/[path][name].[ext]?[hash:8]`,
  })
```

PR #1190, opened 2023-02-28
Allow eslint-webpack-plugin ^4.0

evertharmeling:
`eslint-webpack-plugin:^4` dropped support for `node:12` and `eslint:7`, this library already is on higher versions (`"node": ">=14.0.0"` and `"eslint": "^8.0.0"`), so this should not be a problem.

Or should we still support `eslint-webpack-plugin:3` too?


stof:
This dev dependency should stay compatible with the range allowed in the peer dependency.

And reducing a peer dependency range is a BC break, so support for eslint-webpack-plugin v3 must be kept until the next major version of Encore.

stof (approved):

weaverryan:
Thanks @evertharmeling!

Issue #1140, opened 2022-07-28
Copy files without entries?


CapInSpase:
Hi, I want to copy static files into a separate command, but I get an error when executing

**No entries found! You must call addEntry() or addEntries() or addStyleEntry() or copyFiles() at least once - otherwise... there is nothing to webpack!**

//config for copy static files
```
Encore
    .setOutputPath('public/assets/')
    .setPublicPath('/assetic')
    .setManifestKeyPrefix('assetic/')

    .disableSingleRuntimeChunk()
    .cleanupOutputBeforeBuild()
    .enableBuildNotifications()

    .addPlugin(new CopyWebpackPlugin({
        patterns: [
            { from: './assets/sites', to: '../sites', info: { minimized: true } },
            { from: './assets/bundles', to: '../bundles', info: { minimized: true } },
            { from: './assets/vendor', to: 'vendor', info: { minimized: true } },
        ]
    }))
;
const copyStaticConfig = Encore.getWebpackConfig();
copyStaticConfig.name = 'copyStaticConfig';
Encore.reset();

module.exports = [copyStaticConfig];
```

If use copyFiles() how to set multiply path and not minimized?
[https://symfonycasts.com/screencast/webpack-encore/copy-files](https://symfonycasts.com/screencast/webpack-encore/copy-files)

gimler:
See Plugin docu https://webpack.js.org/plugins/copy-webpack-plugin/#skip-running-javascript-files-through-a-minimizer

For the Error i will Open a pull request in the next days i Run into the Same Problem 😁

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #1247, opened 2024-01-07
bug: accept use with addPlugin only

gimler:
i have a multi config setup. In one of my configs i only call `addPlugin()`.

```js
const path = require('node:path');
const fileName = process.env.COPY_FILE_NAME || false;
let copyConfig = {};
if (fileName) {
  copyConfig = {
    from: `${fileName}`,
    to: `${path.dirname(path.relative(path.resolve(__dirname, 'picture'), fileName))}/[name].[contenthash:8][ext]`,
  }
} else {
  copyConfig = {
    from: './picture',
    to: '[path][name].[contenthash:8][ext]'
  }

  Encore.cleanupOutputBeforeBuild()
};

const CopyPlugin = require('copy-webpack-plugin');
Encore
...
  .addPlugin(
      new CopyPlugin({
        patterns: [copyConfig],
      })
    )
...
```

this config only copy files optional only  a specific ;)

Actually only use `addPlugin()` is not supported. I change the code and add test for the scenario.

releated #1140

### DIFF
--- a/lib/config/validator.js
+++ b/lib/config/validator.js
@@ -43,8 +43,9 @@ class Validator {
         if (this.webpackConfig.entries.size === 0
             && this.webpackConfig.styleEntries.size === 0
             && this.webpackConfig.copyFilesConfigs.length === 0
+            && this.webpackConfig.plugins.length === 0
         ) {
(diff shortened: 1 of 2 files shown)


gimler:
@weaverryan can you take a look please?

weaverryan:
Thanks Gordon!

PR #1254, opened 2024-01-22
Bumping Node minimum to 20

weaverryan:
Node 16 and 18 have reached their end of life.

### DIFF
--- a/.github/workflows/high-depends.yml
+++ b/.github/workflows/high-depends.yml
@@ -14,7 +14,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest, windows-2019]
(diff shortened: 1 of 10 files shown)


nvdbeek:
hi @weaverryan, there is still security support for Node 18. Only the active support is stopped. See: https://endoflife.date/nodejs.

Is it necessary to stop using this LTS right now?

codegain:
@weaverryan Also running Encore with Node 18 which is still supported until ~April 2025: https://nodejs.org/en/about/previous-releases

weaverryan:
I'm sorry - I totally misread the info. I'll create a patch release to readd Node 18.

PR #1256, opened 2024-01-25
Re-adding node 18 support

weaverryan:
Removing Node 18 support was an accident - https://github.com/symfony/webpack-encore/pull/1254#issuecomment-1910589709 - so let's fix that :)


nvdbeek:
> Removing Node 18 support was an accident - [#1254 (comment)](https://github.com/symfony/webpack-encore/pull/1254#issuecomment-1910589709) - so let's fix that :)

Thanks @weaverryan !

PR #1258, opened 2024-02-01
Update NPM package dependencies

abozhinov:
Hello, 
I have dependency errors because using newest postcss, less, stylus loaders and for that we need to allow them in the package.json file. This PR will not break anything and it's pretty safe to merge it.

PR #1259, opened 2024-02-08
Update yarn used for test_apps to latest version

karpilin:
This updates yarn cjs files for testing_apps workflow and locks it to the current version which means we don't have to update on every yarn release


stof (approved):

karpilin:
@stof maybe we should lock the yarn version in test_apps workflow so we don't need to do this every time new version is released?

ie - `yarn set version 4.1.1` instead of `yarn set version berry` ?

karpilin:
Merging this PR will fix issues with dependabot updates #1261 #1262 #1263
@weaverryan

karpilin:
@stof @fabpot @nicolas-grekas? can we please merge this?
It will resolve all dependabot PR failing checks (#1261, #1262, #1263) as well as any other dependency updates like in #1258.

stof (approved):

PR #1310, opened 2024-08-30
Update the package.json to specify the yarn version for corepack

stof:
This makes it easier for contributors using corepack as their way of installing yarn (which is the recommended way nowadays)

### DIFF
--- a/.github/workflows/testing_apps.yml
+++ b/.github/workflows/testing_apps.yml
@@ -98,7 +98,7 @@ jobs:
           name: Install pnpm
           uses: pnpm/action-setup@v4
           with:
-            version: 8.x
+            package_json_file: '${{ matrix.app.working_directory }}/package.json'
 
       -   name: Packing Encore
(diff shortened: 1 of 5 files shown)


Kocal (approved):

Kocal:
We should add `packageManager` config for our test apps aswell

stof:
The packageManager in the main package.json breaks the CI for pnpm jobs because `pnpm/action-setup` detects conflicts between the `packageManager` and the specified version in the workflow. I need to investigate that. Maybe the solution is to change the workflow setup to tell pnpm/action-setup to look at the package.json of the test app instead, as that's where we use pnpm. I will investigate that.

> We should add `packageManager` config for our test apps aswell

Indeed. That's a good catch. I'll do it.

stof:
The yarn-based testing apps were already defining the `packageManager` field for corepack. I added it in all the `pnpm` testing apps, which makes it easier to have the same pnpm version locally than in CI (as the local setup does not care about the workflow definition).

PR #1311, opened 2024-08-30
Migrate to eslint-plugin-jsdoc

stof:
The deprecated valid-jsdoc rule of eslint does not support using typescript syntax for types, while it allows being more precise about types in a much more readable way than using the older jsdoc syntax with separate typedef or callback definitions.
The plugin also implements more rules than what valid-jsdoc does.

This prepares reviving https://github.com/symfony/webpack-encore/pull/818

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,8 +9,8 @@
 
 module.exports = {
     "root": true,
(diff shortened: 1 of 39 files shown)


Kocal (approved):
Thanks!

Issue #1312, opened 2024-08-30
[RFC] Migrate webpack-encore to ESM


stof:
More and more npm packages are shipping as ESM. In node, the compatibility between ESM and CommonJS has some limitation:

- an ES module can import a CommonJS module
- an ES module can dynamically import a CommonJS module
- a CommonJS module cannot import an ES module (Node 22 has an [experimental feature](https://nodejs.org/en/blog/announcements/v22-release-announce#support-requireing-synchronous-esm-graphs) allowing it for synchronous object graphs, i.e. ES modules that don't use top-level await)
- a CommonJS module can dynamically import an ES module through `await import(...)` (but this can only be used in an async context)

Shipping Encore as a CommonJS module means that we cannot upgrade our dependencies to versions that have migrated to ESM (as `Encore.getConfig()` is a synchronous API, any solution requiring an async context is a no-go).

Migrating to ESM needs to be done in a new major version as it is a BC break for downstream projects. This will require writing the `webpack.config.js` file in ESM, either by using `type=module` in the package.json or by using `webpack.config.mjs` instead (which is supported by webpack)

Kocal:
I'm 100% of favor of migrating Encore to ESM, it's something I've also thought recently.

> Shipping Encore as a CommonJS module means that we cannot upgrade our dependencies to versions that have migrated to ESM (as `Encore.getConfig()` is a synchronous API, any solution requiring an async context is a no-go).

Yeah, especially with #985 which forced us to introduce `rpc-sync` dependency to make ESLint API sync, and here in #1303 where I couldn't upgrade to an ESM version (and #1313 aswell).

Also, do we want to write JavaScript ESM compatible code, or do we want to add an additional build step like... TypeScript?

Kocal:
About TypeScript support from my previous message, of course it's a subject for another time, but in some recent versions of Node.js you can execute TypeScript code by stripping types at runtime, making it valid ECMAScript code. Of course, we should not use special TypeScript features like enum or decorators, that are not ECMAScript features.

sovetski:
Any news here? :/ The issue has almost 2 years now

stof:
@m10 and this problem is totally unrelated to the format used by Encore itself.

PR #1313, opened 2024-08-30
Replace clean-webpack-plugin with the core output cleaning feature

### LINKED ISSUE #1289: Remove clean-webpack-plugin
webpack has a builtin `output.clean` option that we should use instead of the unmaintained `clean-webpack-plugin`

stof:
Closes #1289

The options configurable by the configurator callback passed as second argument of `cleanupOutputBeforeBuild` are not the same anymore (the `dry` option used in the example still exists)


stof:
The ESLint job requires #1311 to pass, because the deprecated `valid-jsdoc` rule does not support the `import()` syntax in types.

package.json
```diff
@@ -31,11 +31,11 @@
     "assets-webpack-plugin": "7.0.*",
     "babel-loader": "^9.1.3",
     "chalk": "^4.0.0",
-    "clean-webpack-plugin": "^4.0.0",
     "css-loader": "^6.7.0",
     "css-minimizer-webpack-plugin": "^5.0.0",
     "fastest-levenshtein": "^1.0.16",
     "mini-css-extract-plugin": "^2.6.0",
+    "multimatch": "^5",
```
stof, line 38: The latest multimatch version is 7.0. but as of 6.0, this package is shipped as ESM which makes it unusable in webpack-encore (see #1312)

Kocal, line 38: Can we use [picomatch](https://www.npmjs.com/package/picomatch) instead?

It has 0 dependencies (where `multimatch` has 7 dependencies), and weight much less, see graphs:
- picomatch: https://npmgraph.js.org/?q=picomatch#select=exact%3Apicomatch%404.0.2
- mulitmatch: https://npmgraph.js.org/?q=multimatch%406.0.0

picomatch is also compatible with CJS modules.

Thanks!

Kocal (commented):
Thank you for working on this!

package.json
stof: and picomatch might switch to ESM in the future as well (there is a PR opened about that). So just using picomatch won't let us use CJS forever (my expectation is that more and more packages will migrate to ESM)

stof: note that this still reduces our dependencies compared to `clean-webpack-plugin`

Kocal: What about breaking the cleaning feature and let the user decide what it wants to **keep** (through `output.clean.keep` if I remember)? 
We could directly pass its configuration to Webpack, it will be easier for us, no need for extra work nor dependencies.
WDYT ?

stof: We still need the config to keep the manifest.json for the dev-server (unless the issue mentioned in the code comment is not relevant anymore.

If we can remove that part, I'm fine with doing a bigger BC break dropping the first argument (and letting users configure `keep` through the callback if needed).

stof: However, as I'm not using the dev-server myself (I'm using the watch mode), I don't have a setup to test this.

Kocal: I can give it a try tomorrow

Kocal: Good news, the option `output.clean.keep()` seems totally ignored in dev-server:
- Just to be sure, I've removed by `public/build` folder, 
- Commented `cleanedPaths.push('!manifest.json')`
- Ran the dev-server

And file `public/build/manifest.json` is present, with good URLs, HMR works too, etc... :)

So we can remove `cleanedPaths.push('!manifest.json')` and go for a bigger BC :)

stof: I updated the PR with the simpler implementation.

Kocal:
Thanks for your modifications! 

I've tried your branch locally and everything worked fine, as expected I had to update `.cleanupOutputBeforeBuild(['**/*', '!amp/**'])` to `.cleanupOutputBeforeBuild()` (because of argument check, which works fine). 

We can merge after a rebase IMO

Kocal (approved):

PR #1315, opened 2024-08-30
Remove useless eslint disabling comments

stof:
Those lines don't report such error.

### DIFF
--- a/lib/loaders/css.js
+++ b/lib/loaders/css.js
@@ -57,7 +57,7 @@ module.exports = {
             };
 
             cssLoaders.push({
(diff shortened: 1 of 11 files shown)


stof:
Should I addthe `--report-unused-disable-directives` flag in the `lint` script in our package.json ?

Kocal:
> Should I add the `--report-unused-disable-directives` flag i

Yes please, it would help to remove useless directives.

Kocal:
Thanks @stof, I've added the flag to the `lint` npm script.

PR #1318, opened 2024-08-31
Drop support of webpack-dev-server ^4.8, add support for webpack-dev-server ^5.0

### LINKED ISSUE #1290: Migrate to webpack-dev-server 5
webpack-dev-server has a new major version with a bunch of breaking changes: https://github.com/webpack/webpack-dev-server/blob/master/migration-v5.md

 As we allow projects to customize their settings in webpack-encore, upgrading to the new major version requires a BC break. For simplicity, I suggest that webpack-encore 5.x uses only the new major version of webpack-dev-server instead of trying to support multiple versions (for dependencies rather than peer dependencies, all of yarn, npm and pnpm will install the latest version in the range anyway. Only `yarn --flat` or equivalent non-default resolution modes would benefit from a support of older versions, and I don't think there is a valid set of dependencies for webpack-encore for the flat resolution anyway)

Kocal:
Close #1290 

Upgrade webpack-dev-server didn't have impact for Encore itself, but it can be the case for end-users.

After upgrading webpack-dev-server to ^5.0, my project that use the following devServer configuration is still working as expected:
```js
Encore.configureDevServerOptions((options) => {
    options.allowedHosts = 'all';
    options.client = {
        overlay: false, // hidden overlay: multiple errors caused by third-party scripts
    };
    options.server = {
        type: 'https',
        options: {
            pfx: path.join(process.env.HOME, '.symfony5/certs/default.p12'),
        },
    };
})
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,39 @@ Encore.enableSassLoader((options) => {
 
 * #1324 Drop css-minimizer-webpack-plugin 5 support, only css-minimizer-webpack-plugin 7 is supported (@Kocal)
 
+* #1318 Drop webpack-dev-server 4 support, only webpack-dev-server 5 is supported (@Kocal)
+
+The dev-server options have changed between versions 4 and 5, see [the official migration guide to v5](https://github.com/webpack/webpack-dev-server/blob/master/migration-v5.md).
+For example:
+```js
+// With webpack-dev-server 4:
+Encore.configureDevServerOptions((options) => {
+   options.https = {
+     ca: "./path/to/server.pem",
+     pfx: "./path/to/server.pfx",
+     key: "./path/to/server.key",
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This is a new major version that contains several backwards-compatibility breaks
 
 * #1309 Drop ESLint integration (@Kocal) 
 
+* #1318 Drop webpack-dev-server 4 support, only webpack-dev-server 5 is supported (@Kocal)
```
stof, line 15: I suggest mentioning that the supported option in the configuration callback have changed, with a link to their changelog. They have BC breaks in their configuration.

Kocal: Done

stof (approved):

PR #1320, opened 2024-09-01
Drop css-minimizer-webpack-plugin 5 support, only css-minimizer-webpack-plugin 7 is supported

Kocal:
It also upgrade cssnano to ^7, see https://github.com/webpack-contrib/css-minimizer-webpack-plugin/blob/master/CHANGELOG.md#700-2024-05-07

### DIFF
--- a/.github/workflows/high-depends.yml
+++ b/.github/workflows/high-depends.yml
@@ -14,7 +14,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest, windows-2019]
-                node-versions: ['18', '20', '21', '22']
+                node-versions: ['18', '20', '22']
 
         steps:
             -   name: Checkout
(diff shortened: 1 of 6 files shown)


Kocal:
Failing checks due to Node.21 will be fixed by #1321, more work is needed for other checks.

stof:
This requires bumping the min node version from 18.0.0 to 18.12.0 due to their BC break in 6.0

stof:
The failures are probably not directly related to this PR. https://github.com/symfony/webpack-encore/pull/818 fails the "low dependencies" jobs while it only changes comments.

stof:
I confirm that a build in the master branch also breaks there: https://github.com/symfony/webpack-encore/actions/runs/10666578054/job/29562308888

And #1322 seems to fix it.

PR #1325, opened 2024-09-03
Remove no-op argument

stof:
The package.json check forbidding to add webpack as a dependency was removed in webpack-encore 4.0 when we made it mandatory to depend on webpack explicitly.


stof:
I found this thanks to the typescript type checker (as my work on #818 involved running it to validate my types)

Kocal (approved):

PR #1327, opened 2024-09-04
Fix compat with configuring the devserver server as a string

stof:
webpack-dev-server supports providing a string when only the type needs to be configured. Our code checking the type to detect the usage of `https` was not properly handling such configuration.


stof:
This has been detected by the typescript type checker as it now knows the type of `devServerConfig` thanks to the work done in #818

Kocal (approved):

PR #1330, opened 2024-09-05
Add some types in the internal implementation of Encore

stof:
This will allow type-checking the codebase, by fixing cases where the typescript inference is not sufficient without a hint.

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -82,6 +82,9 @@ class ConfigGenerator {
             this.webpackConfig.runtimeConfig.devServerFinalIsHttps = false;
         }
 
+        /**
+         * @type {import('webpack').Configuration}
+         */
         const config = {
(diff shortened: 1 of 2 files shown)


lib/config-generator.js
```diff
@@ -558,7 +564,7 @@ class ConfigGenerator {
     buildCacheConfig() {
         const cache = {};
 
-        cache.type = 'filesystem';
+        cache.type = /** @type {const} */ ('filesystem');
```
stof, line 567: this syntax with `/** @type */` followed by a parenthesized expression as the JSDoc syntax for a cast.
Using `const` as type allows to do the same than [`as const`](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-inference) in typescript

Kocal (approved):

PR #1331, opened 2024-09-05
test(deps): replace Zombie by Puppeteer

Kocal:
https://github.com/assaf/zombie is archived since december 2023, and its last release is from 4 years ago.

I've replaced it by https://github.com/puppeteer/puppeteer, a modern and maintained API to run and test code in a **real browser** (zombie used `jsdom`). I had to adapt `requestTestPage` to pass the browser as first argument, because I want to launch the browser **only a single time** (in `before()`).

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -5,7 +5,10 @@ jobs:
     eslint:
         name: ESLint
         runs-on: ubuntu-latest
-
+        env:
+            # We don't want Puppeteer to automatically download a browser when dependencies are being installed
+            PUPPETEER_SKIP_DOWNLOAD: 'true'
+            
         steps:
             -   name: Checkout
                 uses: actions/checkout@v4
--- a/test/helpers/setup.js
+++ b/test/helpers/setup.js
@@ -14,7 +14,6 @@ const WebpackConfig = require('../../lib/WebpackConfig');
 const parseRuntime = require('../../lib/config/parse-runtime');
 const webpack = require('webpack');
 const fs = require('fs-extra');
-const Browser = require('zombie');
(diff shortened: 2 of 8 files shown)


.github/workflows/lint.yaml
```diff
@@ -5,7 +5,10 @@ jobs:
     eslint:
         name: ESLint
         runs-on: ubuntu-latest
-
+        env:
+            # We don't want Puppeteer to automatically download a browser when dependencies are being installed
+            PUPPETEER_SKIP_DOWNLOAD: 'false'
```
stof, line 10: `skip_download: false` makes it _not_ download the browser ? Are you sure about that ?

test/helpers/setup.js
```diff
@@ -142,18 +141,22 @@ function stopAllServers() {
...
  *
+ * @param {import('puppeteer').Browser} browser Puppeteer browser instance
  * @param {string} webRootDir Directory path (e.g. /path/to/public) where the web server should be rooted
  * @param {Array} scriptSrcs  Used to create <script src=""> tags.
- * @param {Function} callback Called after the page was requested.
+ * @param {function({
+ *      page: import('puppeteer').Page,
+ *      loadedResources: Array<{ response: import('puppeteer').HTTPResponse }>}
+ * ): void} callback Called after the page was requested.
  * @returns {void}
```
stof, line 151: This actually returns `Promise<void>` now

.github/workflows/lint.yaml
Kocal: Ouch, the fatigue got me, thanks!

stof (approved):
You need to rebase to fix conflict on the yarn.lock file, but otherwise, things look fine.

PR #1333, opened 2024-09-07
Replace Chalk by Picocolors

Kocal:
[Picocolors](https://www.npmjs.com/package/picocolors) is one of the smaller library to output colors in the terminal, and it has no sub-dependencies. It can reduce the Encore package size a lot.

```
Package size report
===================

Package info for "chalk@4.1.2": 100 kB
  Released: 2021-07-30 12:02:52.839 +0000 UTC (162w1d ago)
  Downloads last week: 233,169,927 (56.55%)
  Estimated traffic last week: 23 TB
  Subdependencies: 5
  Latest version: 5.3.0 (62w2d ago)

Added dependencies:
  + picocolors@1.1.0: 12 kB (11.81%)
    Downloads last week: 8,949,572 (N/A% from 1.1.0)
    Estimated traffic last week: N/A
    Subdependencies: 0 (0%)

Estimated new statistics:
  Package size: 100 kB → 76 kB (76.16%)
  Subdependencies: 5 → 6 (1)
  Traffic with last week's downloads:
    For current version: 23 TB → 18 TB (5.6 TB saved)
    For all versions: 41 TB → 31 TB (9.8 TB saved)
```

Migrating from chalk to picocolors is pretty simple, the API is nearly identical except whan you want to color both text and background.

When running `encore --help`:
image

When running `encore dev`: 
image


stof:
This does not actually reduce dependencies, because `chalk` is still a transitive dependency of `css-minimizer-webpack-plugin`. So this PR doesn't actually change the size of downloaded dependencies (`picocolors` was already a transitive dependency, from `css-minimizer-webpack-plugin` > `cssnano` as well as other places)

stof (approved):

Kocal:
In reality yes, but I believe someone (or me) will open this kind of PR on `css-minimizer-webpack-plugin` etc, and only at this time we will be able to reduce the package size

stof:
Looking at the npm graph, the only non-dev transitive dependencies we have are those:

- `@nuxt/friendly-errors-webpack-plugin` requires chalk 2.x
- `jest-util` and `@jest/types` require chalk 4.x. These are required by `jest-worker` which is required by `css-minimizer-webpack-plugin`

Kocal:
- PR opened for `@nuxt/friendly-errors-webpack-plugin`: https://github.com/nuxt/friendly-errors-webpack-plugin/pull/21
- PR already exists for Jest: https://github.com/jestjs/jest/pull/15197

PR #1334, opened 2024-09-07
Upgrade the version of stylus used in dev

stof:
This also removes the optional peerDependency on stylus that was using an outdated range because Encore does not actually use stylus in any way but only stylus-loader. And stylus-loader already declares a required peerDependency on stylus, making package managers ensure its presence when stylus-loader is present.


Kocal (approved):
Looks like legit, thanks!

PR #1335, opened 2024-09-07
Fix ESLint warning

stof:
This configures our CI to enforce fixing ESLint warnings and not just errors.
The only warning that was reported (since #1331) is solved by simplifying the code (removing the function that was keeping a closure on the `browser` mutable variable)


Kocal (approved):
Thanks!

PR #1336, opened 2024-09-08
Make webpack-dev-server optional

Kocal:
As discussed with @stof, we want to make the webpack-dev-server an optional peer dependency:

JavaScript dependencies are problematic, depending on many sub-dependencies, which in turn depend on more sub-dependencies, and so on... Welcome to the dependency hell!

Even if the dev-server functionality isn't used, the dependency tree is immensely more complex (over 300 additional dependencies), but this is an open door to security holes present in “discrete” (say “little-known”), but over-used dependencies.

In recent months, a sort of “witch-hunt” has been set up by some people in the JavaScript ecosystem, to replace sub-dependencies with lighter alternatives (either another dependency, or a native version) in _popular project_. I've started doing this on Encore for a few dependencies, and making the webpack-dev-server optional is a big win.

```
Package size report
===================

Package info for "@symfony/webpack-encore@4.7.0": 61 MB
  Released: 2024-08-29 16:26:01.762 +0000 UTC (1w3d ago)
  Downloads last week: 19,232 (11.57%)
  Estimated traffic last week: 1.2 TB
  Subdependencies: 626

Removed dependencies:
  - webpack-dev-server@4.15.2: 29 MB (47.60%)
    Downloads last week: 2,605,016 (N/A% from 4.15.2)
    Downloads last week from "@symfony/webpack-encore@4.7.0": 19,232 (N/A%)
    Traffic last week: N/A
    Traffic from "@symfony/webpack-encore@4.7.0": 1.2 TB (N/A%)
    Subdependencies: 283 (45.20%)

Estimated new statistics:
  Package size: 61 MB → 43 MB (69.83%)
  Subdependencies: 626 → 295 (-331)
  Traffic with last week's downloads:
    For current version: 1.2 TB → 823 GB (355 GB saved)
    For all versions: 10 TB → 7.1 TB (3.1 TB saved)
```

When upgrading Encore to v5, end-users will have to install the `webpack-dev-server` back to use it again.

### DIFF
--- a/bin/encore.js
+++ b/bin/encore.js
@@ -14,6 +14,7 @@ const parseRuntime = require('../lib/config/parse-runtime');
 const context = require('../lib/context');
 const pc = require('picocolors');
 const logger = require('../lib/logger');
+const featuresHelper = require("../lib/features");
 
 const runtimeConfig = parseRuntime(
     require('yargs-parser')(process.argv.slice(2)),
@@ -56,6 +57,13 @@ if (runtimeConfig.helpRequested) {
 }
 
 if (runtimeConfig.useDevServer) {
+    try {
+        featuresHelper.ensurePackagesExistAndAreCorrectVersion('webpack-dev-server', 'the webpack Development Server');
+    } catch (e) {
+        console.log(e);
+        process.exit(1);
+    }
+
     console.log('Running webpack-dev-server ...');
     console.log();
 
--- a/index.js
+++ b/index.js
@@ -726,7 +726,7 @@ class Encore {
      * });
      * ```
      *
-     * @param {OptionsCallback<import('webpack-dev-server').Configuration>} callback
+     * @param {OptionsCallback<object>} callback
      * @returns {Encore}
      */
     configureDevServerOptions(callback) {
--- a/lib/features.js
+++ b/lib/features.js
@@ -14,6 +14,12 @@ const packageHelper = require('./package-helper');
 /**
  * An object that holds internal configuration about different
  * "loaders"/"plugins" that can be enabled/used.
+ *
+ * @type {{[key: string]: {
+ *     method: string,
+ *     packages: Array<{ name: string, enforce_version?: boolean } | Array<{ name: string }>>,
+ *     description: string,
+ * }}}
  */
 const features = {
     sass: {
@@ -146,7 +152,14 @@ const features = {
             { name: 'svelte-loader', enforce_version: true }
         ],
         description: 'process Svelte JS files'
-    }
+    },
+    'webpack-dev-server': {
+        method: 'configureDevServerOptions()',
+        packages: [
+            { name: 'webpack-dev-server' }
+        ],
+        description: 'run the Webpack development server'
+    },
 };
 
 function getFeatureConfig(featureName) {
@@ -158,12 +171,12 @@ function getFeatureConfig(featureName) {
 }
 
 module.exports = {
-    ensurePackagesExistAndAreCorrectVersion: function(featureName) {
+    ensurePackagesExistAndAreCorrectVersion: function(featureName, method = null) {
         const config = getFeatureConfig(featureName);
 
         packageHelper.ensurePackagesExist(
             packageHelper.addPackagesVersionConstraint(config.packages),
-            config.method
+            method || config.method
         );
     },
 
--- a/test/bin/encore.js
+++ b/test/bin/encore.js
@@ -15,7 +15,9 @@ const expect = chai.expect;
 const path = require('path');
 const testSetup = require('../helpers/setup');
 const fs = require('fs-extra');
-const exec = require('child_process').exec;
+const { exec, execSync, spawn } = require('child_process');
+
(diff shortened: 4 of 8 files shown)


bin/encore.js
```diff
@@ -56,6 +57,13 @@ if (runtimeConfig.helpRequested) {
 }
 
 if (runtimeConfig.useDevServer) {
+    try {
+        featuresHelper.ensurePackagesExistAndAreCorrectVersion('webpack-dev-server', 'the webpack Development Server');
+    } catch (e) {
+        console.log(e);
+        process.exit(1);
+    }
```
Kocal, line 65: I could have used `isPackageInstalled()`, but I would have to rewrite a lot of logic here to enforce the version, generate the good command, etc... 

Adding `method` parameter to `featuresHelper.ensurePackagesExistAndAreCorrectVersion` was a lot easier and _clean_.

lib/features.js
```diff
@@ -14,6 +14,12 @@ const packageHelper = require('./package-helper');
 /**
  * An object that holds internal configuration about different
  * "loaders"/"plugins" that can be enabled/used.
+ *
+ * @type {{[key: string]: {
+ *     method: string,
+ *     packages: Array<{ name: string, version?: string } | Array<{ name: string }>>,
```
stof, line 20: this does not make sense. The proper type is `Array<{ name: string, enforce_version?: boolean }>`

index.js
```diff
@@ -730,6 +731,8 @@ class Encore {
      * @returns {Encore}
      */
     configureDevServerOptions(callback) {
+        featuresHelper.ensurePackagesExistAndAreCorrectVersion('webpack-dev-server');
```
stof, line 734: Shouldn't this be done in WebpackConfig or config-generator like the other similar checks ?

test/bin/encore.js
```diff
@@ -213,4 +215,128 @@ module.exports = Encore.getWebpackConfig();
...
+            throw new Error('Missing "webpack-dev-server" as dev dependency in package.json.');
+        }
+
+        before(() => {
+            execSync('yarn remove webpack-dev-server --dev', { cwd: projectDir });
+        });
+
+        after(() => {
+            // Re-install webpack-dev-server and ensure the project is in a clean state
+            execSync(`yarn add webpack-dev-server@${webpackDevServerConstraint} --dev`, { cwd: projectDir });
```
stof, line 295: why not doing a `git checkout package.json` instead ?

stof (commented):
types relying on `import('webpack-dev-server').Configuration` need to be changed to use `object` instead, as our type declarations cannot depend on optional dependencies

lib/features.js
Kocal: See https://github.com/symfony/webpack-encore/blob/main/lib/features.js#L23, `Array<{ name: string }>` is like "install one of those dependencies"

index.js
Kocal: You're right

test/bin/encore.js
Kocal: I think about it, but I don't wanted to checkout unrelated modifications by accident

Kocal:
> types relying on `import('webpack-dev-server').Configuration` need to be changed to use `object` instead, as our type declarations cannot depend on optional dependencies

You think so? I kept `webpack-dev-server` as a dev dependency, and it looks like fine

lib/features.js
stof: an, I misread the nesting.

Still, the `version` key does not make sense as the optional key we have is `enforce_version`

test/bin/encore.js
stof: wouldn't this also apply to the lock file then ?

Is there a way to write such tests without changing the dev dependencies of encore ?

stof:
@Kocal as long as we use types for ourselves, the dev dependency is indeed enough. But this would block #816 which was the primary reason to add more precise types. Types shipped in the package cannot use dev dependencies.

Kocal:
Ah, I didn't think about that, I will change that

test/bin/encore.js
Kocal: > wouldn't this also apply to the lock file then ?

I'm not sure to understand, you mean `git checkout yarn.lock`? If yes, it won't help to restore remove/re-add `webpack-dev-server` dependency in `package.json`

> Is there a way to write such tests without changing the dev dependencies of encore ?

I first thought about writing E2E tests like the one for our "test apps":
- clone Encore
- install prod dependencies and pack Encore, ensure webpack-dev-server is not present
- create a new project (or reuse a test-app)
- write a bash assertion for `encore dev-server` 

But it's not really the goal of our actual "test apps", I also wanted to check lowest/fixed/highest dependencies, and I prefer to write tests in `test/*.js` files.

Kocal: I've changed to a `git checkout package.json yarn.lock` instead, it's more simple, and too bad if it cancels unrelated modifications.

Kocal:
Checks are finally green,

PR #1341, opened 2024-09-16
Upgrade assets-webpack-plugin to ^7.1, leverage integrity calculations to assets-webpack-plugin and webpack-subresource-integrity

### LINKED ISSUE #969: Entrypoints integrity workaround needed to allow for assets-webpack-plugin or higher
See #968 

ztoben/assets-webpack-plugin#392 makes our integrity calculation no longer work, as the files have not been emitted by the time `processOutput()` is called.

The most obvious fix would be to - if possible - use the existing https://github.com/ztoben/assets-webpack-plugin#integrity

If it IS possible (I don't know if it exposes all the same info we do), then we would need to deprecate the current format of the entrypoints.json file in favor of this new one.

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | Fix #969 
| License       | MIT

Proposal for #969, but the PR is not ready yet:
1. I want https://github.com/ztoben/assets-webpack-plugin/pull/452 to be merged/released, because the version 1.x has been unmaintained 4 years ago
2. The way that `assets-webpack-plugin` add integrity hashes to the `entrypoints.json` is different that our actual format, it's a breaking change (that will impact Encore and WebpackEncoreBundle users), but maybe it can be avoided with `processOutput` 
3. I still need to add/update tests

### DIFF
--- a/lib/config-generator.js
+++ b/lib/config-generator.js
@@ -262,7 +262,8 @@ class ConfigGenerator {
             // will use the CDN path (if one is available) so that split
             // chunks load internally through the CDN.
             publicPath: this.webpackConfig.getRealPublicPath(),
-            pathinfo: !this.webpackConfig.isProduction()
+            pathinfo: !this.webpackConfig.isProduction(),
+            crossOriginLoading: this.webpackConfig.useIntegrity ? 'anonymous' : false,
         };
     }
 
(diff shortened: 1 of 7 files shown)


lib/config-generator.js
```diff
@@ -558,6 +585,10 @@ class ConfigGenerator {
             splitChunks
         );
 
+        if (this.webpackConfig.useIntegrity > 0) {
```
stof, line 588: why `> 0` while it is a boolean ?

package.json
```diff
@@ -216,6 +217,9 @@
     },
     "webpack-notifier": {
       "optional": true
+    },
+    "webpack-subresource-integrity": {
+        "optional": true
```
stof, line 222: listing it here without putting it in `peerDependencies` does not make sense.

lib/config-generator.js
Kocal: Bad search/replace, before introducing `webpackConfig.useIntegrity`, I've used `webpackConfig.integrityAlgorithms.length > 0` to check if integrity was enabled.

Kocal:
Closing in favor of a new PR, which will drop `assets-webpack-plugin` and create the `entrypoints.json` file from an internal plugin.

Issue #1269, opened 2024-05-23
Integrity hashes do not work if contenthash is a query string


rvock:
If you are using the `[contenthash]` as a query string, the integrity hashes are not generated.

## Example webpack.config.js

```js
const Encore = require('@symfony/webpack-encore');

Encore
	// directory where compiled assets will be stored
	.setOutputPath('build/')
	// public path used by the web server to access the output path
	.setPublicPath('/build/')
	.setManifestKeyPrefix('build/')

	// hash as querystring instead of in filename
	.configureFilenames({
		js: '[name].js?[contenthash]',
	})

	.addEntry('example', './src/example.js')

	.disableSingleRuntimeChunk()

	.enableVersioning(Encore.isProduction())

	.enableIntegrityHashes(Encore.isProduction())
;

module.exports = Encore.getWebpackConfig();
```

## Generated entrypoints.json

```json
{
  "entrypoints": {
    "example": {
      "js": [
        "/static/build/example.js?621df54d056c27d5eaa8"
      ]
    }
  },
  "integrity": {}
}
```

The problem exists, because `entry-files-manifest.js` assumes the filename is a a valid file:
https://github.com/symfony/webpack-encore/blob/6b75c4dd963c613ddc19fa2748e0b6bcc4512c16/lib/plugins/entry-files-manifest.js#L52-L57

If the query string would be stripped, the file path should be correct.

PR #1342, opened 2024-09-17
Replace `assets-webpack-plugin` dependency by an internal plugin, to generate `entrypoints.json` file

### LINKED ISSUE #969: Entrypoints integrity workaround needed to allow for assets-webpack-plugin or higher

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | Fix #969 
| License       | MIT

This PR replaces the https://github.com/ztoben/assets-webpack-plugin dependency with an internal plugin, this allows us:
- to drop a dependency, and simplify the code (we don't need 10 ~options to generate the `entrypoints.json` for the WebpackEncoreBundle)
- to not depend on https://github.com/waysact/webpack-subresource-integrity, which is necessary to use the `integrity` options of https://github.com/ztoben/assets-webpack-plugin
- to generate the integrity data **after** files have been emitted

**Things done in this PR:**
- Re-implement https://github.com/ztoben/assets-webpack-plugin functionality as closely as possible
- Ensure tests pass without any modifications
- Ensure the branch works like before on a real project (in dev/prod/dev-server, with or without integrity hashes)

**Things not done in this PR:**
- https://github.com/symfony/webpack-encore/issues/1269 will have to wait for another PR, I don't wanted to fix that here

### DIFF
--- a/lib/webpack/entry-points-plugin.js
+++ b/lib/webpack/entry-points-plugin.js
@@ -1,0 +1,136 @@
+/*
+ * This file is part of the Symfony Webpack Encore package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const crypto = require('crypto');
+const copyEntryTmpName = require('../utils/copyEntryTmpName');
+
+/**
+ * Return the file extension from a filename, without the leading dot and without the query string (if any).
+ *
+ * @param {string} filename
+ * @returns {string}
+ */
+function getFileExtension(filename) {
+    return path.extname(filename).slice(1).split('?')[0];
+}
+
+class EntryPointsPlugin {
+    /**
+     * @param {object} options
+     * @param {string} options.publicPath The public path of the assets, from where they are served
+     * @param {string} options.outputPath The output path of the assets, from where they are saved
+     * @param {Array<string>} options.integrityAlgorithms The algorithms to use for the integrity hash
+     */
+    constructor({
+        publicPath,
+        outputPath,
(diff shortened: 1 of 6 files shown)


lib/webpack/entry-points-plugin.js
```diff
@@ -0,0 +1,132 @@
...
+                    continue;
+                }
+
+                manifest.entrypoints[entryName] = {};
+
+                for (const asset of entry.assets) {
+                    // We don't want to include hot-update files in the manifest
+                    if (asset.name.includes('.hot-update.')) {
+                        continue;
+                    }
```
Kocal, line 73: When using the dev-server and you update a file, an `.hot-update.` file can be emitted, but we don't want to add it in the manifest.

lib/webpack/entry-points-plugin.js
```diff
@@ -0,0 +1,133 @@
...
+        publicPath,
+        outputPath,
+        integrityAlgorithms
+    }) {
+        this.publicPath = publicPath;
+        this.outputPath = outputPath;
+        this.integrityAlgorithms = integrityAlgorithms;
+    }
+
+    apply(compiler) {
```
stof, line 44: I suggest adding `@param {import('webpack').Compiler} compiler`

Kocal: Done, thanks

stof (approved):

PR #1343, opened 2024-09-20
Add tests for CSS Modules with React and Preact

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #... 
| License       | MIT

Quick PR to add functional tests about using CSS Modules with React and Preact, to see it it also break CSS Modules with them in #1319.


lib/loaders/babel.js
```diff
@@ -70,7 +70,10 @@ module.exports = {
             if (webpackConfig.useReact) {
                 loaderFeatures.ensurePackagesExistAndAreCorrectVersion('react');
 
-                babelConfig.presets.push(require.resolve('@babel/preset-react'));
+                babelConfig.presets.push([require.resolve('@babel/preset-react'), {
+                    // TODO: To remove when Babel 8, "automatic" will become the default value
+                    runtime: 'automatic',
+                }]);
```
Kocal, line 76: Quick-win, let's bump `@babel/preset-react` to ^7.9.0 to use [`runtime: 'automatic'`](https://babeljs.io/docs/babel-preset-react#runtime)

package.json
```diff
@@ -48,7 +48,7 @@
     "@babel/eslint-parser": "^7.17.0",
     "@babel/plugin-transform-react-jsx": "^7.12.11",
     "@babel/preset-env": "^7.16.0",
-    "@babel/preset-react": "^7.0.0",
+    "@babel/preset-react": "^7.9.0",
```
stof, line 51: bumping the dev dependency alone is not enough. For our users, what matters for the package managers is the `peerDependencies` section.

btw, we should update our `features` utility to make it enforce versions based on `peerDependencies` rather than `devDependencies` now that we properly define peer dependencies. Or we should drop that runtime check entirely in favor of letting yarn/npm/pnpm show the warning at installation time.

Kocal: Oh yes, forgot about peerDeps, thanks

Kocal: > btw, we should update our `features` utility to make it enforce versions based on `peerDependencies` rather than `devDependencies` now that we properly define peer dependencies. Or we should drop that runtime check entirely in favor of letting yarn/npm/pnpm show the warning at installation time.

Totally! Maybe we can do 1) for v5, and 2) for v6?

stof: the first change can even be done in a minor version (as that's about changing the internal implementation only). Our peerDependencies is already what package managers will report.

PR #1366, opened 2025-02-23
Migrate from Mocha/Sinon/Chai to Vitest

### LINKED ISSUE #1364: Migrate from Mocha/Sinon/Chai to Vitest
Purely an internal subject, but I would like to modernize our tests and use 1 single tool instead of 3 different tools (less frictions, easier to update, ...). 

Since Vitest is based on Vite, it may helps in the future:
- to ship an ESM version (like proposed in #1312) without dropping the CJS version (which won't requires a new major version
- to rewrite the code base in TypeScript

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | yes/no 
| Issues        | Fix #1364 
| License       | MIT


Kocal:
Due to Webpack Encore current and previous state, I don't think all this work is necessary anymore. The v6 will surely never exists! 👼🏻

PR #1368, opened 2025-04-04
Remove warning for usage of configureLoaderRule

### LINKED ISSUE #886: Add method to disable warning about Encore.configureLoaderRule()
When adding svgr we need to alter the image rule.
We could either disable the image rule all together, or the less fragile option is to just remove svg from the test.

```js
Encore.configureLoaderRule("images", (loaderRule) => {
  loaderRule.test = new RegExp(loaderRule.test.toString().replace('|svg', ''));
});
```

This triggers a warning 

> Be careful when using Encore.configureLoaderRule(), this is a low-level method that can potentially break Encore and Webpack when not used carefully.

I'd like a way to disable that warning, an explicit opt-in to say "I know what I'm doing can break it, but please don't want me every time".

Something like

`Encore.disableLowLevelApiWarnings();`

Tofandel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #886  
| License       | MIT

### DIFF
--- a/lib/WebpackConfig.js
+++ b/lib/WebpackConfig.js
@@ -1033,8 +1033,6 @@ class WebpackConfig {
      * @param {OptionsCallback<webpack.RuleSetRule>} callback
      */
     configureLoaderRule(name, callback) {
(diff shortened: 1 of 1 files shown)


Kocal:
Thanks!

PR #1383, opened 2025-11-13
Pin `tmp` dependency version to `^0.2.5` (CVE)

orbeji:
Bumps [tmp](https://github.com/raszi/node-tmp) to 0.2.5


orbeji:
Hi,
Sorry, for the messed update I reverted the changes and only updated the tmp package.
I did it to avoid installing the 0.2.3 version and below which have some [CVE](https://nvd.nist.gov/vuln/detail/CVE-2025-54798)
Thanks

PR #1398, opened 2026-03-11
Remove unmaintained file-loader dependency

### LINKED ISSUE #1396: Migrate from (deprecated) file-loader
https://v4.webpack.js.org/loaders/file-loader/ is deprecated in favor of https://webpack.js.org/guides/asset-modules/

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | Fix #1396 
| License       | MIT

This PR removes the deprecated file-loader dependency in favor of webpack's Asset Modules.
This loader was previously used in `Encore.copyFiles()`.

### DIFF
--- a/lib/webpack/copy-files-loader.js
+++ b/lib/webpack/copy-files-loader.js
@@ -9,39 +9,60 @@
 
 'use strict';
 
-const LoaderDependency = require('webpack/lib/dependencies/LoaderDependency');
 const path = require('path');
 
 module.exports.raw = true; // Needed to avoid corrupted binary files
 
-module.exports.default = function loader(source) {
-    // This is a hack that allows `Encore.copyFiles()` to support
-    // JSON files using the file-loader (which is not something
-    // that is supported in Webpack 4, see https://github.com/symfony/webpack-encore/issues/535)
-    //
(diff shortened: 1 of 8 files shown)


lib/webpack/copy-files-loader.js
```diff
@@ -59,33 +80,41 @@ module.exports.default = function loader(source) {
...
+    // Check if filename contains a hash (for immutability hints)
+    if (/\[(?:content)?hash(?::\d+)?\]/i.test(options.filename)) {
+        assetInfo.immutable = true;
+    }
+
+    // Emit the file to the output directory
+    this.emitFile(outputPath, source, null, assetInfo);
+
+    // Return a module that exports the public URL
+    return `module.exports = __webpack_public_path__ + ${JSON.stringify(outputPath)};`;
```
stof, line 118: exporting CJS will prevent some optimizations (merging modules for instance). Wouldn't it be possible to use ESM instead ?

Kocal: I think it could, let me work on this

PR #1402, opened 2026-03-14
Migrate to ESM & async API

### LINKED ISSUE #1312: [RFC] Migrate webpack-encore to ESM

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | Fix #1312 
| License       | MIT

This PR has been generated with Claude Opus 4.6.

I didn't review every files yet, only some of them (which are now fixed) and the CHANGELOG.

I'm keeping the PR in draft for the moment


Kocal:
Technically this PR is legit ready to be merged, but I still need to deep-review:
- lib/config-generator.js
- lib/WebpackConfig.js

I will wait for few weeks when Encore 6.0 will [have more downloads](https://www.npmjs.com/package/@symfony/webpack-encore?activeTab=versions)

Kocal:
> I don't think we should use non-fully-specified imports as those are non-standard.

I'm not sure to understand, do you speak about https://webpack.js.org/configuration/module/#resolvefullyspecified? It's enabled by default in the user's `package.json` has `type: "module"`, and I would be in favor to keep the default setting.

Or are you speaking about Encore's internal? Because If I run `yarn mocha test/bin/encore.js` after removing `.js` from `await import('webpack/bin/webpack.js');` (in `bin/encore.js` file), the tests fail. But if I re-add this `.js`, tests pass again.

PR #1411, opened 2026-05-04
Add support for typescript 6

### LINKED ISSUE #1409: typescript 6
The peer dependency currently expects typescript 5. Typescript 6.0 has been released.

ts-loader works with typescript 6 (there are BC breaks impacting projects as typescript changed some config settings, but no BC breaks impacting the loader directly), so we could expand our peer dependency in a minor version of Encore.

stof:
| Q             | A                                                                                                                         |
| ------------- | ------------------------------------------------------------------------------------------------------------------------- |
| Bug fix?      | no                                                                                                                    |
| New feature?  | yes                                                                    |
| Deprecations? | no                                                                      |
| Issues        | Fix #1409 |
| License       | MIT                                                                                                                       |


Kocal (commented):
Some checks are failing:
> TS5095: Option 'bundler' can only be used when 'module' is set to 'es2015' or later.

Kocal (approved):

Kocal:
Thank you @stof.

PR #1415, opened 2026-05-13
[CI] Remove low/high-deps jobs, add Dependabot, rework chunk reference in tests, E2E test lowest bounds

### LINKED ISSUE #1414: CI setup improvements
Our CI setup currently suffers from several shortcoming making it hard to keep the CI green over time. this means that new PRs might contain more changes related to fixing the state of the CI than to the actual change (see #1411 for instance).

This issue is based on a discussion I had with @Kocal on Slack.

### Issues with the setup

- the `highest dependencies` workflow breaks any time vue or typescript are doing patch releases at least (and maybe others as well) due to the functional tests being affected in https://github.com/symfony/webpack-encore/blob/6b4ed498e6505c75301714e7d9738e4df2518d33/test/functional.js#L29
- the `lowest dependencies` workflow breaks regularly in ways that are hard to reproduce (and potentially also for those upstream release reasons sometimes)
- the `lowest dependencies` workflow installs dependencies in a weird way that does not match the way the npm ecosystem works. In npm, the lowest bound of `dependencies` does not matter a lot (as npm does not try to resolve a single version per package) and most packages simply bump it to the latest patch at each release, ensuring they get the uptodate dependency. This workflow aims mostly at ensuring tests against the supported range of our optional dependencies

### Proposal

1. Add end-to-end tests (i.e. in our `test_apps`) covering the lowest bounds of our peer dependencies (no need to duplicate them for each package manager, as they will be about testing other things than the package manager compat,, but they should not be _replacing_ our existing test apps handling package manager compat)
2. Remove the `lowest-deps` workflow (the testing apps workflow will now be covering the use case of testing old peer dependencies)
3. Remove the `highest-deps` workflow (which is broken when some highest deps are newer than the locked ones)
4. add dependabot to update the locked dependencies regularly in an automated way (replacing the need cover by `highest-deps` running in a scheduled way currently)
5. Try to find a way to make `test/functional.js` less sensitive to the installed dependencies (so that dependabot PRs don't fail them each time)

Kocal:
| Q             | A                                                                                                                         |
| ------------- | ------------------------------------------------------------------------------------------------------------------------- |
| Bug fix?      | no                                                                                                                    |
| New feature?  | no                                                                            |
| Deprecations? | no                                                                            |
| Issues        | Fix #1414  |
| License       | MIT                                                                                                                       |


.github/dependabot.yml
```diff
@@ -0,0 +1,67 @@
+# https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
+
+version: 2
+updates:
+    - package-ecosystem: 'npm'
+      directory: '/'
```
stof, line 6: We should probably also have (separate) updates for the lock files of test apps (making them exclude packages that are explicitly testing is peer dependency lowest bound)

package.json
```diff
@@ -59,7 +59,7 @@
         "@symfony/mock-module": "file:fixtures/stimulus/mock-module",
         "@symfony/stimulus-bridge": "^3.0.0 || ^4.0.0",
         "@vue/babel-plugin-jsx": "^1.0.0",
-        "@vue/compiler-sfc": "^3.0.2",
+        "@vue/compiler-sfc": "^3.2.14",
```
Kocal, line 62: Should match the `vue` constraint version

package.json
```diff
@@ -128,7 +128,7 @@
         "typescript": "^5.0.0",
         "vue": "^3.2.14",
         "vue-loader": "^17.0.0",
-        "webpack": "^5.72",
+        "webpack": "^5.82",
```
Kocal, line 131: `pnpm view webpack-cli@6.0.0 peerDependencies` gives `{ webpack: '^5.82.0' }`

.github/dependabot.yml
Kocal: Done

package.json
stof: this will need to be documented in the changelog for Encore 7.0 (like any reduction of a peer dependency range), but it makes sense.

webpack-encore.tgz
stof, line 1: this file must not be committed (I suggest adding an ignore rule to prevent mistakes)

Kocal: Oops yes, didn't see it while commiting.
Fixed

package.json
Kocal: Done

package.json
Kocal: (added in the CHANGELOG)

test_apps/install-deps.sh
```diff
@@ -19,3 +19,8 @@ for dir in $(find . -type d -name "pnpm*"); do
   echo "Installing dependencies for $dir"
   (cd "$dir" && pnpm install)
 done
+
+for dir in $(find . -type d -name "lowest-peers-*"); do
+  echo "Installing dependencies for $dir"
+  (cd "$dir" && pnpm install --ignore-workspace)
+done
```
Kocal, line 26: Since all `lowest-peers-*` apps are using pnpm, maybe it could make sense to use pnpm workspaces instead..

stof: well, those e2e apps are intentionally installing different deps than what we lock for the main testsuite, so it might be hard.

PR #1436, opened 2026-05-30
Unify code minification under `minimizer-webpack-plugin`

### LINKED ISSUE #1432: [RFC] Replace dependency on archived css-minimizer-webpack-plugin
Hello hello, 

## Context

`css-minimizer-webpack-plugin` was archived on May 8, 2026:
https://github.com/webpack/css-minimizer-webpack-plugin

Encore currently uses it (v8.0.0) as the default CSS minifier (via cssnano).
This dependency will receive no further bug fixes, security patches, or 
compatibility updates.

The webpack 2026 roadmap explicitly states the intent to replace it with 
a unified `minimizer-webpack-plugin`:
https://webpack.js.org/blog/2026-02-04-roadmap-2026/

## Problem

Since `css-minimizer-webpack-plugin` is archived, Encore is now depending 
on an unmaintained package for a core build step. This raises questions 
about the migration path, especially in the context of:

- Encore 7.0 (ESM focus)
- Possible Rspack support (#1389)
- webpack's native CSS support (`experimental.css`)

## Questions for maintainers

1. Should Encore proactively replace `css-minimizer-webpack-plugin` with 
   an alternative (Lightning CSS via `lightningcss` directly, or wait for 
   webpack's unified minimizer)?
2. Is this blocked by / related to the Rspack support (#1389)?
3. Should users be warned in the meantime via a deprecation notice or 
   documentation update?

## Possible paths

- **Wait** for webpack's `minimizer-webpack-plugin` and adopt it when stable
- **Switch** to Lightning CSS now via `lightningcss` + a custom webpack 
  minimizer, without going through the archived plugin
- **Document** a workaround for users who want to override the minimizer today

---
> ⚠️ This issue was drafted with the help of an AI assistant (Claude). 
> The research, context, and direction were guided by me — but if anything 
> looks off or needs clarification, happy to dig deeper.

PhilDaiguille:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Issues         | Fix #1432
| License        | MIT

## Summary

`css-minimizer-webpack-plugin` has been archived, and `terser-webpack-plugin` was superseded by `minimizer-webpack-plugin`, the webpack team's unified minimizer package introduced as part of their 2026 roadmap. This PR
migrates Encore to `minimizer-webpack-plugin` and reworks the minimizer configuration API around it.

## Dependencies

`css-minimizer-webpack-plugin` and `terser-webpack-plugin` are removed. `minimizer-webpack-plugin` is added as a direct dependency of Encore. Concrete minifier packages (`lightningcss`, `cssnano`, `csso`, `clean-css`,
`esbuild`, `@swc/core`, `@swc/css`, `uglify-js`) become optional peer dependencies, installed only when needed. Terser is bundled inside `minimizer-webpack-plugin`, so JS minification keeps working out of the box with no
extra install.

## API

Two methods configure minification:

- `configureJsMinimizerPlugin(callback)`: new, configures JS minification.
- `configureCssMinimizerPlugin(callback)`: reworked, now backed by `minimizer-webpack-plugin` instead of `css-minimizer-webpack-plugin`.

Both callbacks receive the `MinimizerPlugin` class as their second argument, so you can switch minifier via `options.minify` without importing `minimizer-webpack-plugin` yourself (it is a transitive dependency of Encore
and would not resolve under pnpm).

If the chosen minifier's backing package is not installed, Encore throws a clear, copy-pasteable error at build time.

Available JS minifiers: `terserMinify` (default, bundled), `esbuildMinify`, `swcMinify`, `uglifyJsMinify`.
Available CSS minifiers: `lightningCssMinify`, `cssnanoMinify`, `cssoMinify`, `cleanCssMinify`, `esbuildMinifyCss`, `swcMinifyCss`.

```js
// Switch JS minifier to esbuild (npm install --save-dev esbuild)
Encore.configureJsMinimizerPlugin((options, MinimizerPlugin) => {
  options.minify = MinimizerPlugin.esbuildMinify;
});

// CSS minification with Lightning CSS (npm install --save-dev lightningcss)
Encore.configureCssMinimizerPlugin((options, MinimizerPlugin) => {
  options.minify = MinimizerPlugin.lightningCssMinify;
});
```

## BC Breaks

**`Encore.configureTerserPlugin()` is removed.** Calling it now throws an explicit error with a migration message pointing to `configureJsMinimizerPlugin()`. The callback signature is the same, so migration is a one-line
rename.

**CSS is no longer minified by default.** `cssnano` was previously pulled in transitively by `css-minimizer-webpack-plugin` and is no longer installed. To minify CSS in production, install a minifier and opt in via
`configureCssMinimizerPlugin()`. JS minification is unchanged.

To match the previous default as closely as possible, install `cssnano` and `postcss`, then:

```js
Encore.configureCssMinimizerPlugin((options, MinimizerPlugin) => {
  options.minify = MinimizerPlugin.cssnanoMinify;
});
```


lib/WebpackConfig.js
```diff
@@ -202,10 +202,10 @@ class WebpackConfig {
         this.friendlyErrorsPluginOptionsCallback = () => {};
         /** @type {OptionsCallback<object>} */
         this.manifestPluginOptionsCallback = () => {};
-        /** @type {OptionsCallback<import('terser-webpack-plugin').BasePluginOptions & import('terser-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('terser').MinifyOptions>>} */
-        this.terserPluginOptionsCallback = () => {};
-        /** @type {OptionsCallback<import('css-minimizer-webpack-plugin').BasePluginOptions & import('css-minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('css-minimizer-webpack-plugin').CssNanoOptionsExtended>>} */
-        this.cssMinimizerPluginOptionsCallback = () => {};
+        /** @type {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<object>>} */
```
stof: why `object` here ? Can't we have a better type for the options there ? It think it could be `import('terser').MinifyOptions` instead of `object` (as done in the existing code btw).

lib/WebpackConfig.js
```diff
@@ -350,18 +350,52 @@ class WebpackConfig {
     }
 
     /**
-     * @param {OptionsCallback<import('terser-webpack-plugin').BasePluginOptions & import('terser-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('terser').MinifyOptions>>} terserPluginOptionsCallback
+     * Configures the options passed to the minimizer-webpack-plugin for JS and/or CSS.
+     *
+     * @param {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<object>>|null} jsOptionsCallback
+     * @param {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('cssnano').Options>>|null} cssOptionsCallback
+     */
+    configureMinimizerPlugin(jsOptionsCallback = null, cssOptionsCallback = null) {
```
stof: If we have 2 separate instances of the plugin (one for JS and one for CSS), we could keep 2 separate `configure` methods to pass the callbacks.
What do you think @Kocal ?

lib/WebpackConfig.js
```diff
@@ -350,18 +350,52 @@ class WebpackConfig {
     }
 
     /**
-     * @param {OptionsCallback<import('terser-webpack-plugin').BasePluginOptions & import('terser-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('terser').MinifyOptions>>} terserPluginOptionsCallback
+     * Configures the options passed to the minimizer-webpack-plugin for JS and/or CSS.
+     *
+     * @param {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<object>>|null} jsOptionsCallback
```
stof: this should use `import('terser').MinifyOptions` instead of `object`.

lib/WebpackConfig.js
```diff
@@ -350,18 +350,52 @@ class WebpackConfig {
...
+        }
+
+        if (cssOptionsCallback !== null) {
+            this.minimizerPluginCssOptionsCallback = cssOptionsCallback;
+        }
+    }
+
+    /**
+     * @deprecated Use configureMinimizerPlugin() instead.
+     * @param {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<object>>} terserPluginOptionsCallback
```
stof: `import('terser').MinifyOptions` instead of `object` here as well.

lib/WebpackConfig.js
Kocal: I don't think we can re-use `import('terser').MinifyOptions` or any specific (and single) minimizer like SWC or esbuild. Can we reference all of them? Or does the plugin expose a better type?

lib/WebpackConfig.js
Kocal: I'm also in favor of one method `.configure...` methods per type of minimizer. So, two methods for CSS and JS minification.

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+    Encore.configureMinimizerPlugin(null, (cssOptions) => {
+        cssOptions.minify = MinimizerPlugin.lightningCssMinify;
+    });
+    ```
+
+    See the [minimizer-webpack-plugin documentation](https://github.com/webpack/minimizer-webpack-plugin)
+    and [lightningcss documentation](https://lightningcss.dev) for all available options.
+
+- Deprecate `configureTerserPlugin()` in favor of `configureMinimizerPlugin(jsCallback, null)`.
+- Deprecate `configureCssMinimizerPlugin()` in favor of `configureMinimizerPlugin(null, cssCallback)`.
```
Kocal: If the PR is almost ready, you should move this text to the `7.0.0` section, explaining the different modifications and BCs, and upgrade steps.

lib/WebpackConfig.js
stof: The generic type in the plugin is precisely meant to ensure that the minimizer being used matches the options being provided (allowing each minimizer to define its own option shape).

Our callback API allowing us to *mutate* the options makes it hard to define types if the goal is to support changing the minimizer being used through such mutation. The Encore API is harder to type statically than the original API where the userland code is the one instantiating the plugin with its options (and inference runs on that code).

lib/WebpackConfig.js
```diff
@@ -202,10 +202,10 @@ class WebpackConfig {
         this.friendlyErrorsPluginOptionsCallback = () => {};
         /** @type {OptionsCallback<object>} */
         this.manifestPluginOptionsCallback = () => {};
-        /** @type {OptionsCallback<import('terser-webpack-plugin').BasePluginOptions & import('terser-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('terser').MinifyOptions>>} */
-        this.terserPluginOptionsCallback = () => {};
-        /** @type {OptionsCallback<import('css-minimizer-webpack-plugin').BasePluginOptions & import('css-minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('css-minimizer-webpack-plugin').CssNanoOptionsExtended>>} */
-        this.cssMinimizerPluginOptionsCallback = () => {};
+        /** @type {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('terser').MinifyOptions>>} */
```
Kocal: I still don't understand why do we have to specify `terser` here, what if the user use `esbuild`?

lib/WebpackConfig.js
```diff
@@ -202,10 +202,10 @@ class WebpackConfig {
...
+        this.minimizerPluginJsOptionsCallback = () => {};
+        /** @type {OptionsCallback<import('minimizer-webpack-plugin').BasePluginOptions & import('minimizer-webpack-plugin').DefinedDefaultMinimizerAndOptions<import('cssnano').Options>>} */
```
Kocal: Same here for `cssnano`, what if the user use `lightningcss`?

package.json
```diff
@@ -33,7 +33,8 @@
         "@kocal/friendly-errors-webpack-plugin": "^3.0.0",
         "babel-loader": "^10.0.0",
         "css-loader": "^7.1.0",
-        "css-minimizer-webpack-plugin": "^8.0.0",
+        "cssnano": "^8.0.1",
```
Kocal: During the last Encore releases, we tried to reduce the number of dependencies (for installation speed, reduce disk writes, bandwidth consumption, reduce security issues, ...) and forcing the user to use `cssnano`, where lightest and better alternatives exist, is a big 👎🏻 for me.

For example, [latest cssnano](https://npmgraph.js.org/?q=cssnano#select=cssnano%408.0.1) weights ~30MB and ships with 66 dependencies, where [latest lightningcss](https://npmgraph.js.org/?q=lightningcss#select=lightningcss%401.32.0) weight ~500KB and ships with 1 dependency

Kocal (changes requested):

package.json
PhilDaiguille: That’s a fair point. I used cssnano because it’s what Encore currently relies on and I wanted to avoid changing the existing behavior in this PR. That said, I agree that Lightning CSS looks like a much better default nowadays given its size, dependency count and performance benefits. I’m happy to switch to Lightning CSS if that’s the preferred direction.

Kocal: Since this PR will be part of the new 7.0 release, BC are totally allowed and encouraged. 

I would suggest:
- to not enforce any minimizer, 
- add peerDependencies (based to https://github.com/webpack/minimizer-webpack-plugin/blob/main/package.json#L126-L163)
- add two runtime checks: 
  1. if CSS minification is enabled, and no CSS minifier is installed/enabled/passed -> throw error
  2. if JS minification is enabled, and no JS minifier is installed/enabled/passed -> throw error

WDYT @stof ?

garak:
What about moving to Babel 8?

Kocal:
> What about moving to Babel 8?

Unrelated with the PR

garak:
> > What about moving to Babel 8?
> 
> Unrelated with the PR

Sorry, I'm not a node expert myself, but I thought that the major version change was a breaking change

stof:
> Sorry, I'm not a node expert myself, but I thought that the major version change was a breaking change

this does not make it related to the current PR

lib/WebpackConfig.js
stof: because *we* define the type of minimizer for which you can define the settings. See the discussion in https://github.com/symfony/webpack-encore/pull/1436#discussion_r3332670260

package.json
PhilDaiguille: For my part, I'm fine with doing a BC

Kocal:
I will take care of the rest, thank you for your work :)

Kocal:
PR and description updated, I removed `Encore.configureTerserPlugin()` instead of deprecating it (it didn't make any sense since we are targeting a new major version).
Added a bunch of tests, and checking for CSS/JS minimizers packages to be installed before using their associated minifiers.

Kocal (approved):

Kocal:
Thank you @PhilDaiguille.

lib/utils/apply-options-callback.js
```diff
@@ -8,18 +8,19 @@
  */
 
 /**
- * @typedef {function(this: T, T): T|void} OptionsCallback
+ * @typedef {function(this: T, T, ...*): T|void} OptionsCallback
```
stof, line 11: I suggest creating a *separate* type for the case of the minifier callback, with precise type, instead of making the type less precise for *all* option callbacks.

Kocal: Yes you're right, I was a bit sad few hours ago when I didn't had auto-completion on the callback :(

PR #1498, opened 2026-06-22
Relax the `peerDependencies` of the JS/CSS minifiers

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Issues         | Fix #... 
| License        | MIT

My mistake, while working on https://github.com/symfony/webpack-encore/pull/1436, I forgot to remove these peerDependencies to forward them to the MinimizerWebpackPlugin


stof:
Is this actually working for pnpm or yarn in Plug'n'play mode, which are strict about accessing peer dependencies ?

Removing the peer dependencies does not *forward* them.

test_apps/install-deps.sh
```diff
@@ -12,12 +12,12 @@ done
...
 for dir in $(find . -type d -name "yarn*"); do
   echo "Installing dependencies for $dir"
-  (cd "$dir" && yarn)
+  (cd "$dir" && corepack enable && yarn)
 done
 
 for dir in $(find . -type d -name "pnpm*"); do
   echo "Installing dependencies for $dir"
-  (cd "$dir" && pnpm install)
+  (cd "$dir" && pnpm install --ignore-workspace)
```
stof, line 20: why `--ignore-workspace` ?

stof:
ah, I see that they are still defined in `peerDependenciesMeta`, so yarn will treat them as peer dependencies but without a constraint.

test_apps/install-deps.sh
Kocal: The fix is not related to the PR, but running `pnpm install` inside a `test_apps/*` based on pnpm didn't work really well. I re-used the same options than this code-block lines 23-27:
```shell
for dir in $(find . -type d -name "lowest-peers-*"); do
  echo "Installing dependencies for $dir"
  (cd "$dir" && pnpm install --ignore-workspace)
done
```

Kocal:
Yes, I re-used the same solution than MinimizerWebpackPlugin, see https://github.com/webpack/minimizer-webpack-plugin/blob/b31cc7bb45fa97aefdec62d1c0df6ba40533aa11/package.json#L126-L162

PR #1502, opened 2026-06-23
Prepare TypeScript migration

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Issues         | Fix #... 
| License        | MIT

Related to https://github.com/symfony/webpack-encore/issues/816

This PR prepare the TypeScript migration, by:
- using `allowJs: true` in `tsconfig.json` configuration
- migrating some files under `lib/`, they were easy to migrate
- adding new job in the CI, checking types validity and if build is successful
- the `bin/encore.js` tests are now testing the file `dist/bin/encore.js` 
- still in the CI, node.js dependencies are now installed for `test_apps/*` jobs in order to success the build
- added some guards in `package.json`, `prepare` automatically build the project and `prepublicOnly` check types validity

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -59,10 +59,6 @@ export default [
             ],
             'n/no-unsupported-features/node-builtins': ['error'],
             'n/no-deprecated-api': 'error',
-            // The TypeScript compiler owns module-resolution checking: `.js`
-            // import specifiers resolve to `.ts` source files during the TS
-            // migration, which eslint-plugin-n cannot follow.
-            'n/no-missing-import': 'off',
             'n/no-missing-require': 'off',
             'n/no-unpublished-bin': 'error',
             'n/no-unpublished-import': 'error',
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -13,6 +13,7 @@ import jsdoc from 'eslint-plugin-jsdoc';
 import nodePlugin from 'eslint-plugin-n';
 import vitestPlugin from 'eslint-plugin-vitest';
 import globals from 'globals';
+import tseslint from 'typescript-eslint';
 
 export default [
     js.configs.recommended,
@@ -58,7 +59,10 @@ export default [
             ],
             'n/no-unsupported-features/node-builtins': ['error'],
             'n/no-deprecated-api': 'error',
-            'n/no-missing-import': 'error',
+            // The TypeScript compiler owns module-resolution checking: during
+            // the TS migration, `.js` import specifiers resolve to `.ts` source
+            // files, which eslint-plugin-n cannot follow.
+            'n/no-missing-import': 'off',
             'n/no-missing-require': 'off',
             'n/no-unpublished-bin': 'error',
             'n/no-unpublished-import': 'error',
@@ -82,6 +86,20 @@ file that was distributed with this source code.`,
             jsdoc: {
                 mode: 'typescript',
             },
+        },
+    },
+    {
+        files: ['**/*.ts'],
+        languageOptions: {
+            parser: tseslint.parser,
+        },
+        rules: {
+            // The TypeScript compiler reports unused symbols; the core rule
+            // produces false positives on type-only syntax.
+            'no-unused-vars': 'off',
+            // In TypeScript files, types live in the signature, not in JSDoc.
+            'jsdoc/require-param': 'off',
+            'jsdoc/require-returns': 'off',
         },
     },
     {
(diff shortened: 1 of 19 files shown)


eslint.config.js
```diff
@@ -58,7 +59,10 @@ export default [
             ],
             'n/no-unsupported-features/node-builtins': ['error'],
             'n/no-deprecated-api': 'error',
-            'n/no-missing-import': 'error',
+            // The TypeScript compiler owns module-resolution checking: during
+            // the TS migration, `.js` import specifiers resolve to `.ts` source
+            // files, which eslint-plugin-n cannot follow.
```
stof: Typescripts supports using `ts` extensions (it requires an opt-in in tsconfig.json through `"allowImportingTsExtensions": true`).

This would also make it possible to rely on the node.js type stripping during dev to run tests without running the build.

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "extends": "@tsconfig/node22/tsconfig.json",
+    "compilerOptions": {
+        // Base options
+        "esModuleInterop": true,
+        "skipLibCheck": true,
+        "allowJs": true,
+        "resolveJsonModule": true,
+        "moduleDetection": "force",
+        "isolatedModules": true,
+        "verbatimModuleSyntax": true,
```
stof, line 12: I suggest `"erasableSyntaxOnly": true`, so that typescript forbids using the few syntaxes that cannot be type-erased (typescript enums, which are not recommended anymore, and a few others). This ensures our code is compatible with node's type stripping, for future local dev setup that run tests without running the build.

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+
+        // Library build
+        "rootDir": ".",
+        "outDir": "dist",
+        "declaration": true,
+        "noEmitOnError": true,
+
+        // Strictness
+        "strict": false, // Disabled for now to ease migration from JS to TS, but we should enable it in the future
```
stof: Is this actually making the migration simpler ?

tsconfig.json
Kocal: I didn't know about this one, sure!

eslint.config.js
Kocal: Will do that, thanks

tsconfig.json
Kocal: Actually it has no effect on the three migrates files if I enable or disable strict-checking.
I will enable it

eslint.config.js
Kocal: Unfortunately it didn't work, disabling `n/no-missing-import` show these errors:
```
/Users/kocal/workspace/symfony/webpack-encore/lib/WebpackConfig.js
  46:27  error  Can't resolve './utils/regexp-escaper.js' in '/Users/kocal/workspace/symfony/webpack-encore/lib'  n/no-missing-import

/Users/kocal/workspace/symfony/webpack-encore/lib/config-generator.js
  49:27  error  Can't resolve './utils/string-escaper.js' in '/Users/kocal/workspace/symfony/webpack-encore/lib'  n/no-missing-import

/Users/kocal/workspace/symfony/webpack-encore/test/helpers/assert.js
  15:26  error  Can't resolve '../../lib/utils/regexp-escaper.js' in '/Users/kocal/workspace/symfony/webpack-encore/test/helpers'  n/no-missing-import
```

If I change the imports from `<path>.js` to `<path>.ts` to make ESLint happy, then TypeScript keep the `<path>.ts` in transpiled files under `dist/`, making Encore unusable since native TypeScript execution with Node.js 22.18 is not enabled on files from `node_modules/`.

And `tsc` is already able to validate imports resolutions, I suggest keeping this rule disabled.

stof: Reading the docs again, there is another setting `rewriteRelativeImportExtensions: true` that is needed for that

Kocal: Indeed it works great now, re-enabling `n/no-missing-import` rule. I should read `tsc` output next time 😅 

```
➜  webpack-encore git:(prepare-typescript-migrations) ✗ pnpm run build

> @symfony/webpack-encore@7.1.0 build /Users/kocal/workspace/symfony/webpack-encore
> rm -fr dist/ && tsc

tsconfig.json:14:39 - error TS5096: Option 'allowImportingTsExtensions' can only be used when one of 'noEmit', 'emitDeclarationOnly', or 'rewriteRelativeImportExtensions' is set.

14         "allowImportingTsExtensions": true,
                                         ~~~~
```

Kocal:
I changed the base branch for a dedicated Typescript-migration branch, in case of we have to release fixes during the migration.

PR #1503, opened 2026-06-25
Migrate project to TypeScript ✨

### LINKED ISSUE #816: [RFC] Ship typescript types for the public APIs

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Issues         | Fix #816 
| License        | MIT


Kocal:
**TIL:** packing Encore with `pnpm` and installing it in another project with `pnpm add -D ../path/to/encore.tgz` breaks the auto-completion in `webpack.config.mjs` in PHPStorm (still works in VSCode), when it works with `webpack.config.mts` in PHPStorm (and VSCode too). 
But, installing Encore in a test project with `pnpm add -D github:symfony/webpack-encore#typescript-migration` works! Surely `pnpm` is doing something weird that PHPStorm can't resolve, but at least we know.

PR #1506, opened 2026-06-25
[TypeScript] Migrate `lib/loaders/*`

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Issues         | Fix #... 
| License        | MIT

Related to #1503

### DIFF
--- a/lib/loaders/css-extract.ts
+++ b/lib/loaders/css-extract.ts
@@ -7,25 +7,18 @@
  * file that was distributed with this source code.
  */
 
-/**
- * @import WebpackConfig from '../WebpackConfig.js'
- */
-
 import { fileURLToPath } from 'url';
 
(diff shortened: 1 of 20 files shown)


lib/loaders/css-extract.ts
```diff
@@ -7,25 +7,18 @@
...
 export default {
     /**
      * Prepends loaders with MiniCssExtractPlugin.loader
-     *
-     * @param {WebpackConfig} webpackConfig
-     * @param {Array} loaders An array of some style loaders
-     * @returns {Array}
      */
-    prependLoaders(webpackConfig, loaders) {
+    prependLoaders(webpackConfig: WebpackConfig, loaders: object[]) {
```
stof, line 21: instead of `object[]`, shouldn't this use the appropriate type from webpack ? It is not arbitrary objects.

PR #1519, opened 2026-07-17
Throw a clear error when a callback is passed as the first argument of configureImageRule()/configureFontRule()

### LINKED ISSUE #1221: Overriding the file name for "configureFontRule" or "configureImageRule" doesn't work when using a generator
According to the documentation, a custom callback can be passed to the `Encore.configureFontRule()` and `Encore.configureImageRule()` functions: https://github.com/symfony/webpack-encore/blob/main/index.js#L1457.

However, when using this callback to specify a custom generator, the `filename` option of the generator seems to get ignored:

```js
Encore
    .configureFontRule(rule => {
        rule.test = fontFileExtensionRegex;
        rule.generator = {
            filename: 'whatever/path/[name][ext]',
        };
    });
```

The issue seems to be this default `filename` option: https://github.com/symfony/webpack-encore/blob/main/lib/WebpackConfig.js#L78.

Using the same example, explicitly setting `filename` to `global.undefined` in the regular options seems to fix the issue:

```js
Encore
    .configureFontRule(
        {
            filename: global.undefined,
        },
        rule => {
            rule.test = fontFileExtensionRegex;
            rule.generator = {
                filename: 'whatever/path/[name][ext]',
            };
        },
    );
```

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Issues         | Fix #1221
| License        | MIT

`configureImageRule()` and `configureFontRule()` take `(options, ruleCallback)`. When a callback is passed as the first argument, for example:

```js
Encore.configureFontRule(rule => {
    rule.oneOf[1].generator.filename = 'whatever/[name][ext]';
});
```

the function lands in the `options` parameter. `Object.keys(aFunction)` is empty, so no option validation fails, and the callback is silently dropped: the rule is never customized. This is a common mistake because several other `Encore.configure*()` methods take a single callback.

This PR turns that silent no-op into a helpful error pointing at the correct 2-argument form:

> configureFontRule() expects an options object as its first argument. To pass only a callback, use configureFontRule({}, callback).

Only broken calls are affected (a function is never a valid `options` value), so there is no BC impact on working code.

This addresses the most surprising part of #1221 (the callback being ignored without any feedback). Note that overriding the filename does work today through the documented 2-argument callback by modifying the nested `rule.oneOf[1].generator`; setting `rule.generator` at the top level has no effect because Encore wraps the asset rule in a `oneOf`. Happy to follow up on that ergonomics point separately if you think it is worth improving.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+- Throw a helpful error when a callback is passed as the first argument of `configureImageRule()` or `configureFontRule()` instead of being silently ignored
+
 ## 7.2.0
 
 - Migrate internal code to TypeScript, ship package with type definitions (better DX and IDE support!)
--- a/lib/WebpackConfig.ts
+++ b/lib/WebpackConfig.ts
@@ -1267,6 +1267,12 @@ class WebpackConfig {
         } = {},
         ruleCallback: OptionsCallback<webpack.RuleSetRule> = () => {}
     ) {
+        if (typeof options === 'function') {
+            throw new Error(
+                'configureImageRule() expects an options object as its first argument. To pass only a callback, use configureImageRule({}, callback).'
+            );
+        }
+
         for (const optionKey of Object.keys(options)) {
             if (!(optionKey in this.imageRuleOptions)) {
                 throw new Error(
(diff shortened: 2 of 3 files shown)


stof (approved):

Kocal:
Hi, and thanks for opening this PR!

I just merged https://github.com/symfony/webpack-encore/pull/1521, and with TypeScript types definitions, I'm not sure this kind of PR are still needed (same for existing type validation at runtime). 

WDYT?

Amoifr:
Thanks! I'd say this one is still useful: types only help people whose editor type-checks their `webpack.config.js`, but the recipe generates a plain JS file with no `@ts-check` and no compile step, so for most users nothing flags the mistake and the callback is still silently dropped at runtime. Since the failure mode is a silent no-op rather than an error, I think the runtime guard keeps its value as a safety net under the types. That said, you know the roadmap best, so if you'd rather rely on types going forward I won't fight for it. 😄

Kocal:
Thank you @Amoifr.

PR #1520, opened 2026-07-21
Recommend the project's package manager for missing dependency errors

### LINKED ISSUE #57: friendly-errors-webpack-plugin suggests `npm install`

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Issues         | Fix #57
| License        | MIT

When a required module cannot be resolved, the friendly-errors plugin suggests installing it with a hardcoded `npm install --save`, even in projects using Yarn or pnpm (the original report from @weaverryan, reopened by @stof).

Encore already detects the package manager from the lockfile when recommending its own build dependencies (`package-helper.js`). This reuses that detection for the "module not found" message: a new transformer re-types the errors that point to an installable package, and a matching formatter recommends the right command:

```
This dependency was not found:

* foo in ./assets/app.js

To install it, you can run: pnpm add foo
```

- non-relative modules only are handled, so relative imports (`./missing`) keep the plugin's untouched "This relative module was not found" message, with no install suggestion;
- the lockfile detection is now shared through a small `getPackageManager()` helper, used both by the existing `getInstallCommand()` and the new one;
- added unit tests for the transformer, the formatter (npm/yarn/pnpm) and `getMissingPackagesInstallCommand()`, and updated the plugin's transformer/formatter counts.

I left `CHANGELOG.md` untouched since there is no unreleased section right now, but happy to add an entry wherever you prefer.

Thanks a lot for Encore, and for keeping this one on the radar since 2017! 🙏

### DIFF
--- a/lib/friendly-errors/transformers/missing-dependency.ts
+++ b/lib/friendly-errors/transformers/missing-dependency.ts
@@ -1,0 +1,33 @@
+/*
+ * This file is part of the Symfony Webpack Encore package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import type { FriendlyError } from '@kocal/friendly-errors-webpack-plugin';
+
+const TYPE = 'missing-dependency';
+
+function isRelative(module: string): boolean {
+    // package names can never start with a dot, so this covers "./", "../"
+    // and their Windows counterparts ".\" and "..\"
+    return module.startsWith('.');
+}
+
+function transform(error: FriendlyError): FriendlyError {
+    // the friendly-errors plugin already flags "module not found" errors with
+    // the "module-not-found" type. We re-type the ones pointing to an installable
+    // package (i.e. not a relative path) so that we can recommend an install
+    // command that matches the package manager used in the project, instead of
+    // the plugin's hardcoded "npm install" suggestion.
+    if (error.type !== 'module-not-found' || !error.module || isRelative(error.module)) {
+        return error;
+    }
+
+    return Object.assign({}, error, { type: TYPE });
+}
+
+export default transform;
--- a/lib/friendly-errors/formatters/missing-dependency.ts
+++ b/lib/friendly-errors/formatters/missing-dependency.ts
@@ -1,0 +1,73 @@
+/*
+ * This file is part of the Symfony Webpack Encore package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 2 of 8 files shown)


Kocal:
That seems to be a nice addition, good!

Can you rebase your PR and migrate it to TypeScript please? I just merged https://github.com/symfony/webpack-encore/pull/1503 and it made conflicts on your branch.

Thanks!

Amoifr:
Thanks for the kind words! Rebased on `main` and migrated to TypeScript (the new formatter/transformer now speak fluent `FriendlyError`). All green locally: tsc, eslint, oxfmt and the unit suite. 😄

lib/friendly-errors/transformers/missing-dependency.ts
```diff
@@ -0,0 +1,31 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import type { FriendlyError } from '@kocal/friendly-errors-webpack-plugin';
+
+const TYPE = 'missing-dependency';
+
+function isRelative(module: string): boolean {
+    return module.startsWith('./') || module.startsWith('../');
```
Kocal: No Windows support?

Kocal (commented):
~~Missing a changelog entry~~

lib/friendly-errors/transformers/missing-dependency.ts
Amoifr: Good catch! Package names can never start with a dot, so I simplified the check to `module.startsWith('.')`, which also covers the Windows-style `.\missing` and `..\missing`. Added a test for it.

Kocal:
Thank you @Amoifr.

lib/friendly-errors/transformers/missing-dependency.ts
stof: AFAIK, module specifiers are using `/` all the time.

Amoifr: Indeed, import specifiers normalize to `/`, so the original check was probably fine in practice. I kept the simplified `startsWith('.')` anyway since package names can never start with a dot, it makes the intent clearer.

